PR #8, opened 2022-04-15
refactor(gateway): client abstraction

tarrencev:
this proposes a different client abstraction, taken from my original implementation at `go-starknet`.

- makes the api interfaces safer and more explicit. 
- uses a custom error type that allows callers to easily match errors while still including useful dynamic context.
- encapsulates common components in their own file so it is easier to grok the code

if you like this approach, i can apply the same changes to other endpoints

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -6,3 +6,4 @@ vendor/
 tmp/
 
 examples/**/*.json
+examples/**/*.sum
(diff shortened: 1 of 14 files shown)


drspacemn (approved):
Yeah the API interfaces/custom error handling looks great.

Separate files will be nice for further component build outs.

PR #9, opened 2022-04-15
refactor: transactions api

tarrencev:


drspacemn (approved):
I pulled the rest of the transaction functions into the transactions file and cleaned up some of the redundant structs.

The result is a much cleaner interaction with the starknet functions. User no longer needs to call things like `HexToBig(GetSelectorFromName` and the api can handle that internally

PR #13, opened 2022-04-20
refactor account api

drspacemn:
- rename snsetters -> starknet;
- delete sn_models;
- delete sn_getters


account.go
```diff
@@ -0,0 +1,160 @@
...
+	- private signing key
+	- stark curve definition
+	- full StarknetGateway definition
+	- public key pair for signature verifications
+*/
+func (sc StarkCurve) NewSigner(private, pubX, pubY *big.Int, chainId ...string) (signer *Signer, err error) {
+	if len(sc.ConstantPoints) == 0 {
+		return signer, fmt.Errorf("must initiate precomputed constant points")
+	}
+	var gw *StarknetGateway
```
tarrencev: I think it would be more intuitive to pass in the provider than instantiate it internally. There are configuration options that cant be given here and eventually provider should probably be an interface, in which case other implementations can easily be passed in

account.go
```diff
@@ -0,0 +1,160 @@
...
+}
+
+/*
+	Instantiate a new StarkNet Signer which includes structures for calling the network and signing transactions:
+	- private signing key
+	- stark curve definition
+	- full StarknetGateway definition
+	- public key pair for signature verifications
+*/
+func (sc StarkCurve) NewSigner(private, pubX, pubY *big.Int, chainId ...string) (signer *Signer, err error) {
```
tarrencev: This is an opinion so feel free to ignore. I generally prefer not to use named returns. It is harder to parse the logic, since they can be set arbitrarily and returned implicitly, which makes them more error prone

account.go
```diff
@@ -0,0 +1,160 @@
...
+	for _, tx := range txs {
+		callArray = append(callArray, SNValToBN(tx.ContractAddress), GetSelectorFromName(tx.EntryPointSelector))
+		if len(tx.Calldata) == 0 {
+			callArray = append(callArray, big.NewInt(0), big.NewInt(0))
+		} else {
+			callArray = append(callArray, big.NewInt(int64(len(calldataArray))), big.NewInt(int64(len(tx.Calldata))))
+			for _, cd := range tx.Calldata {
+				calldataArray = append(calldataArray, SNValToBN(cd))
+			}
+		}
```
tarrencev: nit: prefer to minimize branching and return/continue early when possible

```suggestion
		if len(tx.Calldata) == 0 {
			callArray = append(callArray, big.NewInt(0), big.NewInt(0))
			continue
		}
		callArray = append(callArray, big.NewInt(int64(len(calldataArray))), big.NewInt(int64(len(tx.Calldata))))
		for _, cd := range tx.Calldata {
			calldataArray = append(calldataArray, SNValToBN(cd))
		}
```

gateway.go
```diff
@@ -69,7 +69,12 @@ func NewGateway(opts ...GatewayOption) *StarknetGateway {
 func (sg *StarknetGateway) newRequest(
 	ctx context.Context, method, endpoint string, body interface{},
 ) (*http.Request, error) {
-	req, err := http.NewRequestWithContext(ctx, method, sg.Feeder+endpoint, nil)
+	url := sg.Feeder+endpoint
+	if strings.Contains(endpoint, "add_transaction") {
```
tarrencev: `HasSuffix` would work here i think? minor improvement but safer

```suggestion
	if strings.HasSuffix(endpoint, "add_transaction") {
```

starknet.go
```diff
@@ -0,0 +1,235 @@
...
+	if err != nil {
+		return cc, err
+	}
+
+	var buf bytes.Buffer
+	zw := gzip.NewWriter(&buf)
+	_, err = zw.Write(pay)
+	if err != nil {
+		return cc, err
+	}
```
tarrencev: nit: minimize err scope when possible

```suggestion
	if _, err := zw.Write(pay); err != nil {
		return cc, err
	}
```

starknet.go
```diff
@@ -0,0 +1,235 @@
...
+	deployRequest.Type = DEPLOY
+	if len(deployRequest.ConstructorCalldata) == 0 {
+		deployRequest.ConstructorCalldata = []string{}
+	}
+
+	var rawDef RawContractDefinition
+	err = json.Unmarshal(dat, &rawDef)
+	if err != nil {
+		return resp, err
+	}
```
tarrencev: nit: minimize err scope when possible

```suggestion
	if err = json.Unmarshal(dat, &rawDef); err != nil {
		return resp, err
	}
```

tarrencev (approved):
minor feedback, generally looks great!

account.go
drspacemn: I don't feel strongly either way, but consistency in the lib is important to me. 

Looks like I already had a mishmash of return syntax so, lets go forward with unnamed returns and weed named returns out when we come across them?

account.go
drspacemn: agreed

PR #14, opened 2022-04-23
refactor: move gateway to its own package

tarrencev:
basically i moved the starknet gateway to a separate package, defined a rpc provider interface, created a gateway wrapper to conform it to the rpc provider interface, and started implementation on the json rpc provider


gateway/transaction.go
```diff
@@ -9,51 +9,36 @@ import (
...
+type StarknetTransaction struct {
+	TransactionIndex int         `json:"transaction_index"`
+	BlockNumber      int         `json:"block_number"`
+	Transaction      Transaction `json:"transaction"`
+	BlockHash        string      `json:"block_hash"`
+	Status           string      `json:"status"`
 }
 
-// Starknet transaction composition
 type Transaction struct {
```
drspacemn, line 24: Can we get rid of this and 'TransactionReciept' in favor of the types package?

gateway/transaction.go
```diff
@@ -9,51 +9,36 @@ import (
...
 	Nonce              string   `json:"nonce,omitempty"`
 }
 
-type StarknetTransaction struct {
-	TransactionIndex int         `json:"transaction_index"`
-	BlockNumber      int         `json:"block_number"`
-	Transaction      Transaction `json:"transaction"`
-	BlockHash        string      `json:"block_hash"`
-	Status           string      `json:"status"`
+func (t Transaction) Normalize() *types.Transaction {
```
drspacemn, line 35: I understand Normalize on the 'Block' type as its a different payload depending on the provider, but I don't think I understand here. 

Can we use the types.Transaction and type.TransactionReceipt here and omit the call to Normalize?

That gives the caller access to all the non normalized fields if they need them

gateway/starknet_test.go
```diff
@@ -9,23 +9,26 @@ import (
...
+	curve, err := caigo.SC(caigo.WithConstants("./pedersen_params.json"))
 	if err != nil {
 		t.Errorf("Could not init with constant points: %v\n", err)
 	}
 
 	priv, _ := new(big.Int).SetString("3904bd288b88a1dcd73e648b10642d63cb9b2ffd86526deee9d073f0690139e", 16)
 	x, y, _ := curve.PrivateToPoint(priv)
 
-	signer, err := curve.NewSigner(priv, x, y, NewGateway())
+	signer, err := curve.NewSigner(priv, x, y, NewProvider())
```
drspacemn, line 26: `./starknet_test.go:26:45: cannot use NewProvider() (value of type *GatewayProvider) as type types.Provider in argument to curve.NewSigner:
        *GatewayProvider does not implement types.Provider (wrong type for BlockByHash method)
                have BlockByHash(ctx context.Context, hash string) (*types.Block, error)
                want BlockByHash(context.Context, string, string) (*types.Block, error)`

gateway/provider.go
```diff
@@ -0,0 +1,44 @@
...
+	Gateway
+}
+
+func NewProvider(opts ...Option) *GatewayProvider {
+	return &GatewayProvider{
+		*NewClient(opts...),
+	}
+}
+
+func (p *GatewayProvider) BlockByHash(ctx context.Context, hash string) (*types.Block, error) {
```
drspacemn: It looks like this doesn't implement the Provider interface

tarrencev: ah good catch!

gateway/transaction.go
tarrencev: My thought was the base types should conform to the official spec at https://github.com/starkware-libs/starknet-specs/blob/master/api/starknet_api_openrpc.json#L919

They are a subset of the gateway fields right now. I think that if people start depending on members that aren't part of the spec it will cause more migration pain later, since they might need to be removed eventually.

So, if a caller wants to ensure backward compatibility, they should use the provider interface/types. If they want all the gateway fields, they can use the Gateway client

drspacemn: Cool, that makes sense.

types/types.go
```diff
@@ -0,0 +1,141 @@
...
+type TransactionStatus struct {
+	TxStatus        string `json:"tx_status"`
+	BlockHash       string `json:"block_hash"`
+	TxFailureReason struct {
+		ErrorMessage string `json:"error_message,omitempty"`
+	} `json:"tx_failure_reason,omitempty"`
+}
+
+type Transaction struct {
+	TransactionReceipt
```
drspacemn, line 57: Do we need the receipt nested in the transaction here?

Looks like it breaks signer.Execute

account.go
```diff
@@ -46,59 +53,39 @@ func (sc StarkCurve) NewSigner(private, pubX, pubY *big.Int, gw *StarknetGateway
...
 	}
 
 	maxFee := big.NewInt(0)
+	chainID, err := signer.provider.ChainID(ctx)
+	if err != nil {
+		return nil, err
+	}
 
-	hash, err := signer.Curve.HashMulticall(address, nonce, maxFee, UTF8StrToBig(signer.Gateway.ChainId), txs)
+	hash, err := signer.Curve.HashMulticall(address, nonce, maxFee, HexToBN(chainID), txs)
```
drspacemn: This will need to be flipped back to `UTF8StrToBig(chainID)`

tarrencev: Ah my bad will switch it back!

types/types.go
tarrencev: Hm it is to support the `FULL_TXN_AND_RECEIPTS` scope, which returns the transactions with receipts inlined. Otherwise we'll need to introduce a new type to handle the marshaling. I can take a look at `signer.Execute` but lmk if you have any other ideas

drspacemn: I see. Seems like `omitempty` should do it. The gateway rejects the invoke request with the extra tx receipt fields.

tarrencev: cool i updated the structs to be pointers and to omit them if nil. looks like the tests are passing now!

drspacemn (approved):

PR #19, opened 2022-05-03
refactor(jsonrpc): added more functions and code more versatile

Larkooo:


jsonrpc/client.go
```diff
@@ -53,37 +53,75 @@ func (sc *Client) AccountNonce(context.Context, string) (*big.Int, error) {
 	panic("not implemented")
 }
 
+func (sc *Client) BlockNumber(ctx context.Context) (*big.Int, error) {
+	var blockNumber big.Int
+	err := sc.c.CallContext(ctx, &blockNumber, "starknet_blockNumber")
+	if err != nil {
+		return nil, err
+	}
```
tarrencev: nit: minimize err scope when possible
```suggestion
	if err := sc.c.CallContext(ctx, &blockNumber, "starknet_blockNumber"); err != nil {
		return nil, err
	}
```

jsonrpc/client.go
```diff
@@ -53,37 +53,75 @@ func (sc *Client) AccountNonce(context.Context, string) (*big.Int, error) {
...
+	return &blockNumber, err
+}
+
+func (sc *Client) LatestBlock(ctx context.Context, scope string) (*types.Block, error) {
```
tarrencev: Hmm I think the library should only implement the endpoints defined by the spec and abstractions should be left to the consumer.

jsonrpc/client.go
```diff
@@ -53,37 +53,75 @@ func (sc *Client) AccountNonce(context.Context, string) (*big.Int, error) {
...
-	panic("not implemented")
+func (sc *Client) TransactionByHash(ctx context.Context, hash string) (*types.Transaction, error) {
+	var tx types.Transaction
+	err := sc.fetchStructure(ctx, "starknet_getTransactionByHash", tx, hash)
+
+	return &tx, err
 }
 
-func (sc *Client) getBlock(ctx context.Context, method string, args ...interface{}) (*types.Block, error) {
+func (sc *Client) fetchStructure(ctx context.Context, method string, data interface{}, args ...interface{}) error {
```
tarrencev: nit: `fetchStructure` is a bit confusing. i get that structure is referring to the struct passed by reference but without understanding the code it isn't clear. i would just name something generic like `do`

```suggestion
func (sc *Client) do(ctx context.Context, method string, data interface{}, args ...interface{}) error {
```

jsonrpc/client.go
```diff
@@ -53,37 +53,75 @@ func (sc *Client) AccountNonce(context.Context, string) (*big.Int, error) {
...
+
+	return &block, err
 }
 
 func (sc *Client) BlockByNumber(ctx context.Context, number *big.Int, scope string) (*types.Block, error) {
-	return sc.getBlock(ctx, "starknet_getBlockByNumber", toBlockNumArg(number), scope)
+	var block types.Block
+	err := sc.fetchStructure(ctx, "starknet_getBlockByNumber", block, toBlockNumArg(number), scope)
+
+	return &block, err
```
tarrencev: fix: pass block by reference so unmarshal is done into the passed instance
fix: return nil on error, it makes the code more explicit and easier to understand
nit: minimize err scope


```suggestion
	if err := sc.fetchStructure(ctx, "starknet_getBlockByNumber", &block, toBlockNumArg(number), scope); err != nil {
		return nil, err
	}

	return &block, nil
```

jsonrpc/client.go
```diff
@@ -53,37 +53,75 @@ func (sc *Client) AccountNonce(context.Context, string) (*big.Int, error) {
...
+}
+
+func (sc *Client) CodeByAddress(ctx context.Context, address string) (*types.Code, error) {
+	var contract types.Code
+	err := sc.fetchStructure(ctx, "starknet_getCode", contract, address)
```
tarrencev: same as above

jsonrpc/client.go
```diff
@@ -53,37 +53,75 @@ func (sc *Client) AccountNonce(context.Context, string) (*big.Int, error) {
...
 
 func (sc *Client) Invoke(context.Context, types.Transaction) (*types.AddTxResponse, error) {
 	panic("not implemented")
 }
 
-func (sc *Client) TransactionByHash(context.Context, string) (*types.Transaction, error) {
-	panic("not implemented")
+func (sc *Client) TransactionByHash(ctx context.Context, hash string) (*types.Transaction, error) {
+	var tx types.Transaction
+	err := sc.fetchStructure(ctx, "starknet_getTransactionByHash", tx, hash)
```
tarrencev: same as above

jsonrpc/client.go
```diff
@@ -53,37 +53,69 @@ func (sc *Client) AccountNonce(context.Context, string) (*big.Int, error) {
...
-	return sc.getBlock(ctx, "starknet_getBlockByNumber", toBlockNumArg(number), scope)
+	var block types.Block
+	if err := sc.do(ctx, "starknet_getBlockByNumber", &block, toBlockNumArg(number), scope); err != nil {
+		return nil, err
+	}
+
+	return &block, nil
+}
+
+func (sc *Client) CodeByAddress(ctx context.Context, address string) (*types.Code, error) {
```
tarrencev: Lets align with the go-ethereum interface as much as possible: https://pkg.go.dev/github.com/obscuren/go-ethereum/ethclient?utm_source=godoc#Client.CodeAt

```suggestion
func (sc *Client) CodeAt(ctx context.Context, address string) (*types.Code, error) {
```

tarrencev (approved):

PR #29, opened 2022-06-17
feat: sn v0.9.0 support

drspacemn:
- TX Fee support
- Execute/ExecuteMulticall tested against:https://github.com/OpenZeppelin/cairo-contracts/commit/a1404d40dd6945c095ae1e0abdd37190228a928c
- Declare (contract class)
- GW ContractClassByHash, ContractClassAt


account.go
```diff
@@ -53,40 +55,71 @@ func (sc StarkCurve) NewSigner(private, pubX, pubY *big.Int, provider types.Prov
...
 
-	hash, err := signer.Curve.HashMulticall(address, nonce, maxFee, UTF8StrToBig(chainID), txs)
+	// provide good signature so we can get estimate for ECDSA signing
+	if fee, err := account.EstimateFeeMultiCall(ctx, req, nonce, calls); err == nil {
+		req.MaxFee = fee
+	}
+
+	if r, s, err := account.SignMultiCall(req.MaxFee, nonce, calls); err == nil {
+		req.Signature = []string{r.String(), s.String()}
+	}
```
tarrencev: hmm should we return an error if either of these fail?

tarrencev (approved):

account.go
gregoryguillou: I agree, that would make more sense to write:

```suggestion
        fee, err := account.EstimateFeeMultiCall(ctx, req, nonce, calls); 
	if err != nil {
		return nil, err
	}
	req.MaxFee = fee

	r, s, err := account.SignMultiCall(req.MaxFee, nonce, calls); 
	if err != nil {
		return nil, err
	}
	req.Signature = []string{r.String(), s.String()}
```

or if you want to *continue after an error* without the fee or without the signature for instance, you might want to check for the error. It could be that you have an `expected` error on devnet with the fee but if that is the case, you would have to perform a comparison with a "sentinel error" or with `errors.Is()`.

gateway/starknet.go
```diff
@@ -30,6 +30,9 @@ func (sg *Gateway) Call(ctx context.Context, call types.FunctionCall, blockHashO
 	if len(call.Calldata) == 0 {
 		call.Calldata = []string{}
 	}
+	if len(call.Signature) == 0 {
+		call.Signature = []string{}
+	}
```
gregoryguillou, line 35: I do not think it is useful. See https://go.dev/play/p/dM5IPzV97os for a playground. If Calldata is a `[]string` the default value is []string{}, no?

tarrencev: I believe the issue is a nil string slice will be omitted from the encoded json and gateway wants an empty string array

gregoryguillou: oh! I'll check that too.

gregoryguillou: You are right @tarrencev ! Sorry for the comment but I learned something.

account.go
drspacemn: Yes, more explicit. Will change

gateway/starknet.go
drspacemn: Correct, it’s pretty ugly but the api rejects the nil string signature slice. Open to a better default.

account.go
drspacemn: understood. As all '__execute__' functions must pay fees now I'll just return on bad fee estimate and signature

PR #30, opened 2022-06-17
proposal reduce the api surface

gregoryguillou:
Hello,

I do not expect you accept that PR. It provides a few examples to continue the talk. Maybe that would be better to coordinate on the starkware discord. I could not DM you. Anyway, long story short, a few proposal:

1. `SNValToBN` is actually almost the same as `(i *big.Int{}).SetString(z, 0)`. The 0 for 2nd parameter checks for `0x` and does a .`SetString(z, 16)` if it has one: If it does not... It works. I've added the change and a test to show it.
2. There is a concern with not checking the value is correct with `SNValToBN`. If it is in the test or to interact with some backend it is probably okay (though as an oddity `keys` in pathfinder starknet_getEvents is actually an hexa without 0x in 0.2.2-alpha). However if you expose the conversion in an API, if could lead to the value being nil and the API panicking, i.e. a DDoS.
3. You can probably remove all the `HexToBN` and `StrToBig` and replace them with `SNValToBN`. That would remove utilities and simplify the godoc.
4. Another proposal is simply to not expose `SNValToBN` by renaming it `snValToBN`. It would only require a few changes since, except for the example, there is almost no conversion
5. I also propose to remove the `replace` of the `go.mod` in your example with the go workspace clause because you already have set the environment to 1.18.


gregoryguillou:
I've kept `hexToBN` and `strToBig` but they are dead code.

drspacemn:
1) That is great! I must have missed it in the docs. I definitely like the change.

2/3/4) Any thoughts on moving a similar method to `types.Felt` i.e. `func ToFelt(str string) *Felt` and moving to use *types.Felt as a replacement for *big.Int (@tarrencev)?

5) Thank you this will be helpful

tarrencev:
Yeah! Agree on reducing api surface and migrating to `big.Int` to `Felt` and providing util functions there for conversion

gregoryguillou:
Great! I will split this PR in pieces so that we can discuss the points 1-by-1. Again, that would help to have some sort of direct channel because I have questions like the ones below:

- can we not provide some sort of abstraction so that the API can work with the geteway and pathfinder even if the API differ?
- I see you are using the jsonrpc implementation of go-ethereum. I did not figure out how to use named parameters (only positional) with that library and that is why I have been using github.com/ybbus/jsonrpc. Do you know a way to use named parameter with go-ethereum or do you know a way to use positional parameters with pathfinder ?

👍 to move to Felt

drspacemn:
"can we not provide some sort of abstraction so that the API can work with the geteway and pathfinder even if the API differ?"
Yes we expose the 'Provider' interface and both rpc and gateway clients normalize their responses.

I put this together just to see what it would look like with your `(i *big.Int{}).SetString(z, 0)` implementation:
- https://github.com/dontpanicdao/caigo/blob/feat/felt/types/felt.go
- https://github.com/dontpanicdao/caigo/blob/feat/felt/types/felt_test.go

drspacemn:
@gregoryguillou are you interested in being added as a maintainer of `caigo`?

gregoryguillou:
> @gregoryguillou are you interested in being added as a maintainer of `caigo`?

sure. I think if you do that I would be able to add the project to CircleCI. Thank you

gregoryguillou:
Since you have passed my PR with Go workspace and we will go for the `Felt` type. I do not think this PR is useful anymore. @drspacemn should should have a mail to approve the projet with circleci. Great!

PR #33, opened 2022-06-21
feat:sn v090 support

drspacemn:
- clean up Account interface
- only accept maxFee from caller, don't attempt from 'estimate_fee'
- JSON testing framework for starknet_test
- gateway 'FullContract'


examples/starknet/main.go
```diff
@@ -62,20 +62,31 @@ func main() {
...
-	execResp, err := account.Execute(context.Background(), types.Transaction{
-		ContractAddress:   tx.Transaction.ContractAddress,
-		EntryPointSelector: "increment",
-	})
+	increment := []types.Transaction{
+		types.Transaction{
+			ContractAddress:   tx.Transaction.ContractAddress,
+			EntryPointSelector: "increment",
+		},
+	}
```
gregoryguillou: go knows the content of the slice is `types.Transaction`
```suggestion
	increment := []types.Transaction{
		{
			ContractAddress:   tx.Transaction.ContractAddress,
			EntryPointSelector: "increment",
		},
	}
```

examples/starknet/main.go
```diff
@@ -62,20 +62,31 @@ func main() {
...
+
+	feeEstimate, err := account.EstimateFee(context.Background(), increment)
+	if err != nil {
+		panic(err.Error())
+	}
+	fee := &types.Felt{big.NewInt(int64(float64(feeEstimate.Amount) * 1.15))}
```
gregoryguillou: `go vet` suggests (not me 🤣) you "key" the property because what you are using on that case is a positional parameter set which is not related to the fact *big.Int is embedded. As a result, keyed affectation is more future-proof.
```suggestion
	fee := &types.Felt{Int: big.NewInt(feeEstimate.Amount * 115 / 100)}
```

I advise we `* 115 / 100` to avoid the float64 cast https://go.dev/play/p/OwwK8nCB5_-

gregoryguillou (approved):
👍 . I have tested the 2 examples and they work fine with me! I assume you have charged some eth to the account you are using. That might something to maintain.

examples/starknet/main.go
drspacemn: roger roger, will change

examples/starknet/main.go
drspacemn: Yup yup understood. I need to add `go vet` to my workflow.

drspacemn:
Not sure where else to put this, so FYI @gregoryguillou here is the discord channel invite: https://discord.gg/X7uEJG6j

gregoryguillou:
> Not sure where else to put this

I do not know either 👍  and **you** are the OG here. I would not be surprise someone pull the eth from the account, on goerli. But for now, it is just me overthinking. "premature optimization is the root of all evil".

PR #34, opened 2022-06-21
change examples to work with workspace

gregoryguillou:
Hello,

As I said, I'm splitting my PR in several so that it is easier to discuss it and not apply it if needed. This one is just the change from `replace` in the example go.mod to work with `go work` in the examples.

2 quick comments:
1. the change of Program from string to []string (i.e. 7f94c75ead83fa2398cc9d1722a4d1ab6178c179) in the class definition has broken the example. That is because when you deploy a program, it should be uploaded with a gzip/base64 content. As a result, program SHOULD BE a `string` in that case. It might not be possible (at least here) to use the same payload to input/output the API. Not to say that if we want to support the pathfinder API.
2. A simple build/test would have detected it. Maybe we should add the project in a CI. I would +1 for CircleCI. If you want I can build a pipeline.

TL;DR: this PR should be the simplest one.


drspacemn:
lgtm

drspacemn (approved):

PR #36, opened 2022-06-22
rename:jsonrpc

drspacemn:
- Rename 'jsonrpc' -> 'rpc'
- Implement and test rpc methods: syncing, class, classAt, classHashAt, Events
- GW client 'TransactionTrace' impl, stub for RPC client
- 'type ContractClass' -> 'program' changed to 'interface{}' so it can be used in multiple contexts


gregoryguillou (approved):
LGTM. Switching program into an interface might be a good enough solution for now.

Maybe you could create a list of what needs to be implemented to have a good coverage of the API and make caigo in 🥇 solution to interact with Starknet. I like the list a lot!

drspacemn:
Agreed on the list. We may spin up something in `github.com/starknet-edu` or `github.com/starkware-community-libs` for some sort of integration checklist and sync for all the major libs/SDKs

tarrencev:
Generally looks good to me! Imo I would prefer to have an explicit type for `program` but happy to roll with this for now

PR #37, opened 2022-06-24
add the the errorHandler option to the gateway

gregoryguillou:
My understanding is that we are missing the code that set the errorHandler from the options when instantiating a *Gateway with NewClient. I could have provided a default function that would be,`func(err error) error {return err}` but since the function is only used in `do` and there is a test to check the function exists, it can be avoided.

### DIFF
--- a/gateway/gateway.go
+++ b/gateway/gateway.go
@@ -73,6 +73,7 @@ func NewClient(opts ...Option) *Gateway {
 	}
 
 	sg.client = gopts.client
+	sg.errorHandler = gopts.errorHandler
 
 	return sg
 }
--- a/gateway/gateway.go
+++ b/gateway/gateway.go
@@ -73,7 +73,7 @@ func NewClient(opts ...Option) *Gateway {
 	}
 
 	sg.client = gopts.client
-	// sg.errorHandler = gopts.errorHandler
+	sg.errorHandler = gopts.errorHandler
 
 	return sg
 }
--- a/gateway/gateway.go
+++ b/gateway/gateway.go
@@ -73,7 +73,7 @@ func NewClient(opts ...Option) *Gateway {
 	}
 
 	sg.client = gopts.client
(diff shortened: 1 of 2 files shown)


gateway/gateway.go
```diff
@@ -73,6 +73,7 @@ func NewClient(opts ...Option) *Gateway {
 	}
 
 	sg.client = gopts.client
+	sg.errorHandler = gopts.errorHandler
```
drspacemn, line 76: The opts can be init 'WithErrorHandler' 

https://github.com/dontpanicdao/caigo/blob/97acd14475facad663dd11a36b2aeb4c741fa193/gateway/opts.go#L47

gregoryguillou:
There is something I do not understand. I have seen you should use `WithErrorHandler` to register and get an option. Yet I it should not be enough. What do I miss in that case?

1. I have revert (commented) my change
2. I have created a test called TestWithErrorHandler. You can see it is failing:

```shell
go test -v -run TestWithErrorHandler
=== RUN   TestWithErrorHandler
    opts_test.go:28: spy should be true, got false
--- FAIL: TestWithErrorHandler (0.53s)
FAIL
exit status 1
FAIL	github.com/dontpanicdao/caigo/gateway	0.732s
```

Now if you set my change back, you see the test is succeeding:

```shell
go test -v -run TestWithErrorHandler
=== RUN   TestWithErrorHandler
spy triggered
--- PASS: TestWithErrorHandler (0.52s)
PASS
ok  	github.com/dontpanicdao/caigo/gateway	0.803s
```

Can you check again and explain what I am missing?

gateway/gateway.go
gregoryguillou: Can you please check the comment below?

drspacemn:
Understood. `WithError` was applied to the `opts` and not the actual gateway client

drspacemn (approved):

PR #38, opened 2022-06-24
feat: common Felt type 

drspacemn:
Implement https://github.com/dontpanicdao/caigo/pull/30 proposal on 'types.Felt' along with basic Felt util methods.

Goal is to move many '*big.Int' types to '*Felt' types and achieving goals laid out in https://github.com/dontpanicdao/caigo/pull/30


types/felt.go
```diff
@@ -7,106 +7,147 @@ import (
...
 )
 
+const (
+	FIELD_GEN   int    = 3
+	FIELD_PRIME string = "3618502788666131213697322783095070105623107215331596699973092056135872020481"
+)
+
+var (
+	MaxFelt   = StrToFelt(FIELD_PRIME)
+	utfRegexp = regexp.MustCompile(`\w+`)
```
gregoryguillou: That is not clear to me what short string should support, yet:
 
- that cannot be utf strings because utf is a multi-bytes format for instance utf8 is variable size format that can include 32bit characters and it would not fit the max size in that case.
- if that was something other than ascii, it would have to define the charset too (e.g. latin9 or other) . So I suppose that is ascii
- next come the problem of non graphic characters, like backspace. It think the pattern for the text should be only graphic or space characters (including \n for instance) up to 31.
- @ or whitespace should be in and are not supported with \w 
- empty strings would be nil anyway

So I propose the regexp below with a different variable name even if that is not accurate either:

```suggestion
	asciiRegexp = regexp.MustCompile(`^([[:graph:]]|[[:space:]]){1,31}$`)
```

see https://go.dev/play/p/fcYvD7bAd7A my tests

types/felt.go
```diff
@@ -7,106 +7,147 @@ import (
...
-func (f *Felt) Scan(src interface{}) error {
-	var i sql.NullString
-	if err := i.Scan(src); err != nil {
-		return err
-	}
-	if !i.Valid {
-		return nil
+// Big converts a Felt to its big.Int representation.
+func (f *Felt) Big() *big.Int {
+	return new(big.Int).SetBytes(f.Int.Bytes())
```
gregoryguillou, line 32: The code should work, no?
```suggestion
	return f.Int
```

Unless you do not want to expose the Int which I do not know how to do, you probably do not need that method because you can access the underlying field fron the outside

types/felt.go
```diff
@@ -7,106 +7,147 @@ import (
...
-	// Value came in a floating point format.
-	if strings.ContainsAny(i.String, ".+e") {
-		flt := big.NewFloat(0)
-		if _, err := fmt.Sscan(i.String, f); err != nil {
-			return err
+	if utfRegexp.MatchString(str) {
+		hexStr := hex.EncodeToString([]byte(str))
+		if b, ok := new(big.Int).SetString(hexStr, 16); ok {
+			f.Int = b
+			return ok
```
gregoryguillou: Even if I think that is a very good idea. I can think of a case where it will not work and that is when we have an array of shorts (i.e. len/Felt*) to represent a longer string and it as a member that is made of number only. There is such a case in the way aspect.co has built their NFT tokenURI. I think we may want to have a specific set of methods (or a  parameter) to manage short strings.

types/felt_test.go
```diff
@@ -0,0 +1,106 @@
...
+)
+
+var (
+	rawTest string = `{
+		"max_felt": 3618502788666131213697322783095070105623107215331596699973092056135872020481,
+		"felts": [
+			{"value": "0x0", "expected": 0},
+			{"value": "0x1277312773", "expected": 79309121395},
+			{"value": "0xbBbBBBBbbBBBbbbBbbBbbbbBBbBbbbbBbBbbBBbB", "expected": "1071767867375995473349368877325274214414350531515"},
+			{"value": "2458502865976494910213617956670505342647705497324144349552978333078363662855", "expected": "2458502865976494910213617956670505342647705497324144349552978333078363662855"}
```
gregoryguillou, line 21: You could add shorts as part of the test and test that it will fail if shorts are longer than 31 characters or includes non supported characters.

gregoryguillou (approved):
The idea of `quietly` managing short strings is good but I'm afraid it will not work for a few edge cases. Even if that might be more painful, I would prefer we not do that. Or if we do, because I also prefer you push that change, leave a `TODO:` comment to make sure we track the issue and we talk about it later. Except for that, it is good to me.

types/felt.go
drspacemn: Agreed, the new regex is a good change

types/felt.go
drspacemn: This method is more for avoiding a shallow copy of the pointer and instead getting a new big Int to manipulate. Maybe needs a more descriptive method name

types/felt.go
drspacemn: I’m not opposed to a specific short string method. Can you point me to the aspect tokenURI code. I’m not sure I understand why this in conjunction with your ascii regex change wouldn’t do the trick over each array element.

gregoryguillou: This is the line on the aspectco repository https://github.com/aspectco/cairo-contracts/blob/48638b84fb79b02734e0e47d2ef230d57399a212/openzeppelin/utils/ShortString.cairo#L89

The idea is that they return an array of Felts that represents the tokenURI for the NFT. The last 2 Felts are actually a low/high or an uint256 transformed in short-string, thus the `uint256_to_ss` function. TL;DR: the 2 Felts only contain numbers, yet they are short strings and should be translated as such and not as *big.Int

gregoryguillou (approved):

types/felt.go
drspacemn: So in this case `ToFelt` would just treat the low and high like a big.Int and parse that representation which would be expected, correct?

I could add another function for `ShortString` giving the string representation, but I'm not sure I understand why the `ToFelt` wouldn't act as expected in this case.

PR #39, opened 2022-06-25
Improve tests to support multiple RPC API

gregoryguillou:
## Objectives

The idea behind this PR is to extend the test coverage of the RPC API and test functions 1-by-1 to better assert the output:

1. testnet and mainnet are not always in-sync, we should have a set of tests for the 2 of them just by changing some configurations.
2. there is also a simple mock to use as a default test so that we can (1) code what we should expect and (2) have a fast unit test environment. To do that I have replaced the Client.client to an interface that implements the required method from go-ethereum/*rpc.Client
3. We should be able to test different API as some points. For instance, we should also be able to test against potential future API like [juno](https://gojuno.xyz/docs/features/rpc) and [devnet](https://github.com/Shard-Labs/starknet-devnet). The PR will ease that change even if, for now nothing is implemented. Btw, is there a place that keeps track of existing APIs?

## How to improve the tests

The next stage would be to detect change in the API output. We could store some sort of snapshots of the output and compare them with something similar to `reflect.DeepCompare()`. This would probably reduce the code necessary for the test and detect a change in pathfinder or devnet payload. What do you think? 

## `go test -env`

There is now a `-env` flag to `go test` in the `rpc` directory. The dataset used for the tests is selected base on this flag. If you want to point to different environment you can add `.env.testnet` and `.env.mainnet` and set `INTEGRATION_BASE=http://localhost:9545` or whatever URL you need. 

- To run a test on the mainnet run:

```shell
cd rpc
go test -v -env mainnet .
```

- To run a test on the testnet run:

```shell
cd rpc
go test -v -env testnet .
```

- To run a test with the mocks, run:

```shell
cd rpc
go test -v -env mock .
# or because it is the default environment
go test -v .
```


rpc/client_test.go
```diff
@@ -2,150 +2,160 @@ package rpc
...
-		t.Errorf("Could not retrieve transaction receipts: %v\n", block.Transactions[0])
-	}
+	os.Exit(m.Run())
+}
 
-	_, err = client.BlockByNumber(context.Background(), num, "FULL_TXNS")
-	if err != nil {
-		t.Errorf("Could not retrieve block: %v\n", err)
+// beforeEach checks the configuration and initializes it before running the script
+func beforeEach(t *testing.T) *testConfiguration {
```
tarrencev, line 58:
```suggestion
func beforeEach(t *testing.T) *testConfiguration {
    t.Helper()
```

rpc/client_test.go
```diff
@@ -2,150 +2,160 @@ package rpc
...
+		testConfig.base = base
 	}
-
-	_, err = client.BlockByHash(context.Background(), block.BlockHash, "FULL_TXNS")
+	client, err := DialContext(context.Background(), testConfig.base)
 	if err != nil {
-		t.Errorf("Could not retrieve block: %v\n", err)
+		t.Fatal("connect should succeed, instead:", err)
 	}
+	testConfig.client = client
```
tarrencev, line 79: Thoughts on encapsulating the teardown here rather than in each caller?
```suggestion
	testConfig.client = client
	t.Cleanup(func() {
		testConfig.client.Close()
	})
```

tarrencev (approved):
minor suggestions but generally looks great!

rpc/client_test.go
gregoryguillou: I did not know that. That makes a lot of sense. I've pushed the change. I do not understand your comment does not get outdated by it (I also removed the defer from everywhere).

drspacemn (approved):
looks great

PR #40, opened 2022-06-26
Implement and test starknet_addDeployTransaction

gregoryguillou:
This PR implements the RPC starknet_addDeployTransaction verb as documented in [starknet_write_api.json](https://github.com/starkware-libs/starknet-specs/blob/master/api/starknet_write_api.json):

- The function is called `AddDeployTransaction` to match the manifest
- There are 3 input parameters that are contractAddressSalt constructorCallData and contractDefinition
- the AddDeployTransactionOutput type has been created to manage the output.

The PR provides one tests for the testnet as well as a mock to unit test the associated method. For mainnet, only whitelisted contracts can be created from this [conversation](https://discord.com/channels/793094838509764618/793094838987128844/990692360608444508). I have left a comment in the test to implement it later.


rpc/write.go
```diff
@@ -14,6 +24,39 @@ func (sc *Client) Declare(context.Context, types.Transaction) (*types.AddTxRespo
...
+		program, err = encodeProgram(data)
+		if err != nil {
+			return nil, err
+		}
+	}
+	contractDefinition.Program = program
+
+	var result AddDeployTransactionOutput
+	err := sc.do(ctx, "starknet_addDeployTransaction", &result, contractAddressSalt, constructorCallData, contractDefinition)
+
```
tarrencev: nit: prefer to be explicit here with the err return
```suggestion
	if err := sc.do(ctx, "starknet_addDeployTransaction", &result, contractAddressSalt, constructorCallData, contractDefinition); err != nil {
		return nil, err
	}

```

rpc/api_addtransaction_test.go
```diff
@@ -0,0 +1,73 @@
...
+		Contract                types.ContractClass
+		ConstructorCallData     []string
+		ExpectedTransactionHash string
+		ExpectedContractAddress string
+	}
+	var contract types.ContractClass
+	err := json.Unmarshal(counterFile, &contract)
+	if err != nil {
+		t.Fatal("error loading contract:", err)
+	}
```
tarrencev: nit: limit err scope
```suggestion
	if err := json.Unmarshal(counterFile, &contract); err != nil {
		t.Fatal("error loading contract:", err)
	}
```

tarrencev (approved):
couple nits but lgtm!

drspacemn:
Overall looks great. 

Looks like a build artifact got pushed: `rpc/tests/counter.json`

gregoryguillou:
> Looks like a build artifact got pushed: `rpc/tests/counter.json`

yes. I've added it so that we can test the deployment on testnet. If you have a better idea, let me know.

PR #41, opened 2022-06-26
implement and test starknet_getStorageAt

gregoryguillou:
This PR implements the starknet_getStorageAt verb as documented in [starknet_api_openrpc.json](https://github.com/starkware-libs/starknet-specs/blob/master/api/starknet_api_openrpc.json):

The function is called `StorageAt` to match the manifest:
- There are 3 input parameters that are contractAddress, key, blockHashOrTag. They all are strings.
- the key parameter is encoded by the method and you should pass the name key, like `_signer` or `balance`
- The output is the value from the @storage_var return in hexa with the `0x` prefix

### DIFF
--- a/rpc/api.go
+++ b/rpc/api.go
@@ -3,6 +3,7 @@ package rpc
 import (
 	"context"
 	"encoding/json"
+	"fmt"
 	"math/big"
 
 	"github.com/dontpanicdao/caigo"
@@ -124,6 +125,14 @@ func (sc *Client) ClassHashAt(ctx context.Context, address string) (string, erro
 	}
 
 	return *result, nil
+}
+
+// GetStorageAt gets the value of the storage at the given address and key.
+func (sc *Client) GetStorageAt(ctx context.Context, contractAddress, key, blockHashOrTag string) (string, error) {
+	var value string
+	hashKey := fmt.Sprintf("0x%s", caigo.GetSelectorFromName(key).Text(16))
+	err := sc.do(ctx, "starknet_getStorageAt", &value, contractAddress, hashKey, blockHashOrTag)
+	return value, err
 }
 
 // TransactionByHash gets the details and status of a submitted transaction.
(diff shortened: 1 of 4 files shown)


rpc/api.go
```diff
@@ -126,6 +127,14 @@ func (sc *Client) ClassHashAt(ctx context.Context, address string) (string, erro
 	return *result, nil
 }
 
+// GetStorageAt gets the value of the storage at the given address and key.
+func (sc *Client) GetStorageAt(ctx context.Context, contractAddress, key, blockHashOrTag string) (string, error) {
```
tarrencev: https://go.dev/doc/effective_go#Getters
```suggestion
func (sc *Client) StorageAt(ctx context.Context, contractAddress, key, blockHashOrTag string) (string, error) {
```

rpc/api.go
```diff
@@ -126,6 +127,14 @@ func (sc *Client) ClassHashAt(ctx context.Context, address string) (string, erro
...
+	var value string
+	hashKey := fmt.Sprintf("0x%s", caigo.GetSelectorFromName(key).Text(16))
+	err := sc.do(ctx, "starknet_getStorageAt", &value, contractAddress, hashKey, blockHashOrTag)
+	return value, err
```
tarrencev: prefer to be explicit with return values. it allows the reader to more confidently make assumptions
```suggestion
		if err := sc.do(ctx, "starknet_getStorageAt", &value, contractAddress, hashKey, blockHashOrTag); err != nil {
		return nil, err
	}

	return value, nil
```

tarrencev (approved):

rpc/api.go
```diff
@@ -126,12 +127,24 @@ func (sc *Client) ClassHashAt(ctx context.Context, address string) (string, erro
 	return *result, nil
 }
 
+// GetStorageAt gets the value of the storage at the given address and key.
```
drspacemn, line 130: godoc comment update to reflect the change to 'StorageAt'

drspacemn (approved):

rpc/api.go
drspacemn: change applied to: https://github.com/dontpanicdao/caigo/pull/43

PR #42, opened 2022-06-27
Implement starknet_getstateupdatebyhash

gregoryguillou:
`starknet_getstateupdatebyhash` is not yet implemented in pathfinder. I have coded the output as described in [starknet_api_openrpc.json](https://github.com/starkware-libs/starknet-specs/blob/master/api/starknet_api_openrpc.json) and created a mock with a test. I think we can push the PR as is and complete it once it can be fully implemented.

### DIFF
--- a/rpc/api.go
+++ b/rpc/api.go
@@ -138,6 +138,64 @@ func (sc *Client) StorageAt(ctx context.Context, contractAddress, key, blockHash
 	return value, nil
 }
 
+// StorageDiff is a change in a single storage item
+type StorageDiff struct {
+	// ContractAddress is the contract address for which the state changed
+	ContractAddress string `json:"address"`
+	// Key returns the key of the changed value
+	Key string `json:"key"`
+	// Value is the new value applied to the given address
+	Value string `json:"value"`
+}
+
+// ContractItem is a new contract added as part of the new state
+type ContractItem struct {
+	// ContractAddress is the address of the contract
+	Address string `json:"address"`
+	// ContractHash is the hash of the contract code
+	ContractHash string `json:"contract_hash"`
+}
+
+// Nonce is a the updated nonce per contract address
+type Nonce struct {
+	// ContractAddress is the address of the contract
+	ContractAddress string `json:"contract_address"`
+	// Nonce is the nonce for the given address at the end of the block"
+	Nonce string `json:"nonce"`
+}
+
+// StateDiff is the change in state applied in this block, given as a
+// mapping of addresses to the new values and/or new contracts.
+type StateDiff struct {
+	// StorageDiffs list storage changes
+	StorageDiffs []StorageDiff `json:"storage_diffs"`
+	// Contracts list new contracts added as part of the new state
+	Contracts []ContractItem `json:"contracts"`
+	// Nonces provides the updated nonces per contract addresses
+	Nonces []Nonce `json:"nonces"`
+}
+
+type GetStateUpdateOutput struct {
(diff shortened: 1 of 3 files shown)


rpc/api.go
```diff
@@ -135,6 +135,64 @@ func (sc *Client) GetStorageAt(ctx context.Context, contractAddress, key, blockH
 	return value, err
 }
 
+// StorageDiff is a change in a single storage item
+type StorageDiff struct {
+	// ContractAddress is the contract address for which the state changed
+	ContractAddress string `json:"address"`
```
tarrencev: I think we should try align with rpc naming, even tho it is a bit inconsistent itself

```suggestion
	Address string `json:"address"`
```

rpc/api.go
```diff
@@ -135,6 +135,64 @@ func (sc *Client) GetStorageAt(ctx context.Context, contractAddress, key, blockH
...
+	// StateDiff is the change in state applied in this block, given as a
+	// mapping of addresses to the new values and/or new contracts.
+	StateDiff StateDiff `json:"state_diff"`
+}
+
+// GetStateUpdateByHash gets the information about the result of executing the requested block.
+func (sc *Client) GetStateUpdateByHash(ctx context.Context, blockHashOrTag string) (*GetStateUpdateOutput, error) {
+	var result GetStateUpdateOutput
+	err := sc.do(ctx, "starknet_getStateUpdateByHash", &result, blockHashOrTag)
+	return &result, err
```
tarrencev:
```suggestion
	if err := sc.do(ctx, "starknet_getStateUpdateByHash", &result, blockHashOrTag); err != nil {
		return nil, err
	}
	return &result, nil
```

rpc/api.go
```diff
@@ -135,6 +135,64 @@ func (sc *Client) GetStorageAt(ctx context.Context, contractAddress, key, blockH
...
+	OldRoot string `json:"old_root"`
+	// AcceptedTime is when the block was accepted on L1.
+	AcceptedTime int `json:"accepted_time"`
+	// StateDiff is the change in state applied in this block, given as a
+	// mapping of addresses to the new values and/or new contracts.
+	StateDiff StateDiff `json:"state_diff"`
+}
+
+// GetStateUpdateByHash gets the information about the result of executing the requested block.
+func (sc *Client) GetStateUpdateByHash(ctx context.Context, blockHashOrTag string) (*GetStateUpdateOutput, error) {
```
tarrencev: https://go.dev/doc/effective_go#Getters

```suggestion
func (sc *Client) StateUpdateByHash(ctx context.Context, blockHashOrTag string) (*StateUpdateOutput, error) {
```

rpc/api.go
```diff
@@ -135,6 +135,64 @@ func (sc *Client) GetStorageAt(ctx context.Context, contractAddress, key, blockH
...
+type StateDiff struct {
+	// StorageDiffs list storage changes
+	StorageDiffs []StorageDiff `json:"storage_diffs"`
+	// Contracts list new contracts added as part of the new state
+	Contracts []ContractItem `json:"contracts"`
+	// Nonces provides the updated nonces per contract addresses
+	Nonces []Nonce `json:"nonces"`
+}
+
+type GetStateUpdateOutput struct {
```
tarrencev:
```suggestion
type StateUpdateOutput struct {
```

tarrencev (approved):
same minor feedback but generally great work

drspacemn (approved):
lgtm

PR #45, opened 2022-06-27
create a simple github action to run `go test` on rpc

gregoryguillou:


.github/workflows/go.yml
```diff
@@ -0,0 +1,25 @@
+name: Go
+
+on:
+  pull_request:
+    branches: [ master ]
+    types: [ opened, synchronize, closed ]
+  push:
+    branches-ignore:
+      - master
```
tarrencev: I'm curious if we need both the push and pr rules. It seems like the push rule should cover everything pr does? We might want to run on push to master too since it can be possible to pass on a branch and fail in master

tarrencev (approved):
awesome contribution!

.github/workflows/go.yml
gregoryguillou: Oh... so you think merging the PR would trigger a push? tbh, I don't know. What I'll do is remove everything related to PR and we will add them back if something does not work as expected 🤞

tarrencev: i would just copy from another project like: https://github.com/ent/ent/blob/master/.github/workflows/ci.yml

gregoryguillou: We'll see 😃 what pull_request is actually useful for... now that I have disabled it.

drspacemn: let's give it a shot!

PR #46, opened 2022-06-27
Implement and test starknet_estimateFee

gregoryguillou:
This PR implements and tests EstimateFee with RPC. Note that types.FunctionCall does not have all the properties requires by Pathfinder, including MaxFee and Version. I did not find an example with testnet for now.


drspacemn (approved):
Tests, lgtm

drspacemn:
I think we should run `estimateFee` and `addTransaction` throught `types.Transaction`. 

This will also let Gateway/RPC implement the same interface:
```
func (sg *Gateway) EstimateFee(ctx context.Context, tx types.Transaction) (*types.FeeEstimate, error) {
```

PR #48, opened 2022-06-28
add a test to detect when methods are implemented

gregoryguillou:
`TestNotImplemented` is a test that succeed while the method listed remain unimplemented. This test will, once we setup a nightly run help with detecting a new implementation of a method with pathfinder.


tarrencev (approved):
great idea!

PR #49, opened 2022-06-28
Provide a mock to test and multi-env tests on the `gateway` directory

gregoryguillou:
This PR adds the `-env` to the tests in the `gateway` directory. This way we should be able to:

- test with a mock
- run the same set of tests to different environment with just a flag, including testnet/mainnet or devnet

This will be useful to configure the CI step by step.


tarrencev (approved):
lgtm

drspacemn (approved):
lgtm

Issue #54, opened 2022-06-28
add support and test for `devnet` with the RPC API 


gregoryguillou:
## Objectives
`devnet` can be used as a local configuration to speed-up tests. From its documentation, it supports the [RPC API](https://github.com/Shard-Labs/starknet-devnet#json-rpc-api). The objectives of this issue are to populate the set of tests for the `rpc` directory so that they work with devnet. If the exact same configuration cannot be used because `devnet` is empty when starting, adapt the tests so that every rpc method are tested.

## Acceptance criteria
- [ ] every API call from the `rpc` directory should be implemented on devnet, except for those that are not supported yet

## Additional Notes
We would need #78 to be implemented before

gregoryguillou:
We have tests with devnet in the CI

PR #65, opened 2022-07-03
add devnet to github actions and add tests

gregoryguillou:
This PR contributes to #54:

- [x] adds a sidecar the the github action that embeds devnet on 5050
- [x] run a testing script to check the sidecar is up before running tests
- [x] make sure we run some tests on devnet at least with chainid and protocol.

This PR has enabled us to discover issue #66. We now have a devnet on the CI

### DIFF
--- a/.github/workflows/go.yml
+++ b/.github/workflows/go.yml
@@ -11,7 +11,7 @@ jobs:
 
     services:
       devnet:
-        image: shardlabs/starknet-devnet:0.2.4
+        image: shardlabs/starknet-devnet:latest
         ports:
         - 5050:5050
 
--- a/.github/workflows/go.yml
+++ b/.github/workflows/go.yml
@@ -8,6 +8,13 @@ on:
 jobs:
   build:
     runs-on: ubuntu-latest
+
+    services:
+      devnet:
+        image: shardlabs/starknet-devnet:0.2.4
+        ports:
+        - 5050:5050
+
     steps:
     - uses: actions/checkout@v3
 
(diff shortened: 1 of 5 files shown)


rpc/client_test.go
```diff
@@ -91,7 +91,7 @@ func TestChainID(t *testing.T) {
 		ChainID string
 	}
 	testSet := map[string][]testSetType{
-		"devnet":  {{ChainID: "DEVNET"}},
+		"devnet":  {{ChainID: "SN_GOERLI"}},
```
gregoryguillou, line 94: I would think it is a bug on starknet-devnet. I  have open a ticket https://github.com/Shard-Labs/starknet-devnet/issues/146

rpc/client.go
```diff
@@ -58,27 +58,26 @@ func (sc *Client) Close() {
 // ChainID retrieves the current chain ID for transaction replay protection.
 func (sc *Client) ChainID(ctx context.Context) (string, error) {
 	var result string
-	err := sc.c.CallContext(ctx, &result, "starknet_chainId")
+	err := sc.c.CallContext(ctx, &result, "starknet_chainId", []interface{}{}...)
```
gregoryguillou, line 63: That is the solution I have found to address #66 . TL;DR is that we would have to talk to the people at Shardlabs to see if that is a bug and can be corrected. At least I did not face the issue with Pathfinder.

.github/workflows/go.yml
```diff
@@ -8,6 +8,13 @@ on:
 jobs:
   build:
     runs-on: ubuntu-latest
+
+    services:
+      devnet:
+        image: shardlabs/starknet-devnet:0.2.4
```
drspacemn: What's your opinion on pulling latest here instead of v0.2.4?

gregoryguillou: I have fixed

drspacemn (approved):

PR #67, opened 2022-07-06
refactor: conform rpc/gateway to provider interface

tarrencev:
modifications to more closely match jsonrpc spec and enforce that rpc/gateway implement the provider interface


gateway/account.go
```diff
@@ -25,9 +25,9 @@ func (sg *Gateway) AccountNonce(ctx context.Context, address string) (*big.Int,
 	return caigo.HexToBN(resp[0]), nil
 }
 
-func (sg *Gateway) EstimateFee(ctx context.Context, tx types.Transaction) (*types.FeeEstimate, error) {
-	tx.EntryPointSelector = caigo.BigToHex(caigo.GetSelectorFromName(tx.EntryPointSelector))
-	req, err := sg.newRequest(ctx, http.MethodPost, "/estimate_fee", tx)
+func (sg *Gateway) EstimateFee(ctx context.Context, call types.FunctionCall, _ string) (*types.FeeEstimate, error) {
```
drspacemn: Any merit in exposing the last arg and defaulting to:

```
	appendQueryValues(req, url.Values{
		"blockHash": []string{hash},
	})
```

gateway/account.go
```diff
@@ -25,9 +25,9 @@ func (sg *Gateway) AccountNonce(ctx context.Context, address string) (*big.Int,
...
+	call.EntryPointSelector = caigo.BigToHex(caigo.GetSelectorFromName(call.EntryPointSelector))
+	req, err := sg.newRequest(ctx, http.MethodPost, "/estimate_fee", call)
```
drspacemn: Looks like `FunctionCall` doesn't serialize the signature field
```
starknet_test.go:117: testnet: could not estimate fee for transaction: 500: StarkErrorCode.MALFORMED_REQUEST __init__() missing 1 required positional argument: 'signature'
```

gateway/class.go
```diff
@@ -22,24 +22,24 @@ func (sg *Gateway) ClassByHash(ctx context.Context, hash string) (*RawContractDe
...
-	if opts != nil {
-		vs, err := query.Values(opts)
-		if err != nil {
-			return "", err
-		}
-		appendQueryValues(req, vs)
-	}
-
 	var resp string
-	return resp, sg.do(req, &resp)
```
drspacemn, line 44: Suggest using the types.Felt Unmarshal:
```
var resp types.Felt
return &resp, sg.do(req, &resp)
```

gateway/account.go
tarrencev: i think it should work now but am not able to verify locally

gateway/starknet.go
```diff
@@ -44,9 +44,6 @@ func (sg *Gateway) Call(ctx context.Context, call types.FunctionCall, blockHashO
 	if len(call.Calldata) == 0 {
 		call.Calldata = []string{}
 	}
-	if len(call.Signature) == 0 {
-		call.Signature = []string{"0", "0"} // allows rpc and http clients to implement(has to be a better way)
```
drspacemn, line 48: It looks like this is the culprit of the error: 
`500: StarkErrorCode.MALFORMED_REQUEST __init__() missing 1 required positional argument: 'signature'`

So the `account.AccountNonce` call fails in `EstimateFee`

gateway/account.go
```diff
@@ -25,9 +26,20 @@ func (sg *Gateway) AccountNonce(ctx context.Context, address string) (*big.Int,
...
+	tx.ContractAddress = call.ContractAddress
+	tx.Signature = []string{}
+
+	if hash != "" {
+		appendQueryValues(req, url.Values{
+			"blockHash": []string{hash},
+		})
+	}
+
+	req, err := sg.newRequest(ctx, http.MethodPost, "/estimate_fee", tx)
```
drspacemn: needs to be declared before line 37

tarrencev: ah thanks, my bad

PR #71, opened 2022-07-13
feat: embed pedersen params

tarrencev:
uses go:embed to embed the pedersen params and improves the ergonomics of the library. we might consider moving `Curve` to a `curve` package so the api can be e.g. `curve.Sign(...)`

### DIFF
--- a/account.go
+++ b/account.go
@@ -2,7 +2,6 @@ package caigo
 
 import (
 	"context"
-	"fmt"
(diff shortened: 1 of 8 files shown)


drspacemn:
I was thinking of ways to init the curve cleanly. This is awesome!

drspacemn (approved):

PR #83, opened 2022-08-09
feat: migrate gateway to Felt

internnos:
migrated relevant datatypes in `types/Transactions` to `Felt`

please let me know if there are any suggestions

thank you


tarrencev:
@internnos thanks for proposing this! One goal we have for caigo is to normalize the `gateway` and `rpc` interfaces, since eventually gateway will be deprecated. Unfortunately, gateway expects differently formatted inputs. IIRC, rpc expects everything hex encoded and gateway is a combination of hex/decimal.

One idea we've had is to accept a generic `Felt` type and handle the marshaling internally in the gateway/rpc implementation. `Felt` would marshal to hex by default and `gateway` will wrap it to marshal certain params to decimal. Ideally the consumer of caigo doesn't need to care about whether they are hitting the gateway or rpc underneath. What do you think of this approach?

internnos:
> @internnos thanks for proposing this! One goal we have for caigo is to normalize the `gateway` and `rpc` interfaces, since eventually gateway will be deprecated. Unfortunately, gateway expects differently formatted inputs. IIRC, rpc expects everything hex encoded and gateway is a combination of hex/decimal.
> 
> One idea we've had is to accept a generic `Felt` type and handle the marshaling internally in the gateway/rpc implementation. `Felt` would marshal to hex by default and `gateway` will wrap it to marshal certain params to decimal. Ideally the consumer of caigo doesn't need to care about whether they are hitting the gateway or rpc underneath. What do you think of this approach?

aah i see

so i'd like to confirm. does it mean that the following types should be in `Felt`?

- nonce
- calldata
- signature
- ContractAddress
- SenderAddress
- MaxFee
- EntryPointSelector
- TransactionHash

if that's the case, i think i get it!

thanks for the response! @tarrencev

tarrencev:
I'm not sure about all off the top of my head but the rpc spec is documented here: https://github.com/starkware-libs/starknet-specs/blob/master/api/starknet_api_openrpc.json

internnos:
> I'm not sure about all off the top of my head but the rpc spec is documented here: https://github.com/starkware-libs/starknet-specs/blob/master/api/starknet_api_openrpc.json

thanks! i've checked it

internnos:
as per now, I think almost all the datatypes have been migrated to `*Felt`

the one missing are `TransactionHash` and `ClassHash`

however some test cases are still broken and also there's still conflicting files so i'm still patching things up

@tarrencev @gregoryguillou

internnos:
currently `all gateway tests passed` with command `go test -v ./gateway -env testnet`

only one test still broken in `rpc/api_call_test.go/TestCall`

gregoryguillou:
what I have done is create a branch [feature/issue-75-felt](https://github.com/dontpanicdao/caigo/tree/feature/issue-75-felt) that we will use to target all these changes. This way I will be able to stage your changes and let you restart another branch to update the rpc without actually having to re-read everything. In addition, I will squash the commits to make the change easier to read... @internnos

account.go
```diff
@@ -137,7 +137,7 @@ func (account *Account) fmtExecute(ctx context.Context, calls []types.Transactio
 	req := types.FunctionInvoke{
 		FunctionCall: types.FunctionCall{
 			ContractAddress:    account.Address,
-			EntryPointSelector: EXECUTE_SELECTOR,
+			EntryPointSelector: types.StrToFelt(EXECUTE_SELECTOR),
```
gregoryguillou, line 140: It does not work that way... When dealing with entrypoints named `function`, starknet uses `keccak256(function) & (250bits)` as you can see from `GetSelectorFromName`. You cannot take a string a bind it to a Felt like that. What I suggess for now is you leave `EntryPointSelector` in `types FunctionCall` a string. Even though I am pretty sure it might contain the `GetSelectorFromName` of the function

account.go
```diff
@@ -157,22 +157,22 @@ func (account *Account) fmtExecute(ctx context.Context, calls []types.Transactio
 	return &req, nil
 }
 
-func fmtExecuteCalldataStrings(nonce *big.Int, calls []types.Transaction) (calldataStrings []string) {
+func fmtExecuteCalldataStrings(nonce *big.Int, calls []types.Transaction) (calldataStrings []*types.Felt) {
```
gregoryguillou, line 160: From a check, this function is (1) not exported because it starts with a lowercase `f` AND (2) only used once in the original code which should change because transaction will be Felts too. My guess is that it does not make sense anymore

PR #84, opened 2022-08-09
added ExecuteWithNonce to account

broody:
Hi! I needed a way to explicitly set nonce per execution, please let me know if there's an alternative method or other best practices. Thanks!

### DIFF
--- a/account.go
+++ b/account.go
@@ -2,6 +2,7 @@ package caigo
 
 import (
 	"context"
+	"fmt"
 	"math/big"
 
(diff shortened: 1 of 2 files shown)


tarrencev:
Hm i wonder if we should take an `ExecuteDetails` struct similar to starknet.js with nillable values. If they are nil, we handle them internally.

```
type ExecutionDetails struct {
   maxGas: *Felt,
   nonce: *Felt,
   version: uint
   ...
}
```

PR #85, opened 2022-08-09
added details param to Execute()

broody:
@tarrencev what about something like this?

### DIFF
--- a/account.go
+++ b/account.go
@@ -21,6 +21,12 @@ type Account struct {
 	private  *big.Int
 }
 
+type ExecuteDetails struct {
+	MaxFee  *types.Felt
+	Nonce   *big.Int
+	Version *uint64 // not used currently
+}
+
 /*
 	Instantiate a new StarkNet Account which includes structures for calling the network and signing transactions:
 	- private signing key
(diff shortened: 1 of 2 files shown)


account.go
```diff
@@ -92,30 +116,34 @@ func (account *Account) HashMultiCall(fee *types.Felt, nonce *big.Int, calls []t
 func (account *Account) EstimateFee(ctx context.Context, calls []types.Transaction) (*types.FeeEstimate, error) {
```
tarrencev: I think we want to pass the `ExecutionDetails` here too, so the nonce can be manually provided, otherwise the estimate fee payload is different from the subsequent call that uses it

account.go
```diff
@@ -53,8 +59,26 @@ func (account *Account) Sign(msgHash *big.Int) (*big.Int, *big.Int, error) {
 	- implementation has been tested against OpenZeppelin Account contract as of: https://github.com/OpenZeppelin/cairo-contracts/blob/4116c1ecbed9f821a2aa714c993a35c1682c946e/src/openzeppelin/account/Account.cairo
 	- accepts a multicall
 */
-func (account *Account) Execute(ctx context.Context, maxFee *types.Felt, calls []types.Transaction) (*types.AddTxResponse, error) {
-	req, err := account.fmtExecute(ctx, maxFee, calls)
+func (account *Account) Execute(ctx context.Context, calls []types.Transaction, details ExecuteDetails) (*types.AddTxResponse, error) {
+	if details.MaxFee == nil {
+		fee, err := account.EstimateFee(ctx, calls)
```
tarrencev: I think the estimate needs a nonce so should come after the nonce logic

tarrencev:
Like the direction! I think it needs a bit more tweaking then can be merged!

tarrencev (approved):

archseer:
`examples` also needs to be updated

PR #86, opened 2022-08-11
feat(Hash): add computeHashOnElements

0xLucqs:
Implementation of `computeHashOnElements`
reference implementation can be found here: https://github.com/starkware-libs/cairo-lang/blob/13cef109cd811474de114925ee61fd5ac84a25eb/src/starkware/cairo/common/hash_state.py#L6

### DIFF
--- a/caigo.go
+++ b/caigo.go
@@ -147,6 +147,16 @@ func (sc StarkCurve) HashElements(elems []*big.Int) (hash *big.Int, err error) {
 }
 
 /*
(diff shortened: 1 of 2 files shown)


balthazar:
👍🏻  Would be very useful!

tarrencev:
lgtm! as a followup, would be great to use it here: https://github.com/dontpanicdao/caigo/blob/main/account.go#L90

PR #87, opened 2022-08-12
feat(caigo): replace append -> HashElements with ComputeHashOnElements

0xLucqs:


gregoryguillou:
Your PR is technically correct and there is no other instance of that! So LGTM.

~~Excuse my stupid question but when you have a callData array, the way you put it in the payload is `[ len(callData), callData...]`, no?~~ So I got it: the hash of the array add the length at the end! You are not adding the elements!

gregoryguillou (approved):
LGTM

PR #89, opened 2022-08-13
Feat/migrate rpc to felt

internnos:
revert `EntryPointSelector` from `Felt` to `string`


rpc/api_addtransaction_test.go
```diff
@@ -148,7 +148,7 @@ func TestAddInvokeTransaction(t *testing.T) {
 						types.StrToFelt("0x2b"),
 						types.StrToFelt("0x0"),
 					},
-					EntryPointSelector: types.StrToFelt("0x15d40a3d6ca2ac30f4031e42be28da9b056fef9bb7357ac5e85627ee876e5ad"),
+					EntryPointSelector: "0x15d40a3d6ca2ac30f4031e42be28da9b056fef9bb7357ac5e85627ee876e5ad",
```
gregoryguillou, line 151: That is where I do not get it. If that is the entrypoint in string, it should be `__execute__` for instance...

PR #92, opened 2022-08-15
Extract the felt implementation

gregoryguillou:
This PR extracts the Felt implementation from the main package. The goal is to check if we can move to an interface and mask the fact that the underlying data are big.Int as a result there are several changes:

- Since the current implementation of the curve depends on Int, if we do not want to expose Int to the outsde, I think we have to have the curve with the Felt at the moment. Unfortunately, I think there is also a problem in the current account implementation : it leaks the private/public key which in our case is an Int. For now I will not manage that issue since we can probably do it later as a result: (1) I will leave the private/public keys as Int and (2) I will change the Signature, Hash implementation to move to Felts.
- We will have to dereference the current functions/method that rely on Int from the caigo an other package so that we do not know anything BUT the fact that it is a Felt. That is a huge work!


felt/caigo.go
```diff
@@ -1,4 +1,4 @@
-package caigo
+package felt
```
tarrencev, line 1: Prefer for this to move to a `pedersen` or `curve` package and for the functions to be exposed as plain functions rather than struct members. So a caller can do: `pedersen.Verify(...)`, `pedersen.Sign(...)` rather than creating a curve instance each time

felt/felt.go
```diff
@@ -1,4 +1,4 @@
-package types
+package felt
```
tarrencev, line 1: IMO this should stay in `types`, to avoid stuttering

felt/caigo.go
gregoryguillou: You are right! But we might need several curve too

felt/felt.go
gregoryguillou: The problem here is that if you want several implementation of felts, you need to separate the types from felt. Once done, we should be able to rely on an interface.

gregoryguillou:
This PR leads nowhere. It is crazy complex for no benefit at least for now. I will retry to move to the felt without all the refacto

PR #93, opened 2022-08-15
chore: extract curve to separate package

tarrencev:
a quick and dirty example of how we can improve the curve apis. with this approach the curve is initialized a single time and helpers are all plain functions.


drspacemn:
Yeah I like this UX quite a bit, and definitely like the helpers as function as opposed to methods. One question how are you thinking we deal with caigo specific utils to avoid circular imports: 
```
func XToPubKey(x string) (*big.Int, *big.Int) {
	xin := HexToBN(x)

	yout := GetYCoordinate(xin)

	return xin, yout
}
```

Misc:
- I see the curve init we have the curve params assigned twice for `stark-curve-with-constants` and `stark-curve`
- How would the I init the one vs the other?

PR #94, opened 2022-08-17
Feature/gatewaytestnet transactions.go tests

oxlime:
some more gateway tests still have a few more to complete

### DIFF
--- a/gateway/transaction_test.go
+++ b/gateway/transaction_test.go
@@ -3,6 +3,7 @@ package gateway
 import (
 	"testing"
 	"context"
+	"math/big"
 )
 
 func TestGetTransaction(t *testing.T) {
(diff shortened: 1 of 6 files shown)


gateway/transaction_test.go
```diff
@@ -0,0 +1,189 @@
+package gateway
+
+import (
+	"testing"
+	"context"
+	"math/big"
+)
+
+func TestGetTransaction(t *testing.T) {
```
tarrencev:
```suggestion
func TestTransaction(t *testing.T) {
```

Minor suggestion to remove `Get` to match function signature we're testing.

https://go.dev/doc/effective_go#Getters

tarrencev (approved):
Minor suggestion otherwise lgtm! Thanks for adding these!

If you go down the write endpoint testing path, there are some issues we'll need to work through on the marshaling side

drspacemn:
lgtm

PR #96, opened 2022-08-23
implement RPC API v0.2.0 RC1 and pathfinder 0.3.1

gregoryguillou:
There is actually a problem with this PR because RPC API v0.2.0 RC1, i.e. the v0.40.0 of [the starknet openrpc specification](https://github.com/starkware-libs/starknet-specs/blob/master/api/starknet_api_openrpc.json) is not implemented as part of pathfinder 0.3.1. As a result, we should target **Pathfinder 0.3.1** XOR **RPC API v0.2.0** but we cannot target both. 

This is especially obvious for AddInvokeTransaction.

## **Important**

- I do not see how I can preserve the compatibility with the Gateway API (for now). I simply broke it and implemented a set of specific types in RPC. For instance starknet_getBlockByHash and starknet_getBlockByNumber have disappeared and their usage splitted in different functions.

Note> I checked starknet.js [ProviderInterface]( https://github.com/0xs34n/starknet.js/blob/2ae2125448c334820be52db9be10d1de9c69042a/src/provider/interface.ts#L23) and it does not stick like we do at the API. Maybe we could "Move to higher implementation to make the 2 converge back like they do?"

- The use of `OneOf` in input parameters is a pain. @drspacemn points a construct by @tarrencev that implements a type that is a func. I've implemented it and it looks fine. I will rename the function to remove the option from the name though.

- The API has multiple possible output and it is an even worse concern for usability. The way I have dealt with it for now is returned interface{} and that will be the user responsability to infer those. Here is an example: `PendingBlockWithTxHashes` and `BlockWithTxHashes`. Another more obvious one is the `[]Txn` that contains L1Handler, Declare, Deploy, InvokeV0 and Invokev1. We could provide some helper/utils later and maybe also provide a Interface to manage those

- I am following the type definition of the API and as a result, I have types with multiple embedded type

```
// BlockWithTxsHashes The block object
type BlockWithTxsHashes struct {
	Status BlockStatus `json:"status"`
	BlockHeader
	BlockBodyWithTxHashes
}
```

I have added an file named `json_test.go` to experiment with a few constructs. It looks ok from `TestJSONMultiTypeArrayWithDuck` as soon as we don't mix interfaces{} with structs it seems good to go with JSON. I have gone down that road.

- There are inconsistencies in the API and I understand (see Call based on the v0 only invoke) it will move again as soon as next week, for instance. I am reproducing those inconsistencies which I don't like but I cannot guess either how it will evolve. A few examples:
  - `COMMON_TXN_PROPERTIES` and `BROADCASTED_TXN_COMMON_PROPERTIES ` choose if that is a common property or a common txn
  - `type` is a string in `L1_HANDLER_TXN` and `type` is a TXN_TYPE in `BROADCASTED_TXN_COMMON_PROPERTIES`
  - `contract_address` can be an `ADDRESS` or a `Felt` depending on the instance
Regarding this last point, I am sure we could generate the whole RPC code from the spec (not saying I will do it). So I think we should have it "at the API level in this directory" and provide a "Provider" interface that is shared between Gateway and RPC but does the work in `caigo` the same way starknet.js does it.

- There is also a huge change in AddTransaction: I have implemented the ABI interface and instead of passing multiple parameters, there is now only one. The problem with that approach is that building the object will be very difficult and we will need to create some helpers/utils for that.

- I have added to (I think) nice constructs in the tests that relies on external dependencies (note that the dependencies are only in the tests for security/maintenance concerns):
  - "github.com/nsf/jsondiff" helps do the following: (1) we capture the json payload in the RPC with a json.RAWMessage and (2) we send back the json unmarshalled in the type the caller send. (3) we provide a Capture helper in the tests that allow to compare the JSON capture from the type Marshalled back into JSON
  - "github.com/google/go-cmp/cmp" helps to build the full payload with the right type and so on. Once I got it, I can perform a Deep Comparison with the type.

## Implementation details

Below the list of method impacted by the change

### Verbs to add support for

- [x] starknet_blockHashAndNumber
- [x] starknet_getBlockWithTxHashes
- [x] starknet_getBlockWithTxs
- [x] starknet_getStateUpdate
- [x] starknet_getTransactionByBlockIdAndIndex
- [x] starknet_getBlockTransactionCount
- [x] starknet_getNonce
- [x] starknet_pendingTransactions
- [x] starknet_getEvents

### Verbs that have been changed

- [x] starknet_getStorageAt
- [x] starknet_getClassHashAt
- [x] starknet_getClassAt
- [x] starknet_call
- [x] starknet_estimateFee
- [x] starknet_addDeployTransaction (now accepts optional abi field)
- [x] starknet_addDeclareTransaction (now accepts optional abi field)

### verbs that have been removed

- [x] starknet_getBlockByHash
- [x] starknet_getBlockByNumber
- [x] starknet_getTransactionByBlockHashAndIndex
- [x] starknet_getTransactionByBlockNumberAndIndex
- [x] starknet_getCode
- [x] starknet_getBlockTransactionCountByHash
- [x] starknet_getBlockTransactionCountByNumber
- [x] starknet_protocolVersion

### Mandatory tests to make the Indexer work

- [x] starknet_getBlockWithTxHashes covers all the cases
- [x] starknet_getBlockWithTxs covers InvokeV0, DeployTxn and DeclareTxn
- [ ] starknet_getBlockWithTxs covers InvokeV1 (We cannot capture those yet)
- [x] starknet_getClassHashAt
- [x] starknet_getClassAt
- [x] starknet_call
- [x] starknet_getTransactionReceipt

### Other tests

The following tests are worth noticing:

- [x] TestCaptureUnsupportedBlockTxn allows to detect unsupported block transaction types. In order for it to be efficient, you should change the block range to check.
- [x] BlockNumber
- [x] BlockHashAndNumber
- [x] ChainID
- [x] Syncing
- [x] TransactionByHash
- [x] TransactionByBlockIdAndIndex
- [x] BlockTransactionCount
- [x] Nonce
- [x] StateUpdate
- [x] Class
- [x] StorageAt

### Uncovered sofar (or at least to recheck)

- EstimateFee : from what I see, we must apply the right nonce at the time of the execution and we cannot run 2x the same estimate. As a result, using a test require we generate a proper signature.
- Events
- AddInvokeTransaction [the spec](https://github.com/starkware-libs/starknet-specs/blob/c74e12d0af311537febac8f1972168cf023cb257/api/starknet_write_api.json#L11) and the [current implementation](https://github.com/eqlabs/pathfinder/blob/e831a5ad6e0507405f8cd2a369405a526cbec16d/crates/pathfinder/src/rpc/api.rs#L1309)
- AddDeclareTransaction
- AddDeployTransaction
- (not implemented) TransactionTrace
- (not implemented) TraceBlockTransactions


rpc/api.go
```diff
@@ -4,269 +4,454 @@ import (
...
-}
-
 // Call a starknet function without creating a StarkNet transaction.
-func (sc *Client) Call(ctx context.Context, call types.FunctionCall, hash string) ([]string, error) {
+func (sc *Client) Call(ctx context.Context, call types.FunctionCall, blockIDOption BlockIDOption) ([]string, error) {
+	opt := &blockID{}
+	err := blockIDOption(opt)
+	if err != nil {
+		return nil, err
+	}
```
tarrencev:
```suggestion
	if err := blockIDOption(opt); err != nil {
		return nil, err
	}
```

nit: minimize err scope

rpc/api_contract_test.go
```diff
@@ -188,33 +138,33 @@ func TestStorageAt(t *testing.T) {
 	testConfig := beforeEach(t)
 
 	type testSetType struct {
-		ContractHash   string
+		ContractHash   Address
 		StorageKey     string
-		BlockHashOrTag string
+		BlockHashOrTag BlockIDOption
```
tarrencev:
```suggestion
		BlockHashOrTag BlockHashOrTag
```

I might rename this to match the spec

rpc/api_contract_test.go
```diff
@@ -188,33 +138,33 @@ func TestStorageAt(t *testing.T) {
...
 		ExpectedValue  string
 	}
 	testSet := map[string][]testSetType{
 		"mock": {
 			{
-				ContractHash:   "0xdeadbeef",
+				ContractHash:   Address("0xdeadbeef"),
 				StorageKey:     "_signer",
-				BlockHashOrTag: "latest",
+				BlockHashOrTag: WithBlockIDTag("latest"),
```
tarrencev:
```suggestion
				BlockHashOrTag: WithBlockTag("latest"),
```

rpc/api_events_test.go
```diff
@@ -10,36 +10,39 @@ func TestEvents(t *testing.T) {
...
 			{
-				FromBlockNumber:    250000,
+				FromBlock:          WithBlockIDNumber(1),
 				ExpectedEventCount: 142,
 			},
 		},
 		"mainnet": {
 			{
-				FromBlockNumber:    1000,
+				FromBlock:          WithBlockIDNumber(1),
```
tarrencev:
```suggestion
				FromBlock:          WithBlockNumber(1),
```

tarrencev (commented):
minor feedback but generally lgtm. i agree on breaking the Provider interface here given they seem to be diverging pretty significantly now and the node implementations are pretty mature now

rpc/api.go
gregoryguillou: applied

rpc/api_contract_test.go
gregoryguillou: applied

rpc/api_events_test.go
gregoryguillou: applied

rpc/api_contract_test.go
gregoryguillou: I have renamed

drspacemn (approved):
Greg and I discussed comments offline. This LGTM.

Thanks for all your had work here.

Issue #95, opened 2022-08-17
Potential bug on the payload sent by the gateway


FloRichardAloeCorp:
Hello, I'm facing a strange behaviour with the gateway.Deploy function and I think that it might be a bug.
Caigo version : v0.3.1-0.20220810165710-06e091c57d1c
Cairo version : 0.9.1 
## Context
I have to deploy an account and an ERC20 contract. The account is an OpenZeppelin account and the erc20 has been created with OpenZeppelin wizard for cairo. See the **reproduce** section to find the files.

All following instructions have been made with devnet.
### Contract
The ERC20 contract has the following constructor.
```
@constructor
func constructor{
        syscall_ptr: felt*,
        pedersen_ptr: HashBuiltin*,
        range_check_ptr
    }(owner: felt, init_val: Uint256):
    ERC20.initializer('MyToken', 'MTK', 18)
    Ownable.initializer(owner)
    ERC20._mint(owner, init_val)
    return ()
end
```
The code to deploy the contract is the following:
```go
erc20Response, err := gw.Deploy(context.Background(), compiledERC20Contract, types.DeployRequest{
		Type:                gateway.DEPLOY,
		ContractAddressSalt: "0x30a9f5025c1321d78b4027cf28d02014a0319eddcffc7e63319e3ac0b1f0bb1",
		ConstructorCalldata: []string{
			caigo.HexToBN("0x57ecd15c8010362e71eff8ac7d3db6213478e7b1701662321f1409eeaf4720c").String(), // owner
			"2000", // initial supply
			"0",    // Uint256 additionnal parameter
		},
	})
```
### The bug
When I execute the code and retrieve the transaction status , the following error occur
```
{
    "tx_failure_reason": {
        "code": "TRANSACTION_FAILED",
[...]

PR #99, opened 2022-08-25
Feature/deploy_example

FloRichardAloeCorp:
Adding an example that show how to :

-Deploy an OpenZeppelin account
-Deploy an ERC20 contract
-Mint the ERC20 contract
-Transfer tokens from the deployed account to a third-party account
-Document how to provide ETH to the deployed account

Unfortunately, the code can't be tested because of a bug when deploying contracts [#95]


gregoryguillou:
Hello @FlorianRichardSMT,

Thank you for your contribution! We really appreciate the fact that you have found the bug even if we did not corrected yet ;-). We will sort your PR as a tutorial as it is more complete and more task driven than the other examples. Don't worry, I will handle that. There are a few thinks I would like to see corrected though:

1- I will assum people that are using Caigo know about Go but might not know a lot about Starknet:

- Remove the comments that paraphrase the code. If you need to write a comment, prefer writing a block when it is **very** useful. If we don't understand the code, prefer working on having more explit names for functions/vars... (btw it is not the case here. I find the code very readable!
- I like very much you point useful tools like voyager and the faucet. Can you also explain how to compile the contracts? OR embed the contracts json?

2- There are a few (not many) typos that looks more like the fact you did not pay attention: usefull with 2 f, manualy with 1 l, accout no n. You can probably fix 80% of it yourself.

3- In english there is no space before :

4- Use `go workspace (go work init)` instead of referencing the project from the directory. The reason is that the way it is done, if I upgrade Caigo, the project will not upgrade.

5- Do not use "get" in the function names, prefer `balanceOf` rather than `getBalanceOf` see https://go.dev/doc/effective_go#Getters 

6- You might want to keep the private key somewhere so that if the program or the developer fail we do not have to restart everything. As a matter of fact, you can probably fix it with a small comment (in that case only, jk) saying if you might not want to use a private key people know about. 

7- Make the salt predictable/repeatable, even if you, for sure, do not want collisions. A good practice for instance would be to use your *public" key rather than a timestamp. It would prevent people from reusing the same private key.

~~I am not a big fan of having a large main() like that. I cannot think of a better way. Maybe you could code with tests or range through a map of functions?~~ Actually I was re-reading it and it looks okay. we could try to rely on tests rather than main because you can run the tests in a more granular way.... To be frank, I do not know how

@GregoryGuillou

FloRichardAloeCorp:
Thanks for your response @gregoryguillou !
Regarding the 6th item: 
writing the hex representation of the private key into a file would be ok ?

For the main size: 
In my opinion, it's clearer for the developer to read the code from the main and see all steps one by one. I agree with you, it is quite long but facilitate the understanding of the library.

examples/deploy/go.work
```diff
@@ -0,0 +1,3 @@
+go 1.18
+
+use .
```
gregoryguillou: the idea is that you remove the dependency to `github.com/dontpanicdao/caigo` in the go mod. I think adding `use ../..` should do it.

PR #103, opened 2022-09-06
less verbose names + tx status enum

tarrencev:
some minor improvements that can be propagates to other parts


rpc/types.go
```diff
@@ -302,21 +302,19 @@ type BroadcastedInvokeTxnV1 struct {
...
-type ContractEntryPointList []ContractEntryPoint
-
-type ContractABI []ContractABIEntry
+type ABI []ABIEntry
 
 type EntryPointsByType struct {
-	CONSTRUCTOR ContractEntryPointList `json:"CONSTRUCTOR"`
-	EXTERNAL    ContractEntryPointList `json:"EXTERNAL"`
-	L1_HANDLER  ContractEntryPointList `json:"L1_HANDLER"`
+	Constructor []EntryPoint `json:"CONSTRUCTOR"`
```
tarrencev: prefer to use an plain array type over a type alias. the type alias is harder to grok, since you need to read the full name (which doesnt have an enforces scheme) or follow the declaration to see what is aliased

rpc/types.go
```diff
@@ -325,10 +323,10 @@ type ContractClass struct {
 
 	EntryPointsByType EntryPointsByType `json:"entry_points_by_type"`
 
-	Abi *ContractABI `json:"abi,omitempty"`
+	Abi *ABI `json:"abi,omitempty"`
 }
 
-type ContractABIEntry interface {
+type ABIEntry interface {
```
tarrencev: when we have type names that aren't ambiguous, i.e. dont exist in other contexts, we should prefer shorter type names

gregoryguillou (approved):
Like I said. We should have a talk. 👍  for the enum. I am not so big on renaming the types as it will not match the spec anymore. My biggest concern sofar is the TXN and BLOCKS being interface{} as IN parameters. This is a huge concern.

I think we should talk about that. v0.2.0 of the protocol is out. To my understanding, Pathfinder is not their yet. In addition, I assume we cannot remove v0... except if regenesis removes those protocol lecacy verbs.

rpc/types/event.go
```diff
@@ -0,0 +1,31 @@
...
+
+type EventFilter struct {
+	FromBlock BlockID `json:"from_block"`
+	ToBlock   BlockID `json:"to_block"`
+	Address   Hash    `json:"address"`
+	// Keys the values used to filter the events
+	Keys []string `json:"keys"`
+
+	// ContinuationToken a pointer to the last element of the delivered page, use this token in a subsequent query to obtain the next page
+	ContinuationToken *string `json:"continuation_token"`
```
gregoryguillou:
```suggestion
	ContinuationToken *string `json:"continuation_token,omitempty"`
```
I would assume that

PR #106, opened 2022-09-09
ci: run rpc testnet

tarrencev:


tarrencev:
merging despite the fail since this isn't the breaking commit

PR #112, opened 2022-09-15
fix BlockID to get blocknumber 0 succeed

gregoryguillou:


rpc/types/block.go
```diff
@@ -18,9 +18,9 @@ type BlockHashAndNumberOutput struct {
 // blockID is an unexposed struct that is used in a OneOf for
 // starknet_getBlockWithTxHashes.
 type BlockID struct {
-	Number uint64 `json:"block_number,omitempty"`
-	Hash   *Hash  `json:"block_hash,omitempty"`
-	Tag    string `json:"block_tag,omitempty"`
+	Number *uint64 `json:"block_number,omitempty"`
```
tarrencev, line 21: imo it would be better to handle this in the marshaller than by making `Number` a pointer. it is always dirty to pass primitive types as reference

tarrencev (approved):

PR #114, opened 2022-09-18
Updated Caigo Examples

### LINKED ISSUE #109: Uint256 Examples Ignores 'High' field
Not sure if Caigo maintainers here but the example given in Caigo for `balanceOf` ignores the 'high' result for Uint256. See https://github.com/dontpanicdao/caigo/tree/main/examples/deploy
This is not a good idea as the high field could have a value. In JS, Immutable X does it like this https://github.com/immutable/imx-cairo/blob/main/tests/utils/starknetUtils.ts#L31

I'll look into making a PR if I have a capacity.

rzmahmood:
- Added `.idea` to gitignore for Goland users
- Added `replace` in examples for local testing
- Updated examples to they compiler with latest caigo
- Updated Example to use` Uint256` for balance calls
Fixes https://github.com/dontpanicdao/caigo/issues/109

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -15,3 +15,5 @@ examples/**/*.json
 examples/**/*.sum
 
 */**/*abi.json
+
+.idea/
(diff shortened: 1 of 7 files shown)


tarrencev:
lgtm thanks!

PR #115, opened 2022-09-20
fix declare transaction and add test

gregoryguillou:
This PR:
- upgrade the demo contract to cairo 0.10
- fixes AddDeclareTransaction
- add a function to guess the ABI types
- provides a test with the ABI entry

The API does not look compliant with json-rpc v0.2.0 but it has been tested with Pathfinder 0.3.4 with success

### DIFF
--- a/rpc/write.go
+++ b/rpc/write.go
@@ -5,6 +5,8 @@ import (
 	"compress/gzip"
 	"context"
 	"encoding/base64"
+	"encoding/json"
+	"fmt"
 
 	"github.com/dontpanicdao/caigo/rpc/types"
 )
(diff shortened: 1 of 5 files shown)


rpc/write.go
```diff
@@ -76,3 +78,32 @@ func encodeProgram(content []byte) (string, error) {
 	program := base64.StdEncoding.EncodeToString(buf.Bytes())
 	return program, nil
 }
+
+func guessABI(abis []interface{}) (*types.ABI, error) {
```
tarrencev, line 82: works for now but we should ideally move this to the `Unmarshal` implementation for ABI

tarrencev (approved):

rpc/write.go
gregoryguillou: So true!

PR #117, opened 2022-09-21
Allow to have 2 parallel implementations of RPC API

gregoryguillou:
Pathfinder will provide 2 implementation in parallel of the json-rpc: v0.1.0 and v0.2.0. To anticipate that move and allow to maintain the 2 API that rely on the same verbs bit different payload, this PR creates 2 structs, each one providing a separalet interface.


gregoryguillou:
talked to @tarrencev . He said we would rely on branches (at least to start)

PR #124, opened 2022-09-22
Added Unmarshal for ContractClass and Cleanup

rzmahmood:
Resolves the issue https://github.com/dontpanicdao/caigo/issues/113

Also cleans up a lot of duplicated code and functionality which shouldn't be exposed

### DIFF
--- a/examples/deploy/main.go
+++ b/examples/deploy/main.go
@@ -227,7 +227,10 @@ func balanceOf(gw *gateway.GatewayProvider, erc20address, accountAddress string)
 	low := types.StrToFelt(res[0])
 	hi := types.StrToFelt(res[1])
 
-	balance := types.NewUint256(low, hi)
+	balance, err := types.NewUint256(low, hi)
+	if err != nil {
+		return "", nil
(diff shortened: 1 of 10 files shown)


gregoryguillou (changes requested):
Very very valuable PR! 💯  This is what I would like you to change; it is just surface change:
- Remove the dependency to `testify` assert. We do not need that package. 
- Use `Marshal` to marshal the class rather than `MarshalJSON` in your tests. In real code you will do it that way and the tests will be copied.

Other than those 2 minor changes, it looks very good

gregoryguillou:
Hi,

I have built a PR on top of this one, see #126 to fix another problem I know we had that is when using `map[string]interface{}{}` underlying numbers that are >max_int64 are turned into `float64`. The fix consists in using `map[string]json.RawMessage{}` and checking when the program payload is already a gzip/base64 payload.

I have also managed the changes I had pointed out! I am closing the PR as a consequence. Thank you again for your help 👍 !

@gregoryguillou

Issue #135, opened 2022-09-28
Feature: Add Support for Devnet Mint and Mint Lite


rzmahmood:
The Shardlabs devnet allows for minting to an address using the local faucet -> https://github.com/Shard-Labs/starknet-devnet#mint-token---local-faucet

This is necessary when the deployed and funded accounts are not sufficient (e.g. wrong account contract).

It would be great to start adding support for devnet-specific functionality such as this.

gregoryguillou:
Sorry, I did not see this issue. This is something we have this tests https://github.com/dontpanicdao/caigo/blob/5114a4722290c5f43dc1451942779fd21df40139/rpc/xsessions/plugin_test.go#L97 . I am not sure how to introduce that. Could we have a test hellper like `httptest` ?

rzmahmood:
Resolving this due to added functionality

PR #147, opened 2022-10-07
improve types and account management

gregoryguillou:
This PR is a refactor. In includes the following changes:
- Accounts are now shared between the Gateway and RPC implementations. They have been moved up into the `caigo` directory.
- Plugin support has been added to the session key and both implementation
- Types when shared like FunctionCall have been moved up into the `caigo/types` directory. 
- Implementation specific inner types have been introduced with `MarshalJSON/UnmarshalJSON`. They are supposed to implement specificities from the API like the fact calldata should be base10 on the gateway
- `rpc/types` directory has been removed as it should not be used by any higher level package.
- `rpc` has been renamed `rpcv01` has the type system will have to be rewritten and we want to prepare for the yet to come `rpcv02`
- `accounts` proposes to install and demonstrate the session key with a `v0` account.
- `accounts` proposes to install v1 and v0 accounts and demonstrates the addInvokeTransaction (i.e. account.Execute`) with the 2 protocol versions
- `test` provides now facilities for the devnet as requested by https://github.com/dontpanicdao/caigo/issues/135


gregoryguillou:
This PR introduces a `github.com/dontpanicdao/caigo/test` package with some commands to run against devnet as requested by #135

gregoryguillou:
I had to propose a PR https://github.com/Shard-Labs/starknet-devnet/pull/303 so that devnet can support v1 with RPC v0.1.

PR #148, opened 2022-10-10
Fix bug in StarkCurve GenerateSecret and error in seed

taco-paco:


caigo.go
```diff
@@ -90,12 +90,14 @@ func (sc StarkCurve) Sign(msgHash, privKey *big.Int, seed ...*big.Int) (x, y *bi
 	}
 
 	invalidK := true
+	inSeed := big.NewInt(0)
 	for invalidK {
-		inSeed := big.NewInt(0)
 		if len(seed) == 1 {
 			inSeed = seed[0]
 		}
```
gregoryguillou: I assume this also has to be moved above the `for invalidK` loop. It is the same error except it is not starting with 0 but with a user defined seed

gregoryguillou (changes requested):
Thank you very much for taking the time to correct those 2 issues! You might also want to move the inSeed above the for loop when there is a seed parameter set in the function.

Increasing the length of the hash to 32-bytes (beyond fixing the issue you have pointed) will also increase the probability of the secret to not be below sc.N and thus the retry (thus the infinite loop):
- Can you explain what do you expect the impact in CPU will be on the signature or had a simple test?
- Do you, by any chance, have an example of the impact of the bug on `GenerateSecret` being generated with 31 instead of 32 bytes?

caigo.go
taco-paco: My bad. Fixed

gregoryguillou (approved):

gregoryguillou:
I did not write that piece of code and would be interested to have some feedback. Adding @tarrencev and @drspacemn for reviewing the change. The fix here is not the infinite loop but the signature.

gregoryguillou:
Hello @taco-paco ,

Draft PR (#152) makes sure the error is due to the fact we are running from an old SHA. You can rebase your PR or I will merge the one passing the tests. You will remain the author of the fix in both cases.

We will need some tests to see if the number of invalid values based on increases of the secret size. You might want to help with it otherwise I will do it later this week.

I am not into the maths that is why the PR not pass it yet. I need some additional feedback. Let me know if that is a problem your change is staling.

Thank you again for your contribution to the project.

drspacemn:
Looking at the spec yes this looks good:
```
While tlen < qlen, do the following:

              V = HMAC_K(V)

              T = T || V
```

So this lgtm.
Agree with @gregoryguillou on the rest of the PR logistics.

drspacemn (approved):

gregoryguillou:
I've done some checks:
- The tests are failing because the private key is stored in the repository and the actions actually runs from the forked repository. We will have to change that setup but it has nothing to do with the code.
- performance remains the same with the change.

I will merge this PR and add the tests on top of it as part of a separate PR.

PR #153, opened 2022-10-13
Sign: remove unused invalidK parameter

archseer:
It's always true so this is equivalent to a `for {}` loop


gregoryguillou (approved):
you are right

gregoryguillou:
Hello @archseer,
1. Your code is correct. The problem is the CI expose a variable that your branch cannot access because it is on a fork. Your code is actually passing tests.
2. I merged a previous PR and yours has a conflict with it. Do you mind fixing it or create a new PR?

archseer:
Rebased!

PR #155, opened 2022-10-13
Add gateway mainnet block test

mattcoburn24:


mattcoburn24:
I saw my branch was behind the upstream by 1 commit so i fetched it and git rebased it into the branch before merging is that the right way to handle that?

gregoryguillou:
@coburn24,

You can check how you did the rebase/merge [here](https://github.com/coburn24/caigo/network). The result is correct even if I might have done it slightly differently.

Your change LGTM. Can you reformat the code?
- On your vscode: "Code > Preferences > Settings" check "Format on Save"
- If it does not work when you update the file, just run `go fmt *.go` in the `gateway` directory

Also, make sure you do not leave comments you do not want to be in the main branch.

mattcoburn24:
@gregoryguillou looks like i have format on save set to true in my settings.json? maybe its prettier overwriting?
Screen Shot 2022-10-14 at 12 01 05 AM

Issue #162, opened 2022-10-21
Calling `account.EstimateFee` with the gateway API + TX version 1 fails on testnet


archseer:
Gateway returns `500: StarkErrorCode.MALFORMED_REQUEST Transaction of version 1 is expected to contain the 'type' field.`

Code looks like this:

```go
 // create new account
account, err := caigo.NewGatewayAccount(privKey, sender.String(), &txm.client.Gw.Gateway, caigo.AccountVersion1)
if err != nil {
	return txhash, errors.Errorf("failed to create new account: %s", err)
}

// get fee for txm
feeEstimate, err := account.EstimateFee(ctx, txs, caigotypes.ExecuteDetails{})
if err != nil {
	return txhash, errors.Errorf("failed to estimate fee: %s", err)
}
```

Same call succeeds on `starknet-devnet`.

archseer:
Thanks! Looks like it's fixed in #166

PR #166, opened 2022-10-23
add test for gateway execute

gregoryguillou:


gregoryguillou:
This PR should fix #162 . This is what it does:
- In adds a `type` with INVOKE_FUNCTION to the EstimateFee function. This is a change from 0.10.1, i.e. last Tuesday. It looks okay on devnet too
- It fixes the compulation for EstimateFee on the gateway because it was not captured from my understanding
- It adds tests for devnet and testnet in a function `TestGatewayAccount_EstimateAndExecute` so that we can easily reproduce these kind of events

PR #172, opened 2022-10-28
EC multiplication constant time 

krebernisak:
Hi folks! Thanks for the SDK, we're using it at Chainlink.

During our analysis we found an issue with the EC Multiplication algorithm.

Caigo’s PrivateToPoint function will multiply the curve generator by the private key scalar, obtaining the public key point. This function is not-constant time and could enable an attacker with knowledge about the timing of key generation to recover the private key. This information leakage will also be present in other side channels such as electromagnetic emanations or CPU temperature. Figure shows a plot of how long the PrivateToPoint function takes depending on the number of set bits on the private scalar:

Screenshot 2022-10-28 at 10 19 32

**Exploit Scenario**

An attacker gets access to a side-channel on the computer running the GenerateKey function and obtains enough information to recover the private key.

**Recommendation**

Use a constant time algorithm for scalar multiplication exposed via the PrivateToPoint function. Fix the upper bound on the private scalar sampling.

This is a **high difficulty exploit scenario**, but we'd still love to patch this before moving to production.

This PR implements the [__Double-And-Always-Add__ algorithm](https://www.semanticscholar.org/paper/Elliptic-Curves-and-Side-Channel-Analysis-Joye/7fc91d3684f1ab63b97d125161daf57af60f2ad9/figure/1) which should bring ~10x improvements over the original Double-And-Add implementation, measured as standard deviation when timing the PrivateToPoint function on keys (ec multiplication scalars) with 1 to 251 bits set.

To verify run the the benchmark test where we test multiple different implementations:
```
$ go test -bench BenchmarkEcMultAll -run=^$
....
-----------------------------
algo=       Double-And-Add
stats(ns)
  mean=     1.0868900971390551e+06
  variance= 1.9250450004438724e+11
  std-dev=  438753.34761616035

-----------------------------
algo=       Double-And-Always-Add
stats(ns)
  mean=     1.595599511976048e+06
  variance= 2.1961769142645793e+09
  std-dev=  46863.385646628  // ~10x IMPROVEMENT

-----------------------------
algo=       Montgomery-Ladder
stats(ns)
  mean=     1.469757121756487e+06
  variance= 2.8182458667915974e+10
  std-dev=  167876.31955673787 // ~2,5x IMPROVEMENT

-----------------------------
algo=       Montgomery-Ladder-Lsh
stats(ns)
  mean=     1.4823170239520958e+06
  variance= 2.667844449605206e+10
  std-dev=  163335.37429489076 // ~2,5x IMPROVEMENT
```

This PR also brings the fuzz test and adds one signature verification test found in original implementaiton by Starkware here:  [starkware-libs/crypto-cpp](https://github.com/starkware-libs/crypto-cpp/blob/master/src/starkware/crypto/ecdsa_test.cc)

Resources:
- [Wiki | Double And Add](https://en.wikipedia.org/wiki/Elliptic_curve_point_multiplication#Double-and-add)
- [Wiki | Montgomery Ladder](https://en.wikipedia.org/wiki/Elliptic_curve_point_multiplication#Montgomery_ladder)
- [Enhanced Elliptic Curve Scalar Multiplication Secure Against Side Channel Attacks and Safe Errors](https://cosade.telecom-paristech.fr/presentations/s2_p2.pdf)
- [Elliptic Curves and Side-Channel Analysis](https://www.semanticscholar.org/paper/Elliptic-Curves-and-Side-Channel-Analysis-Joye/7fc91d3684f1ab63b97d125161daf57af60f2ad9/figure/1)
- [Elliptic-curve cryptography VIII - Constant-time scalar multiplication](https://www.youtube.com/watch?v=dhTxzfBsjGw)


gregoryguillou:
Sorry for not answering earlier to this PR. Thank you for the feedback and the PR. We will get back to you.

gregoryguillou:
TLDR: 
- a few people reviewed your PR at StarknetCC Lisboa and green-lighted it.
- tests did not pass but the main reason is the devnet/rpc upgrade
- after investigation there is still one case that I do not understand and is failing.

I have reverted your PR for now. We will extract the failing case to understand if it comes from the change or from the test. Once done, I will probably ask you for more help.

PR #173, opened 2022-10-31
fix ci by reverting to earlier devnet

gregoryguillou:
The reason the CI is failing is because devnet has been updated to v0.3.5 with now implements rpc v0.2 instead of rpc v0.1. This PR reverts to v0.3.4 to fix the tests for now. We should be able to upgrade


gregoryguillou:
duplicate

PR #186, opened 2022-12-05
provide support for new Goerli2 chain ID SN_GOERLI2 in gateway provider

alex-sumner:
Following the upgrade of the StarkNet testnets today, the Goerli2 network has a new chain ID SN_GOERLI2. This pull request allows caigo clients to connect to accounts deployed on Goerli2.

Prior to the StarkNet upgrade a caigo client could connect to Goerli2 as follows:
```
 gw := gateway.NewProvider(gateway.WithBaseURL("https://alpha4-2.starknet.io"))
 account, err := caigo.NewGatewayAccount(<account private key>, <account address>, gw, caigo.AccountVersion1)
``` 
But since the upgrade the returned account has all transactions rejected with the message “Signature (…, …), is invalid, with respect to the public key …, and the message hash …”.

With this pull request clients can successfully connect as follows:
```
 gw := gateway.NewProvider(gateway.WithChain("SN_GOERLI2"))
 account, err := caigo.NewGatewayAccount(<account private key>, <account address>, gw, caigo.AccountVersion1)
```
Passing any string which contains "goerli2" in either upper or lower case as the chain ID will work.

### DIFF
--- a/gateway/gateway.go
+++ b/gateway/gateway.go
@@ -17,9 +17,11 @@ const (
 	DEPLOY_ACCOUNT string = "DEPLOY_ACCOUNT"
 	DECLARE        string = "DECLARE"
 	GOERLI_ID      string = "SN_GOERLI"
+	GOERLI2_ID     string = "SN_GOERLI2"
 	MAINNET_ID     string = "SN_MAIN"
 	LOCAL_BASE     string = "http://localhost:5050"
 	GOERLI_BASE    string = "https://alpha4.starknet.io"
+	GOERLI2_BASE   string = "https://alpha4-2.starknet.io"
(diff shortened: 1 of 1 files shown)


gregoryguillou:
Thank you @alex-sumner and sorry for the delay. We needed it! We will add some tests and make sure this network is also tested.

gregoryguillou (approved):
thank

alex-sumner:
Excellent, thanks!On 8 Dec 2022, at 18:44, Gregory Guillou ***@***.***> wrote:﻿
Thank you @alex-sumner and sorry for the delay. We needed it! We will add some tests and make sure this network is also tested.

PR #187, opened 2022-12-08
add devnet v0.4 and remove unsupported transaction

gregoryguillou:


gregoryguillou:
testnet and mainnet have remove support for some operations, including the deployment of a contract free of charge. This explains some of the failing tests on testnet/mainnet.

PR #200, opened 2023-02-09
add custom chain id

wn-paradigm:
add custom chain id


thiagodeev:
Thanks @wn-paradigm! Sorry for the delay.
Closing as this code is no longer part of starknet.go

PR #210, opened 2023-06-27
Migrate to Junos felt pkg

rianhughes:
This PR proposes we use Junos felt package. 

Note: If we want to use Junos felt package, and accept this PR, then we should also close https://github.com/NethermindEth/caigo/pull/91, which updates/uses the internal felt package.


typed.go
```diff
@@ -41,15 +44,55 @@ encoding definition for standard StarkNet Domain messages
...
+		if ok {
+			fmtEnc = append(fmtEnc, versionBig)
+		}
 	case "chainId":
-		fmtEnc = append(fmtEnc, types.StrToFelt(dm.ChainId).Big())
+		chainIdFelt := strToFelt(dm.ChainId)
+		chainIdBig, ok := feltToBig(chainIdFelt)
+		if ok {
+			fmtEnc = append(fmtEnc, chainIdBig)
+		}
```
cicr99: could we make a small method for this logic as it repeats multiple times in this switch?

typed.go
```diff
@@ -41,15 +44,55 @@ encoding definition for standard StarkNet Domain messages
...
 	}
 	return fmtEnc
 }
 
+// strToFelt converts a string containing a decimal, hexadecimal or UTF8 charset into a Felt.
+func strToFelt(str string) *felt.Felt {
+	var f *felt.Felt
```
cicr99, line 66: Is it better to have variable f as a named return parameter of the function?

rpcv02/transaction.go
cicr99, line 1: In `transaction_test.go` file there is a struct:
```
type testSetType struct {
	TxHash      ctypes.Felt
	ExpectedTxn Transaction
}
```
where `ctypes` is referring to  `"github.com/NethermindEth/caigo/types"`
Shouldn't this be using juno Felt type as well?

contracts/contractsv02.go
```diff
@@ -30,12 +36,17 @@ func (p *RPCv02Provider) declareAndWaitWithWallet(ctx context.Context, compiledC
 			Nonce:   big.NewInt(1), // TODO: nonce handling
 		},
 		ContractClass: class,
-		SenderAddress: types.StrToFelt("0x01"), // TODO: contant devnet address
+		SenderAddress: SenderAddress, // TODO: contant devnet address
```
cicr99: Think there is a typo in the comment here
_contant -> constant_

rpcv02/transaction.go
rianhughes: Yes it should. However, (given tests are already broken) I tried to keep the scope of this PR to non-test and so I didn't update any of the test files. I figured it would be better to fix tests package by package in separate PRs.

rianhughes: I could update the test files to use junos package but leave the tests as they were?

typed.go
rianhughes: From my understanding, it comes down to preference. Naming the return variables can serve as documentation, but it also comes with a risk of accidentally shadowing them. I generally prefer not to name the return variables as I think the code is actually more readable by explicitly declaring them at the top (particularly if you're new to Go etc), you also don't have to worry about shadowing etc.

cicr99 (approved):
lgtm

PR #220, opened 2023-06-28
Check rpcv02

rianhughes:
https://github.com/NethermindEth/caigo/issues/213

Depends on: https://github.com/NethermindEth/caigo/pull/210


rpcv02/write.go
```diff
@@ -44,14 +42,14 @@ func (provider *Provider) AddDeclareTransaction(ctx context.Context, declareTran
...
 }
 
 // AddDeployAccountTransaction manages the DEPLOY_ACCOUNT syscall
-func (provider *Provider) AddDeployAccountTransaction(ctx context.Context, deployAccountTransaction BroadcastedDeployAccountTransaction) (*AddDeployTransactionOutput, error) {
-	var result AddDeployTransactionOutput
+func (provider *Provider) AddDeployAccountTransaction(ctx context.Context, deployAccountTransaction BroadcastedDeployAccountTransaction) (*AddDeployTransactionResponse, error) {
+	var result AddDeployTransactionResponse
 	if err := do(ctx, provider.c, "starknet_addDeployAccountTransaction", &result, deployAccountTransaction); err != nil {
 		// TODO: check Pathfinder/Devnet error and return
 		// CLASS_HASH_NOT_FOUND
```
rianhughes: Note: we need to be able to return the CLASS_HASH_NOT_FOUND error

rpcv02/write.go
```diff
@@ -29,8 +27,8 @@ func (provider *Provider) AddInvokeTransaction(ctx context.Context, broadcastedI
 }
 
 // AddDeclareTransaction submits a new class declaration transaction.
-func (provider *Provider) AddDeclareTransaction(ctx context.Context, declareTransaction BroadcastedDeclareTransaction) (*AddDeclareTransactionOutput, error) {
-	var result AddDeclareTransactionOutput
+func (provider *Provider) AddDeclareTransaction(ctx context.Context, declareTransaction BroadcastedDeclareTransaction) (*AddDeclareTransactionResponse, error) {
+	var result AddDeclareTransactionResponse
 	if err := do(ctx, provider.c, "starknet_addDeclareTransaction", &result, declareTransaction); err != nil {
 		// TODO: check Pathfinder/Devnet error and return
 		// INVALID_CONTRACT_CLASS
```
rianhughes: Note: we need to be able to return the INVALID_CONTRACT_CLASS error

rpcv02/types_transaction.go
```diff
@@ -300,85 +293,79 @@ func (b BroadcastedDeclareTransaction) MarshalJSON() ([]byte, error) {
...
+	// output["version"] = fmt.Sprintf("0x%x", b.Version)
+	// contractAddressSalt, ok := big.NewInt(0).SetString(b.ContractAddressSalt, 0)
+	// if !ok {
+	// 	return nil, errors.New("wrong salt")
+	// }
+	// output["contract_address_salt"] = fmt.Sprintf("0x%x", contractAddressSalt)
+	// constructorCalldata := b.ConstructorCalldata
+	// output["constructor_calldata"] = constructorCalldata
+	// output["contract_class"] = b.ContractClass
+	// return json.Marshal(b)
```
cicr99: Do we need to keep this commented or can it be removed?

rpcv02/types_transaction.go
```diff
@@ -300,85 +293,79 @@ func (b BroadcastedDeclareTransaction) MarshalJSON() ([]byte, error) {
...
+	// for _, v := range b.ConstructorCalldata {
+	// 	constructorCall, ok := big.NewInt(0).SetString(v, 0)
+	// 	if !ok {
+	// 		return nil, errors.New("wrong constructor call data")
+	// 	}
+	// 	constructorCalldata = append(constructorCalldata, fmt.Sprintf("0x%x", constructorCall))
+	// }
+	// output["constructor_calldata"] = constructorCalldata
+	// output["class_hash"] = b.ClassHash.String()
+	// return json.Marshal(output)
```
cicr99: How about this one?

rpcv02/errors.go
```diff
@@ -66,4 +66,8 @@ var (
 		code:    40,
 		message: "Contract error",
 	}
+	ErrInvlaidContractClass = &RPCError{
```
cicr99: Typo _ErrInvlaidContractClass_ -> _ErrInvalidContractClass_
Once this is fixed it needs to be replaced in other places where it is used

types/types.go
```diff
@@ -26,13 +26,13 @@ type AddDeployResponse struct {
...
-	ConstructorCalldata []string      `json:"constructor_calldata"`
-	ContractDefinition  ContractClass `json:"contract_definition"`
-}
+// // TODO: remove
+// type DeployRequest struct {
+// 	Type                string               `json:"type"`
+// 	ContractAddressSalt string               `json:"contract_address_salt"`
+// 	ConstructorCalldata []string             `json:"constructor_calldata"`
+// 	ContractDefinition  rpcv02.ContractClass `json:"contract_definition"`
+// }
```
cicr99, line 35: I'm a bit confused about DeployRequest struct. I see that it's still being used in some tests and examples, so those shouldn't be working at the moment. I've seen in other files such as `gateway/starknet.go`, it's being used from `rpcv02` (it's commented atm), but it's not defined there. So is it the idea to move it to `rpcv02` types or to replace it with `rpcv02.DeployAccountTxn` or simply to remove it completely?

plugins/xsessions/signer.go
```diff
@@ -103,9 +105,13 @@ func (plugin *SessionKeyPlugin) PluginCall(calls []ctypes.FunctionCall) (ctypes.
 	for _, signature := range plugin.token.signedSession.Signature {
 		data = append(data, fmt.Sprintf("0x%s", signature.Text(16)))
 	}
+	calldAtafelt, err := utils.HexArrToFelt(data)
```
cicr99: Variable name

rpcv02/types_transaction.go
rianhughes: It's used in write.go and contractsv02.go so I think we should keep it for now

types/types.go
rianhughes: The goal is to move away from using the (ambiguous) `types` package, and use more well-defined packages like `rpcv02`, `rpcv03` etc. 

The motivation for this is that (afaik) the `types` package does not follow any well-defined spec outlined by Starknet (plus the fields in the structs are incorrect so we can't/shouldn't use it when making RPCv02 requests anyway). `RPCv02` and `RPCv03` do/will follow well-defined specs outlined by Starknet. I feel like this ambiguity will create unnecessary confusion and problems in the future, which we can prevent by deprecating `types` and using `rpcv02`. 

For example, if I'm a new user of starknet.go, and want to send a types.DeployAccountTxn to some Sequencer, which representation am I using? Does my request follow the rpcv02 spec, the rpcv03 spec, the gateway spec, or something else entirely? How can I be sure it's even correct / following a spec at all, especially if the network changes the spec it follows? By deprecating the `types` package and using packages like `rpcv02`, `rpcv03`, `gateway`, we remove this ambiguity and prevent any problems that might arise because of it.

rianhughes: I agree that this test breaks some already broken tests, but as we discussed previously we decide to leave fixing tests to other PRs.

cicr99 (approved):

Issue #217, opened 2023-06-28
Fix tests - rpcv02


rianhughes:
Fix tests in the rpcv02 folder

rianhughes:
https://github.com/NethermindEth/starknet.go/pull/226

PR #226, opened 2023-07-04
Fix rpcv02 tests

rianhughes:
https://github.com/NethermindEth/starknet.go/issues/217


cicr99:
Some tests are still failing for me on devnet:
- `TestCall` on `call_test.go`
- `TestChainID` on `provider_test.go`
- `TestDeclareTransaction` on `write_test.go`

cicr99:
Also could we update the workflows? `Testing RPC v0.2 on devnet` is being skipped atm. In addition, testing on `testnet` is running but is failing, we could skip this step until we manage to fix them as well.
Would love to see a green status

rianhughes:
All tests should pass on mainnet, testnet, devnet, and mock, when making requests to Juno

rpcv02/chain.go
```diff
@@ -22,11 +23,18 @@ func (provider *Provider) ChainID(ctx context.Context) (string, error) {
...
-	var result SyncStatus
+	var result interface{}
 	// Note: []interface{}{}...force an empty `params[]` in the jsonrpc request
 	if err := provider.c.CallContext(ctx, &result, "starknet_syncing", []interface{}{}...); err != nil {
 		return nil, err
 	}
-	return &result, nil
+	switch res := result.(type) {
+	case bool:
+		return &SyncStatus{SyncStatus: res}, nil
```
cicr99, line 33: Not sure if this falls under rpc v0.2 scope but the only possible value for `result` when it has bool type is `False`. Should we check that here?

rpcv02/types.go
```diff
@@ -19,15 +19,15 @@ type StorageEntry struct {
 	Value *felt.Felt `json:"value"`
 }
 
-type StorageEntries struct {
-	StorageEntry []StorageEntry
-}
+// type StorageEntries struct {
+// 	StorageEntry []StorageEntry
+// }
```
cicr99, line 24: This can be removed now, right?

rpcv02/call_test.go
```diff
@@ -2,90 +2,82 @@ package rpcv02
...
 		},
 		"testnet": {
 			{
-				FunctionCall: types.FunctionCall{
-					ContractAddress:    types.StrToFelt("0x029260ce936efafa6d0042bc59757a653e3f992b97960c1c4f8ccd63b7a90136"),
-					EntryPointSelector: "decimals",
-					Calldata:           []string{},
+				FunctionCall: FunctionCall{
+					ContractAddress:    utils.TestHexToFelt(t, "0x029260ce936efafa6d0042bc59757a653e3f992b97960c1c4f8ccd63b7a90136"),
+					EntryPointSelector: utils.TestHexToFelt(t, "0x004c4fb1ab068f6039d5780c68dd0fa2f8742cceb3426d19667778ca7f3518a9"), // types.GetSelectorFromNameFelt("decimals"), // 0x004c4fb1ab068f6039d5780c68dd0fa2f8742cceb3426d19667778ca7f3518a9
```
cicr99: what about these comments?

rpcv02/chain.go
rianhughes: I don't think we need to as the endpoint will (/should) never return True

rpcv02/types.go
rianhughes: Yes

cicr99 (approved):
lgtm

PR #239, opened 2023-07-10
Replace caigo with starknet.go

rianhughes:
Renaming the project from caigo to starknet.go has created issues with importing the starknet.go package. Here I simply find-replace "caigo" with "starknet.go"

### DIFF
--- a/examples/account/main.go
+++ b/examples/account/main.go
@@ -5,9 +5,9 @@ import (
 	"fmt"
 	"math/big"
 
-	"github.com/NethermindEth/caigo"
-	"github.com/NethermindEth/caigo/gateway"
-	"github.com/NethermindEth/caigo/types"
+	starknetgo "github.com/NethermindEth/starknet.go"
+	"github.com/NethermindEth/starknet.go/gateway"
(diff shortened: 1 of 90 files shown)


examples/account/main.go
```diff
@@ -35,10 +35,10 @@ func main() {
 	fmt.Println("Counter is currently at: ", callResp[0])
 
 	// init account handler
-	ks := caigo.NewMemKeystore()
+	ks := starknet.go.NewMemKeystore()
```
cicr99: Wouldn't this cause conflict? Also package name in previous files is `starknetgo`
This repeats other times as well

rianhughes: Yes this is an error. I thought I caught all of these! Thanks, I'll have another look now

rianhughes: Ah wait, I didn't catch this because I was neglecting broken tests and examples. I'll fix the examples

rianhughes: I fixed the import errors in examples, but it's worth noting that they are still broken and should be fixed in following PRs

cicr99 (approved):
lgtm

PR #245, opened 2023-07-18
Fix/main folder test

Gabulhas:
Fixed some of the tests in the main folder. (Every one excepet for TestRPCv02Account_EstimateFee)


account.go
```diff
@@ -30,8 +30,17 @@ type account interface {
 	Call(ctx context.Context, call types.FunctionCall) ([]string, error)
 	Nonce(ctx context.Context) (*big.Int, error)
 	EstimateFee(ctx context.Context, calls []types.FunctionCall, details types.ExecuteDetails) (*types.FeeEstimate, error)
-	Execute(ctx context.Context, calls []types.FunctionCall, details types.ExecuteDetails) (*types.AddInvokeTransactionOutput, error)
-	Declare(ctx context.Context, classHash string, contract rpcv02.ContractClass, details types.ExecuteDetails) (types.AddDeclareResponse, error)
+	Execute(
+		ctx context.Context,
+		calls []types.FunctionCall,
+		details types.ExecuteDetails,
+	) (*types.AddInvokeTransactionOutput, error)
```
rianhughes, line 37: This should be on a single line like the other other functions for consistency.

rianhughes: The same goes for other places where the function signature is spread across multiple lines.

account.go
```diff
@@ -326,12 +354,14 @@ func (account *Account) prepFunctionInvokeRPCv02(ctx context.Context, messageTyp
 	switch account.version {
 	case 1:
 		calldata := fmtCalldataStrings(calls)
+		fmt.Println("CALLDATA", calldata)
```
rianhughes, line 357: fmt.Prinln() statements (used during testing) should be removed

accountgw_test.go
```diff
@@ -26,58 +27,75 @@ func TestGatewayAccount_EstimateAndExecute(t *testing.T) {
 	testSet := map[string][]testSetType{
 		"devnet": {{
 			ExecuteCalls: []types.FunctionCall{{
```
rianhughes, line 29: Ideally we should use the rpcv02 types. Also, why were the ContracAddresses removed?

accountgw_test.go
```diff
@@ -26,58 +27,75 @@ func TestGatewayAccount_EstimateAndExecute(t *testing.T) {
...
+		// Convert the counter address to a Felt.
+		counterAddress, err := utils.HexToFelt(testConfig.CounterAddress)
+		if err != nil {
+			t.Fatalf("Failed to convert counter address to Felt: %v", err)
+		}
+
+		// Set the contract address for the execute and query calls.
+		for i := range test.ExecuteCalls {
+			test.ExecuteCalls[i].ContractAddress = counterAddress
+		}
```
rianhughes, line 56: I think we can move this to the start of the function. It'll also save us ranging over the test.ExecuteCalls

gateway/starknet.go
```diff
@@ -45,19 +45,17 @@ type GatewayFunctionCall struct {
...
-	// 	output["entry_point_selector"] = f.EntryPointSelector
-	// }
-	// calldata := []string{}
-	// for _, v := range f.Calldata {
-	// 	data, _ := big.NewInt(0).SetString(v, 0)
-	// 	calldata = append(calldata, data.Text(10))
-	// }
-	// output["calldata"] = calldata
-	// return json.Marshal(output)
+	output := map[string]interface{}{}
```
rianhughes, line 48: Just curious, did json.Marshal(f) cause problems?

providers_test.go
```diff
@@ -257,8 +254,5 @@ func TestGeneral_Syncing(t *testing.T) {
 		if !ok || i.Cmp(big.NewInt(0)) <= 0 {
 			t.Fatal("CurrentBlockNum should be positive number, instead: ", syncv02.CurrentBlockNum)
 		}
```
rianhughes, line 256: Why was the prefix test moved? 

if !strings.HasPrefix(syncv02.CurrentBlockHash, "0x")..

providers_test.go
```diff
@@ -88,25 +89,21 @@ var (
...
 		return "", err
 	}
 	ctx := context.Background()
 	ctx, cancel := context.WithTimeout(ctx, time.Second*60)
 	defer cancel()
-	tx, err := provider.Deploy(context.Background(), class, types.DeployRequest{
-		ContractAddressSalt: "0x0",
-		ConstructorCalldata: []string{},
-	})
+	tx, err := provider.Deploy(context.Background(), class, rpcv02.DeployAccountTxn{})
```
rianhughes, line 100: I think the ContractAddresSaly may need to be set explicitly

rpcv02/types.go
```diff
@@ -73,7 +74,6 @@ type StateUpdateOutput struct {
 
 // SyncStatus is An object describing the node synchronization status
 type SyncStatus struct {
```
rianhughes, line 76: We'll need to keep SyncStatus.SyncStatus, it's used in one of the other PRs.

rpcv02/types.go
```diff
@@ -83,9 +83,6 @@ type SyncStatus struct {
 }
 
 func (s SyncStatus) MarshalJSON() ([]byte, error) {
```
rianhughes, line 85: The !s.SyncStatus {..} section will need readded

rpcv02/types.go
```diff
@@ -97,25 +94,29 @@ func (s SyncStatus) MarshalJSON() ([]byte, error) {
...
+	}
+	if s.CurrentBlockHash, err = utils.HexToFelt(output["current_block_hash"].(string)); err != nil {
+		return err
+	}
+	if s.HighestBlockHash, err = utils.HexToFelt(output["highest_block_hash"].(string)); err != nil {
+		return err
+	}
+
+	fmt.Println(output)
+	fmt.Printf("TYPES %T %T %T\n", output["starting_block_num"], output["current_block_num"], output["highest_block_num"])
```
rianhughes, line 114: fmt.Prinln() should be removed

typed.go
```diff
@@ -67,17 +68,22 @@ func strToFelt(str string) *felt.Felt {
 	asciiRegexp := regexp.MustCompile(`^([[:graph:]]|[[:space:]]){1,31}$`)
```
rianhughes, line 68: I wonder if we could replace strToFelt with utils.HexToFelt() (which doesn't actually require the strin to be a Hex, we should probably update the name)

rianhughes:
Also, tests are failing on mock, testnet, devnet, and mainnet. We can separate these out into separate issues if you want?

account.go
Gabulhas: Got it.
I was trying some new plugins for my editor and seems like it formatted stuff in a weird way.
I think the same happen in other parts of the code.

accountgw_test.go
Gabulhas: Didn't remove the ContractAddress, just replaced the repetition of the same code to a loop.
Will check out the rpcv02 types.

gateway/starknet.go
Gabulhas: Yes, multiple examples of the json.Marshal were causing Stack Overflows.

accountgw_test.go
Gabulhas: You mean something like this instead?

```
	counterAddress, err := utils.HexToFelt(testConfig.CounterAddress)
	if err != nil {
		t.Fatalf("Failed to convert counter address to Felt: %v", err)
	}


		"devnet": {{
			ExecuteCalls: []types.FunctionCall{{
				EntryPointSelector: types.GetSelectorFromNameFelt("increment"),
				ContractAddress: counterAddress,
			}},
			QueryCall: types.FunctionCall{
				EntryPointSelector: types.GetSelectorFromNameFelt("get_count"),
				ContractAddress: counterAddress,
			},
		}},
		
```

providers_test.go
Gabulhas: because syncv02.CurrentBlockHash is type `*felt.Felt` now, and implicity has 0x, that is, if you convert it to string it always has a `0x` prefix

```go
func (z *Felt) String() string {
	return "0x" + z.val.Text(Base16)
}
```

rpcv02/types.go
Gabulhas: I swear I switched to a debugger 😅

typed.go
Gabulhas: Don't think so, as they arent necessarily Hex values

Starknet.js has a similar function (https://github.com/0xs34n/starknet.js/blob/44d2706f9d862564f59328a2677b2359f9d3f3a6/src/utils/shortString.ts#L33)

accountgw_test.go
rianhughes: Yes exactly

cicr99:
Closing this issue due to the amount of conflicts. Also, too many changes to the code base on main, the fixes made in this PR might not apply anymore

PR #256, opened 2023-07-27
Create SimpleCall Example

rianhughes:
Creates a simple example to make a (read) contract call to a mainnet contract


cicr99 (commented):
It looks good.
I think it would be good to add the steps to run the example in `examples/simpleCall/ReadMe.md`.
Also, as the other examples are still not fixed, we could update the readme of the project to point to this example in particular. I think the others can be commented until they are fixed

cicr99 (approved):

Issue #279, opened 2023-08-11
`starknet_estimateMessageFee` endpoint


cicr99:
- [x] Add method for "starknet_estimateMessageFee"
- [x] Implement type for MSG_FROM_L1

See:
- https://github.com/starkware-libs/starknet-specs/pull/100
- https://github.com/starkware-libs/starknet-specs/pull/115

PR #283, opened 2023-08-15
Implement estimateMessageFee

rianhughes:
closes https://github.com/NethermindEth/starknet.go/issues/279

### DIFF
--- a/rpc/types_transaction_receipt.go
+++ b/rpc/types_transaction_receipt.go
@@ -134,6 +134,17 @@ type MsgToL1 struct {
 	Payload []*felt.Felt `json:"payload"`
 }
 
+type MsgFromL1 struct {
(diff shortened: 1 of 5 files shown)


rpc/types_transaction_receipt.go
```diff
@@ -134,6 +134,17 @@ type MsgToL1 struct {
 	Payload []*felt.Felt `json:"payload"`
 }
 
+type MsgFromL1 struct {
+	// FromAddress The address of the L1 contract sending the message
+	FromAddress *felt.Felt `json:"from_address"`
```
cicr99: According to the spec this is an `ETH_ADDRESS` which is a string

cicr99 (approved):

Issue #233, opened 2023-07-07
Splitting of DECLARE_TXN for different versions


cicr99:
DECLARE_TXN was split into:
- DECLARE_TXN_V1
- DECLARE_TXN_V2

Tasks:
- [ ] Implement both types
- [ ] Check transaction types properties

Issue #227, opened 2023-07-07
Implementation of `DeprecatedContractClass` type in addition to `ContractClass`


cicr99:
`starknet_getClass`  now can return two types: DEPRECATED_CONTRACT_CLASS and CONTRACT_CLASS . Same with `starknet_getClassAt` .
 - [x] Implement corresponding types for DEPRECATED_CONTRACT_CLASS and CONTRACT_CLASS
 - [x] Implement SIERRA_ENTRY_POINT type
 - [x] Replace CONTRACT_ENTRY_POINT with DEPRECATED_CAIRO_ENTRY_POINT
 - [x] Modify methods using these types

PR #288, opened 2023-08-21
Implement DeclareTxn rpcv04

### LINKED ISSUE #277: Reintroduce DECLARE_TXN_V0
See the changes here: https://github.com/starkware-libs/starknet-specs/pull/120/files

> This relates to https://github.com/NethermindEth/starknet.go/issues/233 issue for rpcv3

rianhughes:
closes https://github.com/NethermindEth/starknet.go/issues/277

Also implements: https://github.com/NethermindEth/starknet.go/issues/233

Note: I added ContractClassTemp, because this PR requires https://github.com/NethermindEth/starknet.go/issues/227


rpc/types_contract.go
```diff
@@ -118,19 +128,6 @@ func (c *DeprecatedContractClass) UnmarshalJSON(content []byte) error {
 	return nil
 }
 
-// https://github.com/starkware-libs/starknet-specs/blob/v0.3.0/api/starknet_api_openrpc.json#L2372
-type ContractClass struct {
-	// The list of Sierra instructions of which the program consists
-	SierraProgram []*felt.Felt `json:"sierra_program"`
-
-	// The version of the contract class object. Currently, the Starknet OS supports version 0.1.0
```
cicr99, line 126: Can we add these two comments again?

rpc/transaction.go
```diff
@@ -21,17 +21,20 @@ func adaptTransaction(t TXN) (Transaction, error) {
...
+		case &felt.Zero:
+			var tx DeclareTxnV0
+			json.Unmarshal(txMarshalled, &tx)
+			return tx, nil
+		case felt.Zero.SetUint64(1):
 			var tx DeclareTxnV1
 			json.Unmarshal(txMarshalled, &tx)
 			return tx, nil
-		case new(felt.Felt).SetUint64(2):
+		case felt.Zero.SetUint64(2):
```
kirugan: `Set*` methods in felt first change value inside pointer, then return it's pointer.
After calling `felt.Zero.SetUint64(...)` here and above `felt.Zero` contains either 1 or 2.

My suggestion is to setup known versions before hand:
```
var (
    Version0 = felt.Zero
    Version1 = new(felt).SetUint(1)
    Version2 = ...
)
```

Also please note that right now is pointer comparison, not value comparison IMO it's safer to check like through `felt.Equal()`:
switch {
  case t.Version.Equal(Version1):
      ...
  case t.Version.Equal(Version2):
    ....
}
```

File: rpc/transaction.go [hunk 3]
```diff
@@ -46,8 +49,10 @@ func adaptTransaction(t TXN) (Transaction, error) {
 		json.Unmarshal(txMarshalled, &tx)
 		return tx, nil
 	default:
-		panic("not a transaction")
+		return nil, errors.New("Internal error with adaptTransaction() : unknown transaction type")
```
kirugan: few things to consider:
* golang code style suggests to start error messages with lowercase char (because it looks more cleaner when you wrap errors i.e. "outer error: internal error: deep internal error")
* can we add here more specifics what caused an error here i.e. "unknown transaction type %s"

rpc/transaction.go
```diff
@@ -46,8 +49,10 @@ func adaptTransaction(t TXN) (Transaction, error) {
...
 	}
+	return nil, errors.New("Internal error with adaptTransaction()")
```
kirugan: this one became necessary due to missing default case in version switch, wdyt about adding more specific there i.e. "unknown transaction version %s" ?

rpc/types_transaction.go
```diff
@@ -211,6 +228,10 @@ func unmarshalTxn(t interface{}) (Transaction, error) {
 		case TransactionType_Declare:
 
 			switch TransactionType(casted["version"].(string)) {
+			case "0x0":
```
cicr99, line 231: Wouldn't it be better to use `TransactionVersion` type variables here instead of writing the string yourself?

cicr99 (approved):

PR #289, opened 2023-08-21
Rpcv04 tx status

### LINKED ISSUE #281: Transaction status was split into TXN_FINALITY_STATUS and TXN_EXECUTION_STATUS
Check TransactionState in types/types.go and related methods to modify them accordingly

See https://github.com/starkware-libs/starknet-specs/pull/107

### LINKED ISSUE #282: Add optional "revert_reason" for transactions for failed execution
See:
- https://github.com/starkware-libs/starknet-specs/pull/99
- https://github.com/starkware-libs/starknet-specs/pull/106

rianhughes:
closes: https://github.com/NethermindEth/starknet.go/issues/281
closes: https://github.com/NethermindEth/starknet.go/issues/282


contracts/contracts.go
```diff
@@ -42,7 +42,7 @@ func (p *GatewayProvider) declareAndWaitWithWallet(ctx context.Context, compiled
 		return nil, err
 	}
 	if !receipt.Status.IsTransactionFinal() ||
```
augustbleeds, line 44: I noticed that `IsTransactionFinal` still includes the `REJECTED` state (even after the failed tx inclusion changes). Is it because

a. this state is temporarily reported by the gateway for all txs for backward compatibility 
or 
b. will the gateway have to support `REJECTED` forever for old transactions? 

EDIT: after writing I realize that the gateway will be eventually be sunset due to decentralization of the sequencer -- so I'm assuming it's (a)?

rianhughes: The rpc spec (v4) no longer has TXN_STATUS, so this function should be removed. It's probably worth highlighting that the contracts package is outdated, and likely needs updated to follow the spec. So far we have mainly been focusing on the rpc package, little to no work has been done on the other packages. Also, any functions / packages to do with the gateway will be removed because the sequencer will be going offline.

rpc/types.go
```diff
@@ -170,54 +170,64 @@ type FeeEstimate struct {
...
 
 const (
-	TransactionAcceptedOnL1 TransactionState = "ACCEPTED_ON_L1"
-	TransactionAcceptedOnL2 TransactionState = "ACCEPTED_ON_L2"
-	TransactionNotReceived  TransactionState = "NOT_RECEIVED"
-	TransactionPending      TransactionState = "PENDING"
-	TransactionReceived     TransactionState = "RECEIVED"
-	TransactionRejected     TransactionState = "REJECTED"
+	TxnExecutionStatusSUCCEEDED TxnExecutionStatus = "SUCCEEDED"
+	TxnExecutionStatusREVERTED  TxnExecutionStatus = "REVERTED"
```
augustbleeds, line 177: I believe there is one more valid execution status state: `RECEIVED` based off of https://github.com/starkware-libs/starknet-specs/pull/134 .

I confirmed with Oded (from Starkware) who said: ```so indeed, even if starknet_addInvokeTransaction returns the tx hash, it is possible the tx will not be included in a block, and therefore the finality status will not be "accepted by L1" or "accepted by L2", but it will be "received".
The only case this is possible is if finality status is received and the execution status is rejected```

rpc/types.go
```diff
@@ -170,54 +170,64 @@ type FeeEstimate struct {
...
-func (s TransactionState) IsTransactionFinal() bool {
-	if s == TransactionAcceptedOnL2 ||
-		s == TransactionAcceptedOnL1 ||
-		s == TransactionRejected {
-		return true
+type TxnFinalityStatus string
+
+const (
+	TxnFinalityStatusAcceptedOnL1 TxnFinalityStatus = "ACCEPTED_ON_L1"
+	TxnFinalityStatusAcceptedOnL2 TxnFinalityStatus = "ACCEPTED_ON_L2"
```
augustbleeds, line 208: as a corollary to the above, there should also be a `RECEIVED` finality status

contracts/contracts.go
augustbleeds: Thanks for clarifying, I originally thought the old rpc (< v4) was there for backwards compatibility but good to know it can be removed entirely!

rpc/types.go
rianhughes: Hey, thanks for catching this. Technically it's not in the spec, but I agree it should probably be added. @cicr99 what are your thoughts?

augustbleeds: I'm triple-checking with Starknet to make sure I understand it correctly. Still waiting on hearing back but will let you know!

augustbleeds: You can ignore my comments about adding status states. I got a confirmation from Starkware (Ariel) that

> TLDR is
the pending status no longer exists for a transaction, ACCEPTED_ON_L2 is the “lowest” finality status (we might also add RECEIVED there for 0.5.0)
the only difference between pending receipt and receipt is the inclusion of block related fields (block_hash, block_number), note that a pending receipt doesn’t affect the finality, you know with certainty that it will be included in the next block

and 

> tldr the starknet-specs have not included RECEIVED status yet. So currently in v0.4, will the API just error for a tx hash that hasn't been ACCEPTED_ON_L2 yet

Hope that clears things up

rpc/types.go
augustbleeds: ^ This comment can be ignored (see thread above)

rpc/transaction.go
```diff
@@ -118,20 +118,24 @@ func (provider *Provider) WaitForTransaction(ctx context.Context, transactionHas
...
-				if r.Status.IsTransactionFinal() {
-					return r.Status, nil
+				if r.ExecutionStatus == TxnExecutionStatusSUCCEEDED {
+					return r.ExecutionStatus, nil
 				}
 			case L1HandlerTransactionReceipt:
-				if r.Status.IsTransactionFinal() {
-					return r.Status, nil
+				if r.ExecutionStatus == TxnExecutionStatusSUCCEEDED {
+					return r.ExecutionStatus, nil
```
kirugan: Maybe add execution status to the interface and check for failure before the switch? i.e.
```
if receipt.ExecutionStatus() != ...SUCCEEDED {
    return ...
 }
 
 switch ...
 ```

Comment (rianhughes) [reply rpc/transaction.go]: I added the GetExecutionStatus() method to the interface. I also think we can just return the execution status, without having to type cast in the switch statement

File: rpc/types_transaction_receipt.go [hunk 5]
```diff
@@ -6,20 +6,21 @@ import (
...
-	Type         TransactionType  `json:"type"`
-	MessagesSent []MsgToL1        `json:"messages_sent"`
+	ActualFee       *felt.Felt         `json:"actual_fee"`
+	ExecutionStatus TxnExecutionStatus `json:"execution_status"`
+	FinalityStatus  TxnFinalityStatus  `json:"finality_status"`
+	BlockHash       *felt.Felt         `json:"block_hash"`
+	BlockNumber     uint64             `json:"block_number"`
+	Type            TransactionType    `json:"type"`
+	MessagesSent    []MsgToL1          `json:"messages_sent"`
+	RevertReason    string             `json:"revert_reason"`
```
cicr99: I think `Type` and `RevertReason` are not required fields, you could add _omitempty_ at the end
https://github.com/starkware-libs/starknet-specs/blob/6fadca1f1abf48523c0aa799a10061f40f973468/api/starknet_api_openrpc.json#L2195

cicr99 (approved):

PR #290, opened 2023-08-21
RPCv04 Remove BroadcastTx

### LINKED ISSUE #230: Splitting of BROADCASTED_DECLARE_TXN for different versions
BROADCASTED_DECLARE_TXN was split into: 
- BROADCASTED_DECLARE_TXN_V1 (depends on Deprecated_contract_class #227 )
- BROADCASTED_DECLARE_TXN_V2

### LINKED ISSUE #278: Remove Broadcasted_txn types
Also modify transaction types
See https://github.com/starkware-libs/starknet-specs/pull/114

rianhughes:
closes: https://github.com/NethermindEth/starknet.go/issues/278
closes: https://github.com/NethermindEth/starknet.go/issues/230

Note: Tests are fixed in previous rpcv04 PRs
Note: Will have a merge conflict with previous rpcv04 PRs. Should be merged last.


account.go
```diff
@@ -351,13 +351,11 @@ func (account *Account) prepFunctionInvokeRPC(ctx context.Context, messageType s
...
 			return nil, err
 		}
-		return &rpc.BroadcastedInvokeV1Transaction{
-			BroadcastedTxnCommonProperties: rpc.BroadcastedTxnCommonProperties{
-				MaxFee:    maxFeeFelt,
-				Version:   version,
-				Signature: []*felt.Felt{s1Felt, s2Felt},
-				Nonce:     nonceFelt,
-			},
+		return &rpc.InvokeTxnV1{
```
cicr99, line 354: Shouldn't the `type` also be set here?

account.go
```diff
@@ -451,14 +449,13 @@ func (account *Account) EstimateFee(ctx context.Context, calls []types.FunctionC
...
-					Nonce:     call.Nonce,
-					Type:      "INVOKE",
-				},
+
+			estimates, err := account.rpc.EstimateFee(ctx, []rpc.EstimateFeeInput{rpc.InvokeTxnV1{
+				MaxFee:        call.MaxFee,
+				Version:       rpc.TransactionV1,
+				Signature:     call.Signature,
+				Nonce:         call.Nonce,
+				Type:          "INVOKE",
```
cicr99, line 458: Could we use `TransactionType_Invoke` instead? (https://github.com/NethermindEth/starknet.go/blob/a3a745a7309046e1fdbd1d649c867c58759617ca/rpc/types_transaction_receipt.go#L43)

account.go
```diff
@@ -501,14 +498,12 @@ func (account *Account) Execute(ctx context.Context, calls []types.FunctionCall,
...
-					Signature: call.Signature,
-					Nonce:     call.Nonce,
-					Type:      "INVOKE",
-				},
+			resp, err := account.rpc.AddInvokeTransaction(ctx, rpc.InvokeTxnV1{
+				MaxFee:        call.MaxFee,
+				Version:       rpc.TransactionV1,
+				Signature:     call.Signature,
+				Nonce:         call.Nonce,
+				Type:          "INVOKE",
```
cicr99, line 506: Same here about using `TransactionType_Invoke`

contracts/contractsv02.go
```diff
@@ -18,6 +17,7 @@ import (
 type RPCProvider rpc.Provider
 
 func (p *RPCProvider) declareAndWaitWithWallet(ctx context.Context, compiledClass []byte) (*DeclareOutput, error) {
+	panic("This needs updated - Declare transactions have changed significantly over the past few rpc-spec updates")
```
cicr99, line 14: I know after this line the rest of the code of the function is unreachable but can we comment it out same as in the other methods?

rpc/types_transaction.go
```diff
@@ -104,7 +108,12 @@ type DeclareTxnV0 struct {
 }
 
 type DeclareTxnV1 struct {
-	CommonTransaction
+	TransactionHash *felt.Felt         `json:"transaction_hash,omitempty"`
```
cicr99: Why `TransactionHash`? I couldn't find it in the [spec](https://github.com/starkware-libs/starknet-specs/blob/6fadca1f1abf48523c0aa799a10061f40f973468/api/starknet_api_openrpc.json#L1687)

rpc/types_transaction.go
```diff
@@ -104,7 +108,12 @@ type DeclareTxnV0 struct {
...
+	MaxFee          *felt.Felt         `json:"max_fee"`
+	Version         TransactionVersion `json:"version"`
+	Signature       []*felt.Felt       `json:"signature"`
+	Nonce           *felt.Felt         `json:"nonce"`
+	Type            TransactionType    `json:"type"`
 
 	// ClassHash the hash of the declared class
 	ClassHash *felt.Felt `json:"class_hash"`
```
cicr99: Should we add _"omitempty"_ for `ClassHash` and `DeprecatedContractClass`? One of them is required, but you shouldn't get both at the same time, so there's always one of them that will be empty

rpc/types_transaction.go
cicr99: I've seen most of the transaction types have this property as well, but there is the same question

rpc/write.go
```diff
@@ -4,13 +4,11 @@ import (
 	"context"
 )
 
-type BroadcastedInvokeTransaction interface{}
-
-func (provider *Provider) AddInvokeTransaction(ctx context.Context, broadcastedInvoke BroadcastedInvokeTransaction) (*AddInvokeTransactionResponse, error) {
+func (provider *Provider) AddInvokeTransaction(ctx context.Context, invokeTxn AddInvokeTxnInput) (*AddInvokeTransactionResponse, error) {
```
cicr99: Why use `AddInvokeTxnInput` instead of just `InvokeTxnV1`? (See [here](https://github.com/starkware-libs/starknet-specs/blob/6fadca1f1abf48523c0aa799a10061f40f973468/api/starknet_write_api.json#L19))

rpc/types_transaction_response.go
```diff
@@ -14,6 +14,12 @@ type AddDeployAccountTransactionResponse struct {
 	ContractAddress *felt.Felt `json:"contract_address"`
 }
 
+// AddDeployTransactionResponse provides the output for AddDeployTransaction.
+type AddDeployTransactionResponse struct {
+	TransactionHash *felt.Felt `json:"transaction_hash"`
+	ContractAddress *felt.Felt `json:"contract_address"`
+}
```
cicr99: This type looks exactly the same as `AddDeployAccountTransactionResponse`, and I don't see it being used anywhere on the code. What's the point?

Also both types have the same comment on the top (probably because of copy-pasting), if they are meant for different things it 'll be good to change the comments

rpc/contract.go
```diff
@@ -111,15 +111,8 @@ func (provider *Provider) Nonce(ctx context.Context, blockID BlockID, contractAd
...
-// EstimateFee estimates the fee for a given Starknet transaction.
-func (provider *Provider) EstimateFee(ctx context.Context, requests []BroadcastedTransaction, blockID BlockID) ([]FeeEstimate, error) {
-	// tx, ok := request.(*BroadcastedInvokeV1Transaction)
-	// TODO:
-	// NOTE: EntryPointSelector is now just part of Calldata
-	// if ok {
-	// 	tx.EntryPointSelector = fmt.Sprintf("0x%x", types.GetSelectorFromName(tx.EntryPointSelector))
-	// 	request = tx
-	// }
+// EstimateFee estimates the fee for a given StarkNet transaction.
```
cicr99: Use **Starknet** instead of **StarkNet**

contracts/contractsv02.go
```diff
@@ -27,14 +27,13 @@ func (p *RPCProvider) declareAndWaitWithWallet(ctx context.Context, compiledClas
...
-			Type:    "DECLARE",
-			MaxFee:  new(felt.Felt).SetUint64(10000),
-			Version: "0x01",
-			Nonce:   new(felt.Felt).SetUint64(1), // TODO: nonce handling
-		},
-		ContractClass: class,
+	tx, err := provider.AddDeclareTransaction(ctx, rpc.DeclareTxnV1{
+		Type:    "DECLARE",
+		MaxFee:  new(felt.Felt).SetUint64(10000),
+		Version: "0x01",
```
cicr99: I would use [TransactionType_Declare](https://github.com/NethermindEth/starknet.go/blob/a3a745a7309046e1fdbd1d649c867c58759617ca/rpc/types_transaction_receipt.go#L40) for the `Type` and [TransactionV1](https://github.com/NethermindEth/starknet.go/blob/a3a745a7309046e1fdbd1d649c867c58759617ca/rpc/types_transaction.go#L292C2-L292C15) for the `Version`

account.go
rianhughes: Probably, although I think this is out of scope of this PR for two reasons 1) the goal was to replace broacastInvokeTxn with InvokeTxn types (which this PR does) not fix the logic in account.go, and 2) this account file is broken, and will be replaced with the new accounts package.

account.go
rianhughes: Same comment as above

account.go
rianhughes: Same comment as above

rpc/write.go
rianhughes: It's because we used to allow InvokeTxnV0 and InvokeTxnV1 for the rpcv03 spec, which took [broadcastInvokeTxn](https://github.com/starkware-libs/starknet-specs/blob/v0.3.0/api/starknet_api_openrpc.json#L1902), I'll update it though

contracts/contractsv02.go
rianhughes: Same comment as above

rpc/types_transaction.go
rianhughes: Looks like it found it's way back in after merging main a couple of times. It wasn't there in the original PR, but you're right that it should be removed.

rpc/types_transaction_response.go
rianhughes: This is no longer needed, I'll remove it

account.go
cicr99: I thought to point it out since `Type` was a required field for the `InvokeTxnV1` type, but you are right, this file is already broken and will be replaced soon

rpc/block_test.go
```diff
@@ -220,34 +220,31 @@ func TestBlockWithTxsAndInvokeTXNV0(t *testing.T) {
...
-						Type:    "INVOKE",
-						MaxFee:  utils.TestHexToFelt(t, "0xde0b6b3a7640000"),
-						Version: TransactionV0,
-						Signature: []*felt.Felt{
-							utils.TestHexToFelt(t, "0x7bc0a22005a54ec6a005c1e89ab0201cbd0819621edd9fe4d5ef177a4ff33dd"),
-							utils.TestHexToFelt(t, "0x13089e5f38de4ea98e9275be7fadc915946be15c14a8fed7c55202818527bea"),
-						},
-						Nonce: nil,
+		Transactions: []BlockTransaction{
+			BlockInvokeTxnV0{
```
cicr99, line 225: Shouldn't this also have a `TransactionHash` field set? I assume you should be able to use the same one that is being removed

rpc/block_test.go
```diff
@@ -306,19 +303,17 @@ func TestBlockWithTxsAndInvokeTXNV0(t *testing.T) {
...
 			if !ok {
 				t.Fatal("expected invoke v0 transaction")
 			}
-			invokeV0Block, ok := blockWithTxs.Transactions[test.LookupTxnPositionInOriginal].(InvokeTxnV0)
+			invokeV0Block, ok := blockWithTxs.Transactions[test.LookupTxnPositionInOriginal].(BlockInvokeTxnV0)
 			if !ok {
 				t.Fatal("expected invoke v0 transaction")
 			}
-
-			require.Equal(t, invokeV0Want.Hash(), invokeV0Block.Hash(), "expected equal hash")
```
cicr99, line 318: You could also modify this one according to previous comment about `TransactionHash`

rpc/block_test.go
```diff
@@ -348,11 +343,10 @@ func TestBlockWithTxsAndDeployOrDeclare(t *testing.T) {
 			Timestamp:        1661486036,
 		},
 		Status: "ACCEPTED_ON_L1",
-		Transactions: []Transaction{
-			DeployTxn{
-				TransactionHash: utils.TestHexToFelt(t, "0x35bd2978d2061b3463498f83c09322ed6a82e4b2a188506525e272a7adcdf6a"),
-				ClassHash:       utils.TestHexToFelt(t, "0x1ca349f9721a2bf05012bb475b404313c497ca7d6d5f80c03e98ff31e9867f5"),
-				DeployTransactionProperties: DeployTransactionProperties{
+		Transactions: []BlockTransaction{
+			BlockDeployTxn{
```
cicr99, line 350: Same thing here about `TransactionHash`

rpc/block_test.go
```diff
@@ -376,20 +370,17 @@ func TestBlockWithTxsAndDeployOrDeclare(t *testing.T) {
...
-					TransactionHash: utils.TestHexToFelt(t, "0x5ad2f85499ea92d33d4a44c8cd4640d1ee4e25c3ee6df0bdf0a76c12c052f0a"),
-					BroadcastedTxnCommonProperties: BroadcastedTxnCommonProperties{
-						Type:      TransactionType_Declare,
-						MaxFee:    utils.TestHexToFelt(t, "0x27a64c6e425"),
-						Version:   TransactionV1,
-						Signature: []*felt.Felt{utils.TestHexToFelt(t, "0x1454ab28f0bf18f0fd8002bc92169e6443feba6c605728c86850c0dcc9f6f9a"), utils.TestHexToFelt(t, "0xf545949c899ff1d16c61629996e898db2697a2e3e7fa9071b016500ca5c1d1")},
-						Nonce:     utils.TestHexToFelt(t, "0x333"),
-					},
+		Transactions: []BlockTransaction{
+			BlockDeclareTxnV1{
```
cicr99, line 378: Same thing here about `TransactionHash`

rpc/block_test.go
```diff
@@ -404,21 +395,18 @@ func TestBlockWithTxsAndDeployOrDeclare(t *testing.T) {
...
-					TransactionHash: utils.TestHexToFelt(t, "0x45d04652ba51685b7b82fc17b3d5741a7c43992369c0b0aebd60916fa23b9b2"),
-					BroadcastedTxnCommonProperties: BroadcastedTxnCommonProperties{
-						Type:      TransactionType_Declare,
-						MaxFee:    utils.TestHexToFelt(t, "0x50c8f30287c"),
-						Version:   TransactionV2,
-						Signature: []*felt.Felt{utils.TestHexToFelt(t, "0x6be01a56087382337a29fd4577dd20fd82cc9f38f69b8d19e07fc101c3c5ad9"), utils.TestHexToFelt(t, "0x4c633a5582d3932fbfcea8abd45c7453e88a562f1a38877b9575d6a6b926ea2")},
-						Nonce:     utils.TestHexToFelt(t, "0xd"),
-					},
+		Transactions: []BlockTransaction{
+			BlockDeclareTxnV2{
```
cicr99, line 404: Same thing here about `TransactionHash`

rpc/types_transaction.go
```diff
@@ -304,53 +268,19 @@ func (v *TransactionVersion) BigInt() (*big.Int, error) {
...
-}
+type AddDeclareTxnInput interface{}
 
-type BroadcastedDeclareTransaction interface{}
+var _ AddDeclareTxnInput = DeclareTxnV1{}
+var _ AddDeclareTxnInput = DeclareTxnV2{}
 
-var _ BroadcastedDeclareTransaction = BroadcastedDeclareTransactionV1{}
-var _ BroadcastedDeclareTransaction = BroadcastedDeclareTransactionV2{}
+type AddDeployAccountTxnInput interface{}
```
cicr99: Why do you need this one? There is only one `DeployAccountTxn`. Also, this is only being used in `AddDeployAccountTransaction` method which could just receive a `DeployAccountTxn` as an input parameter ([spec](https://github.com/starkware-libs/starknet-specs/blob/6fadca1f1abf48523c0aa799a10061f40f973468/api/starknet_write_api.json#L152C55-L152C73))

cicr99 (approved):

PR #292, opened 2023-08-22
Rpcv03 starknet get blocks

### LINKED ISSUE #231: Support PendingBlock for `starknet_getBlockWithTxHashes` and `starknet_getBlockWithTxs`

rianhughes:
closes https://github.com/NethermindEth/starknet.go/issues/231

### DIFF
--- a/rpc/block.go
+++ b/rpc/block.go
@@ -50,7 +50,6 @@ func WithBlockTag(tag string) BlockID {
 }
 
 // BlockWithTxHashes gets block information given the block id.
-// TODO: add support for PendingBlock
 func (provider *Provider) BlockWithTxHashes(ctx context.Context, blockID BlockID) (interface{}, error) {
 	var result Block
 	if err := do(ctx, provider.c, "starknet_getBlockWithTxHashes", &result, blockID); err != nil {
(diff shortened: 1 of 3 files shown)


rpc/block.go
```diff
@@ -95,5 +105,14 @@ func (provider *Provider) BlockWithTxs(ctx context.Context, blockID BlockID) (in
...
 		return nil, err
 	}
+	// if header.Hash == nil it's a pending block
+	if result.BlockHeader.BlockHash == nil {
+		return PendingBlock{
+			ParentHash:       result.ParentHash,
+			Timestamp:        result.Timestamp,
+			SequencerAddress: result.SequencerAddress,
+			Transactions:     result.Transactions,
+		}, nil
```
cicr99, line 115: Is it worth it to enclose this logic in a separate method as it is already used before? something like getPendingBlockFromBlock?

rianhughes: I'm happy leaving it as it is given its only repeated twice. If we repeated it more often then I would say move it to a separate function, but the current version is quicker to read and understand.

cicr99 (approved):

PR #293, opened 2023-08-22
rpcv03 starknet_getClass

rianhughes:
closes https://github.com/NethermindEth/starknet.go/issues/227

PR #294, opened 2023-08-22
Rpcv03 declare txn

### LINKED ISSUE #233: Splitting of DECLARE_TXN for different versions

rianhughes:
closes https://github.com/NethermindEth/starknet.go/issues/233

Note: This should be merged immediately after https://github.com/NethermindEth/starknet.go/pull/293. I created ContractClassTemp so that I could work on this issue while waiting for ContractClass to be updated in another [issue](https://github.com/NethermindEth/starknet.go/pull/243)


rpc/types_transaction.go
```diff
@@ -93,17 +93,34 @@ func (tx L1HandlerTxn) Hash() *felt.Felt {
...
 }
 
-type DeclareTxn struct {
+type DeclareTxnV1 struct {
 	CommonTransaction
 
 	// ClassHash the hash of the declared class
 	ClassHash *felt.Felt `json:"class_hash"`
 
+	DeprecatedContractClass `json:"contract_class"`
```
cicr99: This property is not included in version 03, is it from 04? Should we add it here anyway?
I don't think it'll cause any problems

rpc/types_transaction.go
```diff
@@ -93,17 +93,34 @@ func (tx L1HandlerTxn) Hash() *felt.Felt {
...
+
+type DeclareTxnV2 struct {
+	CommonTransaction
+
 	// SenderAddress the address of the account contract sending the declaration transaction
 	SenderAddress *felt.Felt `json:"sender_address"`
+
+	CompiledClassHash *felt.Felt `json:"compiled_class_hash"`
+
+	ContractClassTemp `json:"contract_class,omitempty"`
```
cicr99: Same thing with this one?

rpc/types_transaction.go
```diff
@@ -93,17 +93,34 @@ func (tx L1HandlerTxn) Hash() *felt.Felt {
...
 	// ClassHash the hash of the declared class
 	ClassHash *felt.Felt `json:"class_hash"`
 
+	DeprecatedContractClass `json:"contract_class"`
+
+	// SenderAddress the address of the account contract sending the declaration transaction
+	SenderAddress *felt.Felt `json:"sender_address"`
+}
+
+type DeclareTxnV2 struct {
```
cicr99, line 106: If there's a difference in the ContractClass property between these two types then the following is probably not necessary, but otherwise,  we could think of unifying most of the properties of DeclareTxnV1 and  DeclareTxnV2 in the same type, as most of their properties are the same

rpc/types_contract.go
```diff
@@ -27,6 +27,25 @@ type EntryPointsByType struct {
 	L1Handler   []EntryPoint `json:"L1_HANDLER"`
 }
 
+type DeprecatedContractClass struct {
```
cicr99: This will probably cause some conflicts with #293 
Also it will need to be fixed to actually use the new `ContractClass` type instead of `ContractClassTemp`
I think we should wait for #293 to get in, then fix this one and merge it, wyt?

rpc/types_transaction.go
rianhughes: I think we should keep them separate for now for two reasons 1) it's much easier from a users perspective to know what fields the need to fill in if they are using DeclareTxV1 and DeclareTxV2, and 2) in RPCv04 there are three declare transaction types and merging them will cause unnecessary confusion from a dev + users perspective imo

cicr99 (approved):

Issue #301, opened 2023-08-30
Create account package


rianhughes:
This PR is broken down into the following PRS:

The following methods should be included:
- AddInvokeTransaction
- Sign
- SignInvokeTransaction
- EstimateFee
- BuildInvokeTx
- TransactionHash
- Call
- Nonce
- Other read, write and trace methods

It will also require hashing logic, and signature logic.

Later it should also incorporate the contracts package so users can easily interact with contract methods by loading the contract, like in web3py, starknet-js, etc.

Note: there is a draft PR getting everything needed to submit a correct AddInvokeTransaction [here](https://github.com/NethermindEth/starknet.go/pull/302).

PR #302, opened 2023-08-30
New account

rianhughes:
https://github.com/NethermindEth/starknet.go/issues/301


mattcoburn24 (approved):

rianhughes:
Note: look into check-summing addresses for ks

rianhughes:
Note: Hash functions will need updated after rpcv04 merge, as they depend on the transaction type. See Juno impl

rianhughes:
TestAddDeclare Fails because the spec doesn't require the ContractClass field, but the Sequencer does?? Other than that, the package should be good to review.

mattcoburn24 (approved):

.github/workflows/test_account.yml
```diff
@@ -0,0 +1,41 @@
...
+        ports:
+          - 5050:5050
+
+    steps:
+      - uses: actions/checkout@v3
+
+      - name: Set up Go
+        uses: actions/setup-go@v3
+        with:
+          go-version: 1.18
```
kirugan: I suggest using `go.mod` file. Here is how we do it in Juno (probably also need to increase setup-go version):
```
      - uses: actions/setup-go@v4.1.0
        with:
          go-version-file: go.mod
```

utils/Felt.go
```diff
@@ -31,5 +32,16 @@ func FeltToBigInt(f *felt.Felt) (*big.Int, bool) {
 }
 
 func BigIntToFelt(big *big.Int) (*felt.Felt, error) {
-	return new(felt.Felt).SetString(big.String())
+	return new(felt.Felt).SetBytes(big.Bytes()), nil
```
kirugan: Are you sure it's 1 - 1 mapping of bytes? If so let's rewrite `FeltToBigInt` function above - converting from strings probably is inefficient.

utils/Felt.go
```diff
@@ -31,5 +32,16 @@ func FeltToBigInt(f *felt.Felt) (*big.Int, bool) {
...
+}
+func FeltArrToBigIntArr(f []*felt.Felt) (*[]*big.Int, error) {
```
kirugan: Slice contains a pointer to underlying array data, can we change signature so only []*big.Int is returned (if this signature was unintentional)

kirugan: And whitespace before declaration please 🙏

utils/Test.go
```diff
@@ -15,11 +15,10 @@ func TestHexToFelt(t testing.TB, hex string) *felt.Felt {
 	return f
 }
 
-func TestBigIntToFelt(t testing.TB, big *big.Int) *felt.Felt {
+func TestBigIntToFelt(t testing.TB, big *big.Int) (*felt.Felt, error) {
```
kirugan: But error is always nil, right? 🤔

rpc/types_transaction_interfaces.go
```diff
@@ -0,0 +1,70 @@
...
+type InvokeTxnType interface{}
+
+var _ InvokeTxnType = InvokeTxnV0{}
+var _ InvokeTxnType = InvokeTxnV1{}
+
+type DeclareTxnType interface{}
+
+var _ DeclareTxnType = DeclareTxnV0{}
+var _ DeclareTxnType = DeclareTxnV1{}
+var _ DeclareTxnType = DeclareTxnV2{}
```
kirugan, line 66: Could you please check is it possible to use nil pointer here:
```
var _ Type = (*SpecificType)(nil)
```

account/hash.go
```diff
@@ -0,0 +1,72 @@
...
+		callData = append(callData, tx.ContractAddress, tx.EntryPointSelector)
+
+		if len(tx.Calldata) == 0 {
+			callData = append(callData, &felt.Zero, &felt.Zero)
+			continue
+		}
+
+		callData = append(callData, new(felt.Felt).SetUint64(uint64(len(callArray))), new(felt.Felt).SetUint64(uint64(len(tx.Calldata))+1))
+		for _, cd := range tx.Calldata {
+			callArray = append(callArray, cd)
```
kirugan: no need to add it one by one, you can unpack it and get rid of `range`:
```
callArray = append(callArray, tx.Calldata...)
```

account/account.go
```diff
@@ -0,0 +1,406 @@
...
+		provider:       provider,
+		AccountAddress: accountAddress,
+		publicKey:      publicKey,
+		ks:             keystore,
+	}
+
+	chainID, err := provider.ChainID(context.Background())
+	if err != nil {
+		return nil, err
+	}
```
cicr99, line 62: Why do this after creating `account` instance? Couldn't this have been done before so you could set `ChainId` as well when creating it?

account/account.go
```diff
@@ -0,0 +1,406 @@
...
+		calldataHash,
+		tx.MaxFee,
+		account.ChainId,
+		[]*felt.Felt{tx.Nonce},
+	)
+}
+
+func (account *Account) TransactionHashInvoke(tx rpc.InvokeTxnType) (*felt.Felt, error) {
+
+	// https://docs.starknet.io/documentation/architecture_and_concepts/Network_Architecture/transactions/#deploy_account_hash_calculation
```
cicr99: Could this be another link?
It's probably https://docs.starknet.io/documentation/architecture_and_concepts/Network_Architecture/transactions/#v0_hash_calculation or the one for invoke v1 maybe

account/account.go
```diff
@@ -0,0 +1,406 @@
...
+func (account *Account) TransactionHashDeployAccount(tx rpc.DeployAccountTxn, contractAddress *felt.Felt) (*felt.Felt, error) {
+
+	// https://docs.starknet.io/documentation/architecture_and_concepts/Network_Architecture/transactions/#deploy_account_transaction
+
+	// There is only version 1 of deployAccount txn
+	if tx.Version != rpc.TransactionV1 {
+		return nil, ErrTxnTypeUnSupported
+	}
+
+	Prefix_DEPLOY_ACCOUNT := new(felt.Felt).SetBytes([]byte("deploy_account"))
```
cicr99: Why isn't this one a global variable same as `TRANSACTION_PREFIX` and the others?

account/account.go
```diff
@@ -0,0 +1,406 @@
...
+			account.ChainId,
+			[]*felt.Felt{txn.Nonce},
+		)
+	}
+	return nil, ErrTxnTypeUnSupported
+}
+
+func (account *Account) TransactionHashDeclare(tx rpc.DeclareTxnType) (*felt.Felt, error) {
+
+	Prefix_DECLARE := new(felt.Felt).SetBytes([]byte("declare"))
```
cicr99: Should we use `DECLARE_PREFIX` instead?

account/account_test.go
```diff
@@ -0,0 +1,538 @@
...
+					Calldata: []*felt.Felt{
+						utils.TestHexToFelt(t, "0x1"),
+						utils.TestHexToFelt(t, "0x049d36570d4e46f48e99674bd3fcc84644ddd6b96f7c741b1562b82f9e004dc7"),
+						utils.TestHexToFelt(t, "0x83afd3f4caedc6eebf44246fe54e38c95e3179a5ec9ea81740eca5b482d12e"),
+						utils.TestHexToFelt(t, "0x0"),
+						utils.TestHexToFelt(t, "0x3"),
+						utils.TestHexToFelt(t, "0x3"),
+						utils.TestHexToFelt(t, "0x049d36570d4e46f48e99674bd3fcc84644ddd6b96f7c741b1562b82f9e004dc7"),
+						utils.TestHexToFelt(t, "0x1"),
+						utils.TestHexToFelt(t, "0x0"),
```
cicr99: Wouldn't it be better to use `utils.TestHexArrToFelt` directly instead?

account/account_test.go
```diff
@@ -0,0 +1,538 @@
...
+				ChainID:        "SN_MAIN",
+				FnCall: rpc.FunctionCall{
+					Calldata: []*felt.Felt{
+						utils.TestHexToFelt(t, "0x1"),
+						utils.TestHexToFelt(t, "0x5dbdedc203e92749e2e746e2d40a768d966bd243df04a6b712e222bc040a9af"),
+						utils.TestHexToFelt(t, "0x2f0b3c5710379609eb5495f1ecd348cb28167711b73609fe565a72734550354"),
+						utils.TestHexToFelt(t, "0x0"),
+						utils.TestHexToFelt(t, "0x1"),
+						utils.TestHexToFelt(t, "0x1"),
+						utils.TestHexToFelt(t, "0x52884ee3f"),
```
cicr99: Same comment here about `utils.TestHexArrToFelt`

account/account_test.go
```diff
@@ -0,0 +1,538 @@
...
+					utils.TestHexToFelt(t, "0x1"),
+					utils.TestHexToFelt(t, "0x49d36570d4e46f48e99674bd3fcc84644ddd6b96f7c741b1562b82f9e004dc7"),
+					utils.TestHexToFelt(t, "0x83afd3f4caedc6eebf44246fe54e38c95e3179a5ec9ea81740eca5b482d12e"),
+					utils.TestHexToFelt(t, "0x0"),
+					utils.TestHexToFelt(t, "0x3"),
+					utils.TestHexToFelt(t, "0x3"),
+					utils.TestHexToFelt(t, "0x49d36570d4e46f48e99674bd3fcc84644ddd6b96f7c741b1562b82f9e004dc7"),
+					utils.TestHexToFelt(t, "0x1"),
+					utils.TestHexToFelt(t, "0x0"),
+				},
```
cicr99: Same comment here about `utils.TestHexArrToFelt`

account/account.go
rianhughes: Yes, we could pass chainId in as a variable but this is much easier to work with imo, and it's less error prone (especially for new users)

rpc/types_transaction_interfaces.go
rianhughes: Yes, you can use nil pointer here.

In principle, I think you can pass any type in place of DeclareTxnType since it's just an empty interface, so this doesn't constrain what types are allowed. But it's just the best way I can think of to categorise types like DeclareTxnV0. Would love to know if there's a better/cleaner way

utils/Test.go
rianhughes: I don't think so, because you can pass in invalid strings ["!InValid:)"]

cicr99 (approved):
lgtm

rpc/types_transaction_interfaces.go
kirugan: Ah I see, forgot that it's an empty interface. If you want constraint your interfaces with types you can introduce an empty method like:
```
type SomeCommonType interface {
     SomeCommonType()
 }
 
 type Concrete struct{}
 func (c Concrete) SomeCommonType()
 ```
 Not sure that we need it here IMO. One benefit of using pointers instead of structs - no memory taken by these unused structs (but maybe compiler already optimized for that)

rianhughes: Unfortunately there's no common field for the transactions.

Issue #212, opened 2023-06-27
Implement RPCv03


rianhughes:
release [0.3.0](https://github.com/starkware-libs/starknet-specs/releases/tag/v0.3.0)

- [x] #232
- [x] #274
- [x] #236
- [x] #231
- [x] #227
- [x] #233
- [x] #230
- [x] #235
- [x] #228
- [x] #229
- [x] #234
- [x] #237
- [x] #266
- [x] #270
- [x] #272
- [x] https://github.com/NethermindEth/starknet.go/issues/312
- [x] https://github.com/NethermindEth/starknet.go/issues/314
- [x] https://github.com/NethermindEth/starknet.go/issues/315

cicr99:
Just updated this issue for more details on what needs to be done to achieve this. Feel free to add anything I might be missing

rianhughes:
`method` updates needed for supporting this version:
- [x] `starknet_getStateUpdates` - results needs updated
- [x] `starknet_getClass` - results needs updated  
- [x] `starknet_getClassAt` - results needs updated  
- [x] `starknet_call` - errors needs updated
- [x] `starknet_estimateFee` - parameters, results, and errors need updated
- [x] `starknet_getEvents` - results needs updated

`component` updates needed for supporting this version:
- [x] `EVENT_CHUNK` needs created
- [x] `EVENT_FILTER`  types need updated
- [x] `STATE_UPDATE` types need updated
- [x] `PENDING_STATE_UPDATE` types need created
- [x] `DECLARE_TXN` types need updated, and `DECLARE_TXN_V1` and `DECLARE_TXN_V1` types needs created
- [x] `BROADCAST_DECLARE_TXN` types need updated, `BROADCAST_DECLARE_TXN_V1`, `BROADCAST_DECLARE_TXN_V2` types needs created
- [x] `DEPLOY_ACCOUNT_TXN_PROPERTY`, and `DEPLOY_TXN_PROPERTY` 
- [x] `BROADCAST_DEPLOY_TXN` - needs deleted
- [x] `BROADCAST_INVOKE_TXN`, `INVOKE_TXN` - types need updated
- [x] `COMMON_RECIPET_PROPERTIES` types need updates
- [x] `INVOKE_TXN__RECIPET`,...,`L1_HANDLER_RECEIPT` types need updated 
- [x] `MSG_TO_L1` -types need updated
- [x] `CONTRACT_CLASS` types need updated
[...]

PR #304, opened 2023-09-06
Rpcv03 get class

rianhughes:
Fixes some small issues required for https://github.com/NethermindEth/starknet.go/issues/212

Namely, Class and ClassHashAt should use/return felts instead of strings


cicr99 (approved):
lgtm

PR #309, opened 2023-09-07
Add DeployTxn type back

cicr99:
DeployTxn type was mistakenly removed from the code, so this pr adds it back


rianhughes:
You'll also need to add the Deploy transaction case in [adaptTransaction()](https://github.com/NethermindEth/starknet.go/blob/reintroduce_DeployTxn/rpc/transaction.go#L13), eg something like:

```
case TransactionType_Deploy:
		var tx DeployTxn
		json.Unmarshal(txMarshalled, &tx)
		return tx, nil
```

rianhughes (approved):

PR #316, opened 2023-09-08
rpcv03 implement simulate transaction

### LINKED ISSUE #313: starknet_simulateTransaction
implement https://github.com/starkware-libs/starknet-specs/blob/v0.3.0/api/starknet_trace_api_openrpc.json#L41

rianhughes:
closes https://github.com/NethermindEth/starknet.go/issues/313

### DIFF
--- a/rpc/errors.go
+++ b/rpc/errors.go
@@ -1,5 +1,17 @@
 package rpc
 
+import "errors"
+
+func tryUnwrapToRPCErr(err error, rpcErrors ...*RPCError) error {
+	for _, rpcErr := range rpcErrors {
+		if errors.Is(err, rpcErr) {
+			return rpcErr
+		}
+	}
+
+	return err
+}
+
 type RPCError struct {
 	code    int
 	message string
--- a/rpc/provider.go
+++ b/rpc/provider.go
@@ -47,6 +47,7 @@ type api interface {
 	AddInvokeTransaction(ctx context.Context, broadcastedInvoke BroadcastedInvokeTransaction) (*AddInvokeTransactionResponse, error)
 	AddDeclareTransaction(ctx context.Context, declareTransaction BroadcastedDeclareTransaction) (*AddDeclareTransactionResponse, error)
 	AddDeployAccountTransaction(ctx context.Context, deployAccountTransaction BroadcastedDeployAccountTransaction) (*AddDeployAccountTransactionResponse, error)
+	SimulateTransactions(ctx context.Context, blockID BlockID, txns []BroadcastedTransaction, simFlags []SimulationFlag) ([]SimulatedTransaction, error)
 }
 
 var _ api = &Provider{}
--- a/rpc/trace.go
+++ b/rpc/trace.go
@@ -16,3 +16,15 @@ func (provider *Provider) TransactionTrace(ctx context.Context, hash string) err
 func (provider *Provider) TraceBlockTransactions(ctx context.Context, hash string) error {
 	return ErrNotImplemented
 }
+
+// simulate a given transaction on the requested state, and generate the execution trace
+func (provider *Provider) SimulateTransactions(ctx context.Context, blockID BlockID, txns []BroadcastedTransaction, simulationFlags []SimulationFlag) ([]SimulatedTransaction, error) {
+
+	var output []SimulatedTransaction
+	if err := do(ctx, provider.c, "starknet_simulateTransactions", &output, blockID, txns, simulationFlags); err != nil {
+		return nil, tryUnwrapToRPCErr(err, ErrContractNotFound, ErrContractError, ErrBlockNotFound)
+	}
+
+	return output, nil
+
+}
--- a/rpc/trace_test.go
+++ b/rpc/trace_test.go
@@ -2,7 +2,11 @@ package rpc
 
 import (
 	"context"
+	"encoding/json"
+	"os"
 	"testing"
+
+	"github.com/test-go/testify/require"
 )
 
 // TestNotImplemented checks the method not yet implemented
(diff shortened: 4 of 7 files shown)


rpc/trace.go
```diff
@@ -16,3 +17,29 @@ func (provider *Provider) TransactionTrace(ctx context.Context, hash string) err
...
+// simulate a given transaction on the requested state, and generate the execution trace
+func (provider *Provider) SimulateTransactions(
+	ctx context.Context,
+	blockId BlockID,
+	txns []BroadcastedTransaction,
+	simulationFlags []SimulationFlag) ([]SimulatedTransaction, error) {
+
+	var output []SimulatedTransaction
+	if err := do(ctx, provider.c, "starknet_simulateTransactions", &output, blockId, txns, simulationFlags); err != nil {
+		fmt.Println("err", err)
```
cicr99: This can be removed

rpc/trace.go
```diff
@@ -16,3 +17,29 @@ func (provider *Provider) TransactionTrace(ctx context.Context, hash string) err
 func (provider *Provider) TraceBlockTransactions(ctx context.Context, hash string) error {
 	return ErrNotImplemented
 }
+
+// simulate a given transaction on the requested state, and generate the execution trace
+func (provider *Provider) SimulateTransactions(
+	ctx context.Context,
+	blockId BlockID,
+	txns []BroadcastedTransaction,
+	simulationFlags []SimulationFlag) ([]SimulatedTransaction, error) {
```
cicr99: Could we leave this in just one line? I think it would be best to be consistent with the rest of the code.

We could use a tool for code formatting and add it to the workflow, so it makes sure the code is uniformly formatted, wyt?

rpc/types_trace.go
```diff
@@ -0,0 +1,104 @@
...
+}
+
+type SimulationFlag string
+
+const (
+	SKIP_VALIDATE SimulationFlag = "SKIP_VALIDATE"
+	SKIP_EXECUTE  SimulationFlag = "SKIP_EXECUTE"
+)
+
+// The execution trace and consuemd resources of the required transactions
```
cicr99: typo on _consuemd -> consumed_

rpc/types_trace.go
```diff
@@ -0,0 +1,104 @@
...
+// the execution trace of an L1 handler transaction
+type L1HandlerTxnTrace struct {
+	//the trace of the __execute__ call or constructor call, depending on the transaction type (none for declare transactions)
+	FunctionInvocation FnInvocation `json:"function_invocation"`
+}
+
+type EntryPointType string
+
+const (
+	Eternal     EntryPointType = "EXTERNAL"
```
cicr99: typo on _Eternal -> External_

rpc/types_trace.go
```diff
@@ -0,0 +1,104 @@
...
+	L1Handler   EntryPointType = "L1_HANDLER"
+	Constructor EntryPointType = "CONSTRUCTOR"
+)
+
+type CallType string
+
+const (
+	LibraryCall EntryPointType = "LIBRARY_CALL"
+	Call        EntryPointType = "CALL"
```
cicr99: These should be of CallType

rpc/types_trace.go
```diff
@@ -0,0 +1,104 @@
...
+	CallType CallType `json:"call_type"`
+
+	//The value returned from the function invocation
+	Result []*felt.Felt `json:"result"`
+
+	// The calls made by this invocation
+	NestedCalls []NestedCall `json:"calls"`
+
+	// The events emitted in this invocation
+	InvocationEvent []Event `json:"events"`
```
cicr99: Can we name this InvocationEvent**s** instead?

rpc/types_trace.go
```diff
@@ -0,0 +1,104 @@
...
+
+var _ TxnTrace = InvokeTxnTrace{}
+var _ TxnTrace = DeclareTxnTrace{}
+var _ TxnTrace = DeployAccountTxnTrace{}
+var _ TxnTrace = L1HandlerTxnTrace{}
+
+// the execution trace of an invoke transaction
+type InvokeTxnTrace struct {
+	ValidateInvocation    FnInvocation `json:"validate_invocation"`
+	ExecuteInvocation     FnInvocation `json:"execute_invocation"`
```
cicr99, line 40:
```suggestion
        //the trace of the __execute__ call or constructor call, depending on the transaction type (none for declare transactions)
	ExecuteInvocation     FnInvocation `json:"execute_invocation"`
```

rpc/trace.go
```diff
@@ -16,3 +17,29 @@ func (provider *Provider) TransactionTrace(ctx context.Context, hash string) err
...
+	if err := do(ctx, provider.c, "starknet_simulateTransactions", &output, blockId, txns, simulationFlags); err != nil {
+		fmt.Println("err", err)
+		switch {
+		case errors.Is(err, ErrContractNotFound):
+			return nil, ErrContractNotFound
+		case errors.Is(err, ErrContractError):
+			return nil, ErrContractError
+		case errors.Is(err, ErrBlockNotFound):
+			return nil, ErrBlockNotFound
+		default:
```
kirugan: I checked the codebase, this pattern is quite common (25 lines in 5 files), wdyt about moving it in separate function:
```
func unwrapErr(err error, rpcErrors ...*RPCError) *RPCError {
	for _, rpcErr := range rpcErrors {
		if errors.Is(err, rpcErr) {
			return rpcErr
		}
	}

	return nil
}
```

rpc/trace_test.go
```diff
@@ -34,3 +38,49 @@ func TestNotImplemented(t *testing.T) {
...
+	testConfig := beforeEach(t)
+
+	var simulateTxIn SimulateTransactionInput
+	var expectedResp SimulateTransactionOutput
+	if testEnv == "mainnet" {
+		simulateTxnRaw, err := os.ReadFile("./tests/simulateInvokeTx.json")
+		require.Nil(t, err, "Error ReadFile simulateInvokeTx")
+
+		err = json.Unmarshal(simulateTxnRaw, &simulateTxIn)
+		require.Nil(t, err, "Error unmarshalling simulateInvokeTx")
```
kirugan: maybe `require.NoError(...)` here and few lines above?

kirugan: Also do you think additional message is helpful? If test fails it prints stacktrace which leads to line that caused it to fail

rpc/provider.go
```diff
@@ -47,6 +47,7 @@ type api interface {
 	AddInvokeTransaction(ctx context.Context, broadcastedInvoke BroadcastedInvokeTransaction) (*AddInvokeTransactionResponse, error)
 	AddDeclareTransaction(ctx context.Context, declareTransaction BroadcastedDeclareTransaction) (*AddDeclareTransactionResponse, error)
 	AddDeployAccountTransaction(ctx context.Context, deployAccountTransaction BroadcastedDeployAccountTransaction) (*AddDeployAccountTransactionResponse, error)
+	SimulateTransactions(ctx context.Context, blockId BlockID, txns []BroadcastedTransaction, simFlags []SimulationFlag) ([]SimulatedTransaction, error)
```
kirugan: naming `blockID` is more common in the codebase

rpc/trace_test.go
rianhughes: Personally I find these comments help me quickly identify what went wrong and where the error arose. I can remove them if you it would be better to do so

rpc/types_trace.go
```diff
@@ -0,0 +1,104 @@
...
+
+	EntryPointType EntryPointType `json:"entry_point_type"`
+
+	CallType CallType `json:"call_type"`
+
+	//The value returned from the function invocation
+	Result []*felt.Felt `json:"result"`
+
+	// The calls made by this invocation
+	NestedCalls []NestedCall `json:"calls"`
```
kirugan: You can make recursive type i.e. refer in type to itself like:
```
NestedCalls []FnInvocation
```

rpc/trace_test.go
kirugan: Got it, no need - if they help to identify problem quickly let's keep them.

rpc/trace.go
rianhughes: This is very nice, thanks

rpc/errors.go
```diff
@@ -1,5 +1,17 @@
...
+import "errors"
+
+func unwrapErr(err error, rpcErrors ...*RPCError) *RPCError {
+	for _, rpcErr := range rpcErrors {
+		if errors.Is(err, rpcErr) {
+			return rpcErr
+		}
+	}
+
+	return nil
```
cicr99: Shouldn't this return `err`?

cicr99 (approved):

PR #319, opened 2023-09-12
Rpcv03 starknet trace block transactions

### LINKED ISSUE #315: starknet_traceBlockTransactions
implement https://github.com/starkware-libs/starknet-specs/blob/v0.3.0/api/starknet_trace_api_openrpc.json#L111

rianhughes:
closes https://github.com/NethermindEth/starknet.go/issues/315

### DIFF
--- a/rpc/mock_test.go
+++ b/rpc/mock_test.go
@@ -5,6 +5,7 @@ import (
 	"encoding/json"
 	"fmt"
 	"math/big"
+	"os"
 	"strings"
 
 	"github.com/NethermindEth/juno/core/felt"
@@ -66,6 +67,8 @@ func (r *rpcMock) CallContext(ctx context.Context, result interface{}, method st
 		return mock_starknet_estimateFee(result, method, args...)
 	case "starknet_getBlockWithTxHashes":
 		return mock_starknet_getBlockWithTxHashes(result, method, args...)
+	case "starknet_traceBlockTransactions":
+		return mock_starknet_traceBlockTransactions(result, method, args...)
 	default:
 		return errNotFound
 	}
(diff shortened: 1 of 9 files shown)


rpc/mock_test.go
```diff
@@ -623,3 +626,41 @@ func mock_starknet_getBlockWithTxHashes(result interface{}, method string, args
...
+			return err
+		}
+		BlockTrace, err := json.Marshal(rawBlockTrace.Result)
+		if err != nil {
+			return err
+		}
+		if nil != json.Unmarshal(BlockTrace, &r) {
+			return err
+		}
+		return nil
```
kirugan: let's simplify to `return json.Unmarshal(BLockTrace, &r)`

rpc/provider.go
```diff
@@ -48,6 +48,7 @@ type api interface {
 	AddDeclareTransaction(ctx context.Context, declareTransaction BroadcastedDeclareTransaction) (*AddDeclareTransactionResponse, error)
 	AddDeployAccountTransaction(ctx context.Context, deployAccountTransaction BroadcastedDeployAccountTransaction) (*AddDeployAccountTransactionResponse, error)
 	SimulateTransactions(ctx context.Context, blockID BlockID, txns []BroadcastedTransaction, simFlags []SimulationFlag) ([]SimulatedTransaction, error)
+	TraceBlockTransactions(ctx context.Context, blockHash *felt.Felt) (*[]Trace, error)
```
kirugan: is it possible to return just `[]Trace` not a pointer to slice?

rpc/mock_test.go
```diff
@@ -623,3 +626,41 @@ func mock_starknet_getBlockWithTxHashes(result interface{}, method string, args
...
+	if !ok || r == nil {
+		return errWrongType
+	}
+	if len(args) != 1 {
+		return errWrongArgs
+	}
+	blockHash, ok := args[0].(*felt.Felt)
+	if !ok {
+		fmt.Printf("args[0] should be felt, got %T\n", args[0])
+		return errWrongArgs
```
kirugan: wdty about wrapping this error i.e. errors.Wrap(err, "...msg above..."

rpc/provider.go
rianhughes: The traces can be quite large (eg a few MB for mainnet), so I thought it would be more efficient to return a pointer. Wdyt?

rianhughes: Leaving a note for any future readers, Go slices don't copy the entire data because internally they use pointers.

cicr99 (approved):
lgtm

PR #328, opened 2023-09-13
Implement starknet_traceTransaction method

### LINKED ISSUE #312: starknet_traceTransaction
implement https://github.com/starkware-libs/starknet-specs/blob/v0.3.0/api/starknet_trace_api_openrpc.json#L11C22-L11C47

cicr99:
Closes #312

### DIFF
--- a/rpc/errors.go
+++ b/rpc/errors.go
@@ -30,6 +30,10 @@ var (
 		code:    1,
 		message: "Failed to write transaction",
 	}
+	ErrNoTraceAvailable = &RPCError{
+		code: 10,
+        message: "No trace available for transaction",
+	}
 	ErrContractNotFound = &RPCError{
 		code:    20,
 		message: "Contract not found",
@@ -37,6 +41,10 @@ var (
 	ErrBlockNotFound = &RPCError{
 		code:    24,
 		message: "Block not found",
+	}
+	ErrInvalidTxnHash = &RPCError{
+		code:    25,
+		message: "Invalid transaction hash",
 	}
 	ErrHashNotFound = &RPCError{
 		code:    25,
--- a/rpc/errors.go
+++ b/rpc/errors.go
@@ -2,6 +2,8 @@ package rpc
 
 import "errors"
 
+var ErrNotImplemented = errors.New("not implemented")
+
 func tryUnwrapToRPCErr(err error, rpcErrors ...*RPCError) error {
 	for _, rpcErr := range rpcErrors {
 		if errors.Is(err, rpcErr) {
--- a/rpc/errors.go
+++ b/rpc/errors.go
@@ -14,9 +14,25 @@ func tryUnwrapToRPCErr(err error, rpcErrors ...*RPCError) error {
 	return err
 }
 
+func isErrNoTraceAvailableError(err error) (*RPCError, bool) {
+	clientErr, ok := err.(*RPCError)
+	if !ok {
+		return nil, false
+	}
+	switch clientErr.code {
+	case ErrNoTraceAvailable.code:
+		noTraceAvailableError := ErrNoTraceAvailable
+		noTraceAvailableError.data = clientErr.data
+		return noTraceAvailableError, true
+	}
+	return nil, false
+}
+
+
 type RPCError struct {
 	code    int
 	message string
+	data    any
 }
 
 func (e *RPCError) Error() string {
@@ -25,6 +41,10 @@ func (e *RPCError) Error() string {
 
 func (e *RPCError) Code() int {
 	return e.code
+}
+
+func (e *RPCError) Data() any {
+	return e.data
 }
 
 var (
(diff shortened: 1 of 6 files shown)


rpc/errors.go
```diff
@@ -30,6 +32,10 @@ var (
 		code:    1,
 		message: "Failed to write transaction",
 	}
+	ErrNoTraceAvailable = &RPCError{
```
rianhughes, line 55: The data field needs added to the errors: https://github.com/starkware-libs/starknet-specs/blob/v0.3.0/api/starknet_trace_api_openrpc.json#L330

rpc/trace_test.go
```diff
@@ -11,31 +11,44 @@ import (
...
+		}
+		expectedrespRaw, err := os.ReadFile("./tests/0xff66e14fc6a96f3289203690f5f876cb4b608868e8549b5f6a90a21d4d6329.json")
+		require.NoError(t, err, "Error ReadFile for TestTraceTransaction")
+
+		err = json.Unmarshal(expectedrespRaw, &rawjson)
+		require.NoError(t, err, "Error unmarshalling testdata TestTraceTransaction")
+		expectedResp = rawjson.Result
+	}
+
 	type testSetType struct {
```
rianhughes, line 34: Might be worth testing the errors

rianhughes:
Also, the rpc tests are failing

rpc/trace.go
```diff
@@ -2,16 +2,63 @@ package rpc
...
 	"context"
-	"errors"
+	"encoding/json"
 
 	"github.com/NethermindEth/juno/core/felt"
 )
 
-var ErrNotImplemented = errors.New("not implemented")
+// For a given executed transaction, return the trace of its execution, including internal calls
+func (provider *Provider) TransactionTrace(ctx context.Context, transactionHash *felt.Felt) (TxnTrace, error) {
```
rianhughes, line 11: The return type should be *RPCError. 

When you return `noTraceAvailableError` in this function, what's actually returned is `noTraceAvailableError.Error()` (so we lose the code and data information).

rpc/trace_test.go
```diff
@@ -11,30 +11,63 @@ import (
...
 	for _, test := range testSet {
-		var out string
-		err := do(context.Background(), testConfig.provider.c, test.MissingMethod, &out)
-
-		if err == nil || err.Error() != "Method Not Found" {
-			t.Fatalf("Method %s is now available, got %v\n", test.MissingMethod, err)
+		resp, err := testConfig.provider.TransactionTrace(context.Background(), test.TransactionHash)
+		if err != nil {
+			require.Equal(t, test.ExpectedError, err)
+		} else {
```
rianhughes, line 69: Might be neater without the else

rpc/trace_test.go
```diff
@@ -11,30 +11,63 @@ import (
...
 
 	for _, test := range testSet {
-		var out string
-		err := do(context.Background(), testConfig.provider.c, test.MissingMethod, &out)
-
-		if err == nil || err.Error() != "Method Not Found" {
-			t.Fatalf("Method %s is now available, got %v\n", test.MissingMethod, err)
+		resp, err := testConfig.provider.TransactionTrace(context.Background(), test.TransactionHash)
+		if err != nil {
+			require.Equal(t, test.ExpectedError, err)
```
rianhughes, line 68: Note that you're comparing the `error` type, not `RPCError` (so the data and code fields aren't actually being checked)

rpc/errors.go
```diff
@@ -2,6 +2,8 @@ package rpc
 
 import "errors"
 
+var ErrNotImplemented = errors.New("not implemented")
```
rianhughes, line 5: I don't think this error is really needed anymore

rpc/trace_test.go
cicr99: This is funny! I pointed out the opposite over here https://github.com/NethermindEth/starknet.go/pull/332#discussion_r1327199608

rpc/trace.go
cicr99: I'm not sure that's right.
The mock_test method for this pr returns an `error` as well and in this [line](https://github.com/NethermindEth/starknet.go/pull/328/files#diff-617ac3083426917993255a7947979919a2700fce3d4d158586d30561ee91e396R697) I return an RPCError.  When I tried in the test to compare the resulting error and the expected one it also considered the other fields, besides the error message I mean. In fact it failed becuase the data field was different

rpc/trace_test.go
cicr99: I think this has to do with this same comment https://github.com/NethermindEth/starknet.go/pull/328#discussion_r1327217983. Let's settle that and then revisit this?

rpc/errors.go
cicr99: I thought so too when I removed it from the `trace.go` file, but turns out is also being used in `block_test.go`. The best place to put it was in the `errors.go` file

rpc/trace_test.go
rianhughes: Lol, I'm happy either way. Guess we all have our own styles :)

rpc/trace.go
rianhughes: You're right!! My bad

rianhughes (approved):

PR #332, opened 2023-09-14
rpcv4_write_errors

### LINKED ISSUE #322: Fix errors for `starknet_addInvokeTransaction` and `starknet_addDeclareTransaction`
- Errors for `starknet_addInvokeTransaction`: https://github.com/starkware-libs/starknet-specs/blob/6fadca1f1abf48523c0aa799a10061f40f973468/api/starknet_write_api.json#L36
- Errors for `starknet_addDeclareTransaction`: https://github.com/starkware-libs/starknet-specs/blob/6fadca1f1abf48523c0aa799a10061f40f973468/api/starknet_write_api.json#L101

rianhughes:
closes https://github.com/NethermindEth/starknet.go/issues/322

### DIFF
--- a/rpc/provider.go
+++ b/rpc/provider.go
@@ -45,8 +45,8 @@ type api interface {
 	TransactionByHash(ctx context.Context, hash *felt.Felt) (Transaction, error)
 	TransactionReceipt(ctx context.Context, transactionHash *felt.Felt) (TransactionReceipt, error)
 	AddInvokeTransaction(ctx context.Context, broadcastedInvoke BroadcastedInvokeTransaction) (*AddInvokeTransactionResponse, *RPCError)
-	AddDeclareTransaction(ctx context.Context, declareTransaction BroadcastedDeclareTransaction) (*AddDeclareTransactionResponse, error)
-	AddDeployAccountTransaction(ctx context.Context, deployAccountTransaction BroadcastedDeployAccountTransaction) (*AddDeployAccountTransactionResponse, error)
+	AddDeclareTransaction(ctx context.Context, declareTransaction BroadcastedDeclareTransaction) (*AddDeclareTransactionResponse, *RPCError)
+	AddDeployAccountTransaction(ctx context.Context, deployAccountTransaction BroadcastedDeployAccountTransaction) (*AddDeployAccountTransactionResponse, *RPCError)
 	SimulateTransactions(ctx context.Context, blockID BlockID, txns []BroadcastedTransaction, simFlags []SimulationFlag) ([]SimulatedTransaction, error)
 	TraceBlockTransactions(ctx context.Context, blockHash *felt.Felt) ([]Trace, error)
 }
--- a/rpc/provider.go
+++ b/rpc/provider.go
@@ -44,7 +44,7 @@ type api interface {
 	TransactionByBlockIdAndIndex(ctx context.Context, blockID BlockID, index uint64) (Transaction, error)
 	TransactionByHash(ctx context.Context, hash *felt.Felt) (Transaction, error)
 	TransactionReceipt(ctx context.Context, transactionHash *felt.Felt) (TransactionReceipt, error)
-	AddInvokeTransaction(ctx context.Context, broadcastedInvoke BroadcastedInvokeTransaction) (*AddInvokeTransactionResponse, error)
+	AddInvokeTransaction(ctx context.Context, broadcastedInvoke BroadcastedInvokeTransaction) (*AddInvokeTransactionResponse, *RPCError)
 	AddDeclareTransaction(ctx context.Context, declareTransaction BroadcastedDeclareTransaction) (*AddDeclareTransactionResponse, error)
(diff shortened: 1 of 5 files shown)


rpc/write.go
```diff
@@ -2,51 +2,82 @@ package rpc
...
 
 // AddInvokeTransaction estimates the fee for a given Starknet transaction.
-func (provider *Provider) AddInvokeTransaction(ctx context.Context, broadcastedInvoke BroadcastedInvokeTransaction) (*AddInvokeTransactionResponse, error) {
-	// TODO: EntryPointSelector now part of calldata
-	// tx, ok := broadcastedInvoke.(BroadcastedInvokeV0Transaction)
-	// if ok {
-	// 	tx.EntryPointSelector = fmt.Sprintf("0x%x", types.GetSelectorFromName(tx.EntryPointSelector))
-	// 	broadcastedInvoke = tx
-	// }
+func (provider *Provider) AddInvokeTransaction(ctx context.Context, broadcastedInvoke BroadcastedInvokeTransaction) (*AddInvokeTransactionResponse, *RPCError) {
```
cicr99: Why use `BroadcastedInvokeTransaction` instead of `InvokeTxnV1` as in the [spec](https://github.com/starkware-libs/starknet-specs/blob/6fadca1f1abf48523c0aa799a10061f40f973468/api/starknet_write_api.json#L19)?

rpc/write.go
```diff
@@ -2,51 +2,82 @@ package rpc
...
+			return nil, tryUnwrapToRPCErr(
+				err,
+				ErrInsufficientAccountBalance,
+				ErrInsufficientMaxFee,
+				ErrInvalidTransactionNonce,
+				ErrValidationFailure,
+				ErrNonAccount,
+				ErrDuplicateTx,
+				ErrUnsupportedTxVersion,
+				ErrUnexpectedError,
```
cicr99: I think you can remove `ErrUnexpectedError` from this list, it was already checked in the condition above

rpc/write.go
```diff
@@ -2,51 +2,82 @@ package rpc
...
+			ErrInsufficientAccountBalance,
+			ErrInsufficientMaxFee,
+			ErrInvalidTransactionNonce,
+			ErrValidationFailure,
+			ErrNonAccount,
+			ErrDuplicateTx,
+			ErrContractClassSizeTooLarge,
+			ErrUnsupportedTxVersion,
+			ErrUnsupportedContractClassVersion,
+			ErrUnexpectedError,
```
cicr99: Same here about `ErrUnexpectedError`

rpc/write.go
```diff
@@ -2,51 +2,82 @@ package rpc
...
+			err,
+			ErrInsufficientAccountBalance,
+			ErrInsufficientMaxFee,
+			ErrInvalidTransactionNonce,
+			ErrValidationFailure,
+			ErrNonAccount,
+			ErrClassHashNotFound,
+			ErrDuplicateTx,
+			ErrUnsupportedTxVersion,
+			ErrUnexpectedError,
```
cicr99: Same here about `ErrUnexpectedError`

rpc/write_test.go
```diff
@@ -59,3 +59,45 @@ func TestDeclareTransaction(t *testing.T) {
...
+			},
+		},
+		"testnet": {},
+	}[testEnv]
+
+	for _, test := range testSet {
+		resp, err := testConfig.provider.AddInvokeTransaction(context.Background(), test.InvokeTx)
+		if err != nil {
+			require.Equal(t, *err, test.ExpectedError, "AddInvokeTransaction did not give expected error")
+			continue
```
cicr99: Instead of using continue here, I would rather have it as an if else statement, in other words:
```
if err != nil {
	require.Equal(t, *err, test.ExpectedError, "AddInvokeTransaction did not give expected error")
} else {
	require.Equal(t, *resp, test.ExpectedResp)
}
```
I think it would be more readable that way and we already have it like this in other places of the code

rpc/write.go
rianhughes: The removal of Broadcast types is addressed in a separate PR. We'll can address it via merge conflict - depending which PR gets approved and merged first

cicr99: I see. There is a [comment](https://github.com/NethermindEth/starknet.go/pull/290#issue-1859560594) in #290 (Remove BroadcastedTxn) to merge it last on account of all the merge conflicts it will generate. Let's do that, we'll merge this PR first and leave #290 for last.

rpc/provider.go
```diff
@@ -49,7 +49,6 @@ type api interface {
 	TransactionByBlockIdAndIndex(ctx context.Context, blockID BlockID, index uint64) (Transaction, error)
 	TransactionByHash(ctx context.Context, hash *felt.Felt) (Transaction, error)
 	TransactionReceipt(ctx context.Context, transactionHash *felt.Felt) (TransactionReceipt, error)
-	TransactionTrace(ctx context.Context, transactionHash *felt.Felt) (TxnTrace, error)
```
cicr99: Why is this being removed?

rianhughes: This is weird. It should have been included in the "merge main" commit but seemed to have been dropped?? I'll add it back.

cicr99 (approved):

Issue #318, opened 2023-09-11
use unwrapErr trhoughout the codebase


rianhughes:
Certain rpc methods use a switch statement to return rpc errors. However, we should use unwrapErr().

depends on https://github.com/NethermindEth/starknet.go/pull/316

cicr99:
The function to be used is now called `tryUnwrapToRPCErr` which can be found [here](https://github.com/NethermindEth/starknet.go/blob/a3a745a7309046e1fdbd1d649c867c58759617ca/rpc/errors.go#L32C6-L32C23)

PR #346, opened 2023-09-20
Refactor to use use unwrapErr() throughout the codebase

Bitcoinnoobie:
with related to issue #318


stranger80:
@Bitcoinnoobie I took liberty to rename this PR for readability

rianhughes (approved):

rianhughes:
This [issue](https://github.com/NethermindEth/starknet.go/issues/361) should be resolved before this PR is merged. We should run tests on PRs

cicr99:
There are a few places (other that the ones that were addressed) where no `switch` statement is used but it would be good to handle the errors using `tryUnwrapToRPCErr` method as well:
- [BlockHashAndNumber](https://github.com/NethermindEth/starknet.go/blob/8906950018b4a6d76fcaecc8499c76ad6f679d99/rpc/block.go#L2)
-  [BlockWithTxHashes](https://github.com/NethermindEth/starknet.go/blob/8906950018b4a6d76fcaecc8499c76ad6f679d99/rpc/block.go#L56) 
- [StateUpdate](https://github.com/NethermindEth/starknet.go/blob/8906950018b4a6d76fcaecc8499c76ad6f679d99/rpc/block.go#L79)
- [BlockWithTxs](https://github.com/NethermindEth/starknet.go/blob/8906950018b4a6d76fcaecc8499c76ad6f679d99/rpc/block.go#L103)
- [TransactionByHash](https://github.com/NethermindEth/starknet.go/blob/8906950018b4a6d76fcaecc8499c76ad6f679d99/rpc/transaction.go#L66)
- [TransactionReceipt](https://github.com/NethermindEth/starknet.go/blob/8906950018b4a6d76fcaecc8499c76ad6f679d99/rpc/transaction.go#L103)

Could these be modified too?

cicr99:
> This [issue](https://github.com/NethermindEth/starknet.go/issues/361) should be resolved before this PR is merged. We should run tests on PRs

Also, this was already merged in, but in order to be applied to this PR I think it's necessary to fetch the latest changes from main into this branch

cicr99 (approved):

PR #349, opened 2023-09-21
Create deployAccount example using rpc

rianhughes:


examples/deployAccount/main.go
```diff
@@ -0,0 +1,204 @@
...
+
+	// Get keys
+	pub, priv := getRandomKeys()
+
+	classHash, err := utils.HexToFelt(predeployedClassHash)
+	if err != nil {
+		panic(err)
+	}
+
+	tx := rpc.BroadcastedDeployAccountTransaction{
```
cicr99: This type was removed in a previous PR, I guess it should be `DeployAccountTxn` instead

cicr99 (approved):

PR #354, opened 2023-09-22
Block with tx hashes

### LINKED ISSUE #347: Block with Txns and Block with TxnHashs
Transaction types no longer contain a TransactionHash field in rpcv04 which means the current Block type cant be used for both BlockWithTxns, and BlockWithTxnHashes. So we should return rpc.Block for BlockWithTxs(), and need to implement a new type rpc.BlockWithTxHashs for BlockWithTxHashes().

Note this depends on:https://github.com/NethermindEth/starknet.go/pull/290

cicr99:
Closes #347


rpc/block_test.go
```diff
@@ -93,50 +93,60 @@ func TestBlockWithTxHashes(t *testing.T) {
...
+					Transactions:     txHashes,    
+				},
 			},
-		},
 			{
 				BlockID: BlockID{Hash: &felt.Zero},
-				ExpectedBlockWithTxHashes: &Block{
+				ExpectedBlockWithTxHashes: &BlockTxHashes{
 					BlockHeader: BlockHeader{
+						BlockHash:        &felt.Zero,
```
rianhughes: Setting the BlockHash to Zero here means that the Block isn't actually checked in the test in lines [194-196](https://github.com/NethermindEth/starknet.go/pull/354/files#diff-6e6821e21ccb8985a9ddf2d9eb0f9a494f729218e6d04a70247837a19431e3a0R194-R196). 

It might be better to set it to some other value (you'll probably have to update the mock function as well)

rpc/block_test.go
```diff
@@ -163,11 +173,11 @@ func TestBlockWithTxHashes(t *testing.T) {
 		if err != test.ExpectedError {
 			t.Fatal("BlockWithTxHashes match the expected error:", err)
 		}
-		switch resultBlock := result.(type) {
-		case Block:
-			block, ok := result.(*Block)
+		switch result.(type) {
+		case BlockTxHashes:
```
rianhughes: Also, I don't think this case is actually being triggered for the mock test

rpc/block_test.go
```diff
@@ -163,11 +173,11 @@ func TestBlockWithTxHashes(t *testing.T) {
 		if err != test.ExpectedError {
 			t.Fatal("BlockWithTxHashes match the expected error:", err)
 		}
-		switch resultBlock := result.(type) {
-		case Block:
-			block, ok := result.(*Block)
+		switch result.(type) {
```
rianhughes: Also, we should add a default. This method returns an interface, which could be anything. We should check for this

rpc/block.go
```diff
@@ -51,7 +51,7 @@ func WithBlockTag(tag string) BlockID {
 
 // BlockWithTxHashes gets block information given the block id.
 func (provider *Provider) BlockWithTxHashes(ctx context.Context, blockID BlockID) (interface{}, error) {
```
rianhughes, line 53: What do you think about returning a concrete type here?

rianhughes: Actually, nvm, I don't think I'm correct here. Currently the function should either return BlockTxHashes or a PendingBlocktxHashes

rpc/block_test.go
rianhughes: The reason is the method returns a pointer for BlockTxHashes, and a non-pointer for PendingTxHashes

rpc/block.go
```diff
@@ -61,7 +61,7 @@ func (provider *Provider) BlockWithTxHashes(ctx context.Context, blockID BlockID
 
 	// if header.Hash == nil it's a pending block
 	if result.BlockHeader.BlockHash == nil {
-		return PendingBlock{
+		return PendingBlockTxHashes{
```
rianhughes: We return PendingBlockTxHashes (by value) here but a pointer (reference) in line 72. We should be consistent

rpc/block.go
cicr99: I left it this way to be similar to the other method we have `BlockWithTxs`. If we change this one I believe it's a good idea to change the other as well

rpc/block_test.go
cicr99: You are right! I added this because of the mock_test but forgot about that condition below

rpc/block.go
cicr99: I think this can be solved in another PR. I opened a new issue for it https://github.com/NethermindEth/starknet.go/issues/355

rianhughes (approved):

PR #377, opened 2023-10-03
Create contracts package and hash logic

### LINKED ISSUE #376: Create new contracts package and hashing logic
Users will need to compute things like the class_hash for both casm contracts and sierra contracts. To do this new casm types will need to be defined, which I've put in a new contracts package. The old contracts package has been deleted (the few methods defined there will be move over into the accounts package). We also need to implement new hashing functions to compute the class_hash's for both these contract types. I've created a new hash package to store these types (later the hashing logic from the starknet.go file should probably be moved into the hash package).

rianhughes:
closes: https://github.com/NethermindEth/starknet.go/issues/376


contracts/contracts.go
```diff
@@ -1,172 +1,50 @@
...
-
-type DeclareOutput struct {
-	classHash       string
-	transactionHash string
+type CasmClass struct {
+	Prime            string                     `json:"prime"`
+	Version          string                     `json:"compiler_version"`
+	ByteCode         []*felt.Felt               `json:"bytecode"`
+	EntryPointByType CasmClassEntryPointsByType `json:"entry_points_by_type"`
+	// Hints            any                        `json:"hints"`
```
cicr99, line 15: Why is this commented? Is it not necessary?

starknetgo.go
```diff
@@ -220,6 +221,26 @@ func (sc StarkCurve) PedersenHash(elems []*big.Int) (hash *big.Int, err error) {
...
+NOTE: This function just wraps the Juno implementation
+
+(ref: https://github.com/NethermindEth/juno/blob/main/core/crypto/poseidon_hash.go#L74)
+*/
+func (sc StarkCurve) PoseidonArray(felts ...*felt.Felt) *felt.Felt {
+	return junoCrypto.PoseidonArray(felts...)
+}
+
+/*
+Provides the starknet keccek hash .
```
cicr99: Small typo on keccek

contracts/contracts.go
rianhughes: It doesn't seem to be needed to compute the class hash, so I don't think we need it. Figured I would leave this comment in to highlight it's a part of the CasmClass.

artifacts/artifacts.go
```diff
@@ -8,6 +8,12 @@ import (
 	"github.com/NethermindEth/starknet.go/types"
 )
 
+//go:embed hello_starknet_compiled.casm.json
+var ExampleWorldCasm []byte
+
+//go:embed hello_starknet_compiled.sierra.json
+var ExampleWorldSierra []byte
```
kirugan: Seems like you use these variables only in tests. Most likely they will end up in a final binary even though they are not used there. You can verify it with `strings <binary> | grep <content from json>`

contracts/contracts.go
```diff
@@ -1,172 +1,50 @@
...
-	// 	TransactionHash: tx.TransactionHash.String(),
-	// }, nil
+func UnmarshalCasmClass(filePath string) (*CasmClass, error) {
+	file, err := os.Open(filePath)
+	if err != nil {
+		return nil, err
+	}
+	defer file.Close()
+
+	content, err := ioutil.ReadAll(file)
```
kirugan: you can replace both calls with just `os.ReadFile(filePath)`

contracts/contracts_test.go
```diff
@@ -1,130 +1,30 @@
...
-		}
-		fmt.Println("deployment transaction", tx.TransactionHash)
-	}
+func TestUnmarshalContractClass(t *testing.T) {
+	compiledClass := artifacts.ExampleWorldSierra
+	var class rpc.ContractClass
+	err := json.Unmarshal(compiledClass, &class)
+	require.NoError(t, err)
+	require.Equal(t, class.SierraProgram[0].String(), "0x1")
+	require.Equal(t, class.SierraProgram[1].String(), "0x3")
```
kirugan: Consider changing `require.Equal(...)` to `assert.Equal(...)` because `require` functions stop execution immediately if arguments are not equal, but `assert` will keep going. It's helpful to see in one run which fields are correct and which are not. 

In codebases usually `require.NoError(...)` is used instead of `assert.NoError(...)` because most likely other arguments will be nil and there is no point to keep going (facing nil pointer panic)

hash/hash.go
```diff
@@ -0,0 +1,95 @@
...
+
+import (
+	"github.com/NethermindEth/juno/core/felt"
+	starknetgo "github.com/NethermindEth/starknet.go"
+	"github.com/NethermindEth/starknet.go/contracts"
+	"github.com/NethermindEth/starknet.go/rpc"
+	"github.com/NethermindEth/starknet.go/utils"
+)
+
+// computeHashOnElementsFelt hashes the array of felts provided as input
```
kirugan:
```suggestion
// ComputeHashOnElementsFelt hashes the array of felts provided as input
```

hash/hash.go
```diff
@@ -0,0 +1,95 @@
...
+	if err != nil {
+		return nil, err
+	}
+
+	// https://docs.starknet.io/documentation/architecture_and_concepts/Network_Architecture/transactions/#deploy_account_hash_calculation
+	return starknetgo.Curve.PoseidonArray(ContractClassVersionHash, ExternalHash, L1HandleHash, ConstructorHash, ABIHash, SierraProgamHash), nil
+}
+
+func hashEntryPointByType(entryPoint []rpc.SierraEntryPoint) *felt.Felt {
+	flattened := []*felt.Felt{}
```
kirugan:
```suggestion
	flattened := make([]*felt.Felt{}, 0, len(entryPoint))
```
small optimization

hash/hash.go
```diff
@@ -0,0 +1,95 @@
...
+	bigIntArr := utils.FeltArrToBigIntArr(feltArr)
+
+	hash, err := starknetgo.Curve.ComputeHashOnElements(bigIntArr)
+	if err != nil {
+		return nil, err
+	}
+	return utils.BigIntToFelt(hash), nil
+}
+
+// calculateTransactionHashCommon [specification] calculates the transaction hash in the StarkNet network - a unique identifier of the transaction.
```
kirugan:
```suggestion
// CalculateTransactionHashCommon [specification] calculates the transaction hash in the StarkNet network - a unique identifier of the transaction.
```

hash/hash.go
```diff
@@ -0,0 +1,95 @@
...
+	L1HandleHash := hashCasmClassEntryPointByType(casmClass.EntryPointByType.L1Handler)
+	ConstructorHash := hashCasmClassEntryPointByType(casmClass.EntryPointByType.Constructor)
+	ByteCodeHasH := starknetgo.Curve.PoseidonArray(casmClass.ByteCode...)
+
+	// https://github.com/software-mansion/starknet.py/blob/development/starknet_py/hash/casm_class_hash.py#L10
+	return starknetgo.Curve.PoseidonArray(ContractClassVersionHash, ExternalHash, L1HandleHash, ConstructorHash, ByteCodeHasH)
+}
+
+func hashCasmClassEntryPointByType(entryPoint []contracts.CasmClassEntryPoint) *felt.Felt {
+	flattened := []*felt.Felt{}
```
kirugan:
```suggestion
	flattened := make([]*felt.Felt{}, 0, len(entryPoint))
```

artifacts/artifacts.go
rianhughes: I've moved them to new tests folders so they don't get included in the final binary

artifacts/artifacts.go
```diff
@@ -8,6 +8,12 @@ import (
...
+
 //go:embed account.json
 var AccountCompiled []byte
```
cicr99: As the old contract files are being removed it probably would be good to also remove some of the files from the artifacts that were only used by those files. This is one of them for example. Not sure if they could be used for the tests of the contract methods, but if they are, it would be just a matter of re-adding them

rianhughes: Good point, turns out we don't need the artifacts pkg anymore, so I've deleted it

contracts/contracts_test.go
```diff
@@ -1,130 +1,33 @@
-package contracts
+package contracts_test
```
cicr99, line 1: Any particular reason you used this here instead of just `contracts`? I saw you did so as well in `hash_test.go`

kirugan (approved):

cicr99 (approved):

PR #386, opened 2023-10-04
Test declare txn

### LINKED ISSUE #358: Add Test AddDeclareTxn to account package
We need to add a test for the AddDeclareTxn method. It would be nice to successfully submit a txn to the network using the accounts package. 

depends https://github.com/NethermindEth/starknet.go/pull/302

rianhughes:
closes https://github.com/NethermindEth/starknet.go/issues/358
depends https://github.com/NethermindEth/starknet.go/pull/377


rpc/write.go
```diff
@@ -25,6 +25,14 @@ func (provider *Provider) AddInvokeTransaction(ctx context.Context, invokeTxn In
 }
 
 func (provider *Provider) AddDeclareTransaction(ctx context.Context, declareTransaction AddDeclareTxnInput) (*AddDeclareTransactionResponse, error) {
+
+	switch txn := declareTransaction.(type) {
+	case DeclareTxnV2:
+		// DeclareTxnV2 should not have a populated class hash field. It is only needed for signing.
+		txn.ClassHash = nil
+		declareTransaction = txn
```
kirugan, line 33: Two questions:
1. Do these lines affect provided transaction from caller side? i.e. is it copy or reference
2. And same question about copy/reference in switch type assertion: `txn := declareTransaction.(type)` ?

rianhughes: Since `declareTransaction` points to a struct, we are working with a copy, so the data/transaction on the callers side won't be modified (in contrast to the case if we were passing my reference).

So what happens is 1) `declareTransaction` is a copy of the input, 2) `txn` is an copy of `declareTransaction`,3) we update  `txn`, and 4) point  `declareTransaction` to  `txn`.

kirugan: Got it, thanks

kirugan (approved):

PR #389, opened 2023-10-04
Implement Wait For Transaction Receipt in accounts

### LINKED ISSUE #387: Implement WaitForTransaction() method for accounts
The old contracts pkg is being deprecated / replaced, and WaitForTransaction will be moved into the accounts pkg

rianhughes:
closes https://github.com/NethermindEth/starknet.go/issues/387


account/account.go
```diff
@@ -293,6 +295,27 @@ func (account *Account) PrecomputeAddress(deployerAddress *felt.Felt, salt *felt
...
+	t := time.NewTicker(pollInterval)
+	for {
+		select {
+		case <-ctx.Done():
+			return nil, ctx.Err()
+		case <-t.C:
+			receipt, err := account.TransactionReceipt(ctx, transactionHash)
+			if err != nil {
+				if err.Error() == rpc.ErrHashNotFound.Error() {
+					continue
```
kirugan, line 309: Is it ok that in case of wrong tx hash it's going to poll over and over again?

account/account_test.go
```diff
@@ -530,6 +532,51 @@ func TestTransactionHashDeclare(t *testing.T) {
...
+	}[testEnv]
+
+	for _, test := range testSet {
+		mockRpcProvider.EXPECT().TransactionReceipt(context.Background(), test.Hash).Return(test.ExpectedReceipt, test.ExpectedErr)
+		resp, err := acnt.WaitForTransactionReceipt(context.Background(), test.Hash, 2*time.Second)
+
+		if err != nil {
+			require.Equal(t, test.ExpectedErr, err)
+		} else {
+			require.Equal(t, test.ExpectedReceipt.GetExecutionStatus(), (*resp).GetExecutionStatus())
```
kirugan, line 593: wdyt about making condition on expected behaviour rather than actual result:
```
if test.ExpectedErr != nil {
   ...
} else {
   ...
}
```
Pros is that if implementation for some reason will return nil error and it will be unexpected by your testdata (i.e. test.ExpectedReceipt will be nil) then line 574 will throw nil pointer dereference panic instead of asserting equality

account/account_test.go
```diff
@@ -530,6 +532,51 @@ func TestTransactionHashDeclare(t *testing.T) {
 	require.Equal(t, expectedHash.String(), hash.String(), "TransactionHashDeclare not what expected")
 }
 
+func TestWaitForTransactionReceiptMOCK(t *testing.T) {
```
kirugan, line 537: Do we need tests for polling ?

account/account.go
rianhughes: I was wondering the same thing. I think it should loop until the context cancels. I was tempted to add an additional paramter to specify the timeout, but in principle the context should do this (afaik), so I decided to leave it as is.

account/account_test.go
rianhughes: Good point!

account/account_test.go
rianhughes: I've added a separate test for polling using devnet

account/account.go
rianhughes: I don't think there's a distinction between incorrect tx hash and a tx hash that hasn't been added to the chain. This is why users should specify a timeout in the context. Otherwise, as you say, it will poll indefinitely.

account/account_test.go
```diff
@@ -530,6 +532,107 @@ func TestTransactionHashDeclare(t *testing.T) {
...
+				Hash:                         new(felt.Felt).SetUint64(3),
+				ShouldCallTransactionReceipt: false,
+				ExpectedReceipt:              nil,
+				ExpectedErr:                  context.DeadlineExceeded,
+			},
+		},
+	}[testEnv]
+
+	for _, test := range testSet {
+		ctx, cancel := context.WithTimeout(context.Background(), time.Duration(test.Timeout)*time.Second)
```
kirugan: small enhancement proposal: to change type of timeout field to time.Duration (there would be no need to cast integer constants to time.Duration i.e. `Timeout: time.Duration(1)` because constants in Go untyped and autocasts to expected type if conversion is possible)

kirugan (approved):

cicr99 (approved):

PR #392, opened 2023-10-05
Simple Invoke example

Akashneelesh:


examples/simpleInvoke/main.go
```diff
@@ -0,0 +1,82 @@
...
+	fmt.Println("Starting simpeCall example")
+	c, err := ethrpc.DialContext(context.Background(), base)
+	if err != nil {
+		fmt.Println("Failed to connect to the client, did you specify the url in the .env.mainnet?")
+		panic(err)
+	}
+	clientv02 := rpc.NewProvider(c)
+	account_address, _ := utils.HexToFelt(account_addr)
+	ks := starknetgo.NewMemKeystore()
+	fakePrivKeyBI, _ := new(big.Int).SetString(privateKey, 0)
```
rianhughes: It would be good to handle the errors

examples/simpleInvoke/main.go
```diff
@@ -0,0 +1,82 @@
...
+	if err != nil {
+		panic(err.Error())
+	}
+	fmt.Printf("Response : ", resp)
+
+}
+
+// func main() {
+// 	ExecuteIncreaseValue()
+// }
```
rianhughes: This can be removed

examples/simpleInvoke/main.go
```diff
@@ -0,0 +1,82 @@
+package main
```
rianhughes: It would be good to add a Readme, outlining how to execute the example

examples/simpleInvoke/main.go
```diff
@@ -0,0 +1,82 @@
...
+
+	accnt, err := account.NewAccount(clientv02, account_address, public_key, ks)
+	if err != nil {
+		panic(err.Error())
+	}
+	accnt.BuildInvokeTx(context.Background(), &InvokeTx, &[]rpc.FunctionCall{FnCall})
+	txHash, err := accnt.TransactionHashInvoke(InvokeTx)
+	if err != nil {
+		panic(err.Error())
+	}
```
rianhughes: Why are you computing the transaction hash? If AddInvokeTransaction succeeds then the transaction hash will be in the response

examples/simpleInvoke/main.go
```diff
@@ -0,0 +1,82 @@
...
+	fmt.Printf("TxHash :", txHash)
+	resp, err := accnt.AddInvokeTransaction(context.Background(), InvokeTx)
```
rianhughes: Since this is a write-transaction, you'll need to sign it before you submit it to the network.

examples/simpleInvoke/main.go
Akashneelesh: Got it I'll remove that off.

examples/simpleInvoke/main.go
Akashneelesh: Sure I'll do that.

examples/simpleInvoke/main.go
Akashneelesh: Yep got it.

examples/simpleInvoke/main.go
Akashneelesh: I was referring to the testAddInvoke in the tests, and this was the same way of how it was used, there werent any signing done.

examples/simpleInvoke/main.go
Akashneelesh: Done

examples/simpleInvoke/main.go
```diff
@@ -0,0 +1,84 @@
...
+	privateKey     string = "0x0687bf84896ee63f52d69e6de1b41492abeadc0dc3cb7bd351d0a52116915937"
+	public_key     string = "0x58b0824ee8480133cad03533c8930eda6888b3c5170db2f6e4f51b519141963"
+	someContract   string = "0x4c1337d55351eac9a0b74f3b8f0d3928e2bb781e5084686a892e66d49d510d"
+	contractMethod string = "increase_value"
+)
+
+func main() {
+	godotenv.Load(fmt.Sprintf(".env.%s", name))
+	base := os.Getenv("INTEGRATION_BASE") //please modify the .env.testnet and replace the INTEGRATION_BASE with an starknet goerli RPC.
+	fmt.Println("Starting simpeCall example")
```
cicr99: Could it be _simple invoke example_ instead?

examples/simpleInvoke/main.go
```diff
@@ -0,0 +1,84 @@
...
+	c, err := ethrpc.DialContext(context.Background(), base)
+	if err != nil {
+		fmt.Println("Failed to connect to the client, did you specify the url in the .env.mainnet?")
```
cicr99: Should be `.env.testnet`, right?

examples/simpleInvoke/main.go
cicr99: I think the signing was happening in `BuildInvokeTx` method just before this line. However, this method was removed in PR #399, so this implementation will need updating

examples/simpleInvoke/main.go
```diff
@@ -0,0 +1,84 @@
...
+
+	maxfee, err := utils.HexToFelt("0x9184e72a000")
+	if err != nil {
+		panic(err.Error())
+	}
+
+	InvokeTx := rpc.InvokeTxnV1{
+		MaxFee:        maxfee,
+		Version:       rpc.TransactionV1,
+		Nonce:         new(felt.Felt).SetUint64(108), // You will have to modify this accordingly
```
rianhughes: Account has a nonce method that we should probably use instead of setting the nonce by hand

examples/simpleInvoke/main.go
```diff
@@ -0,0 +1,84 @@
...
+	"github.com/NethermindEth/starknet.go/rpc"
+	"github.com/NethermindEth/starknet.go/utils"
+	ethrpc "github.com/ethereum/go-ethereum/rpc"
+	"github.com/joho/godotenv"
+)
+
+var (
+	name           string = "testnet"
+	account_addr   string = "0x06f36e8a0fc06518125bbb1c63553e8a7d8597d437f9d56d891b8c7d3c977716"
+	privateKey     string = "0x0687bf84896ee63f52d69e6de1b41492abeadc0dc3cb7bd351d0a52116915937"
```
rianhughes: Can we add a comment to highlight that users should only add keys here for testing purposes. If they use their private key and it gets leaked then they will lose their funds.

examples/simpleInvoke/main.go
```diff
@@ -0,0 +1,84 @@
...
+		fmt.Println("Failed to connect to the client, did you specify the url in the .env.mainnet?")
+		panic(err)
+	}
+	clientv02 := rpc.NewProvider(c)
+	account_address, err := utils.HexToFelt(account_addr)
+	if err != nil {
+		panic(err.Error())
+	}
+	ks := starknetgo.NewMemKeystore()
+	fakePrivKeyBI, _ := new(big.Int).SetString(privateKey, 0)
```
rianhughes: We should handle the error

examples/simpleInvoke/main.go
Akashneelesh: Got it I'll update it.

examples/simpleInvoke/main.go
Akashneelesh: Yep I'll update this.

examples/simpleInvoke/main.go
Akashneelesh: Yes right.

examples/simpleInvoke/main.go
Akashneelesh: I have updated it now, which handles this.

examples/simpleInvoke/main.go
Akashneelesh: Sure I have updated it with the note

examples/simpleInvoke/main.go
Akashneelesh: I have updated it, instead of setting it manually, nonce is fetched through the account's nonce method

examples/simpleInvoke/main.go
Akashneelesh: I have updated it, along with all the necessary changes

examples/simpleInvoke/main.go
Akashneelesh: I have now updated it, with all the necessary implementation.

examples/simpleInvoke/main.go
Akashneelesh: I have now added a readme.

Akashneelesh:
I have now committed it with all the updated implementations and have covered all the comments that was mentioned previously.

examples/simpleInvoke/main.go
```diff
@@ -0,0 +1,117 @@
...
+
+	//Initializing the account
+	accnt, err := account.NewAccount(clientv02, account_address, public_key, ks)
+	if err != nil {
+		panic(err.Error())
+	}
+
+	//Getting the nonce from the account, and then converting it into felt
+	nonce_string, _ := accnt.Nonce(context.Background(), rpc.BlockID{Tag: "latest"}, accnt.AccountAddress)
+	nonce, _ := utils.HexToFelt(*nonce_string)
```
cicr99: We should handle the errors here as well

cicr99:
Tests are failing, it appears to be some kind of dependency issue, probably from the changes you are trying to commit in `go.mod`. We need them to be successful in order to merge the PR.
Also it would be good to merge main into this branch to get the latest changes

Akashneelesh:
> Tests are failing, it appears to be some kind of dependency issue, probably from the changes you are trying to commit in `go.mod`. We need them to be successful in order to merge the PR. Also it would be good to merge main into this branch to get the latest changes

Ohh fine

Akashneelesh:
> > Tests are failing, it appears to be some kind of dependency issue, probably from the changes you are trying to commit in `go.mod`. We need them to be successful in order to merge the PR. Also it would be good to merge main into this branch to get the latest changes
> 
> Ohh fine

I have done that and have updated, but still the build fails

PR #394, opened 2023-10-05
Create curve pkg and organize remaining files

### LINKED ISSUE #393: move all EC logic into a pkg

cicr99:
- Created new packages for:
    - `curve`
    - `typed`
    - `merkle`
- `providers_test` was removed as there was already pretty much the same test in the rpc package
- Example for curve was removed as it was no longer needed
- Fixed broken example for deploy account (because of the curve package and a previous change in BigIntToArr method

Closes #393


curve/utils.go
```diff
@@ -1,8 +1,9 @@
-package starknetgo
+package curve
```
rianhughes, line 1: I think we can merge utils into curve. The StarkCurve type is defined in curve, along with a collection of StarkCurve methods. Utils just defines some more methods on StarkCurve.

curve/utils_test.go
```diff
@@ -1,4 +1,4 @@
-package starknetgo
+package curve
```
rianhughes: utils_test can also be merged into curve_test

account/hash.go
```diff
@@ -2,15 +2,15 @@ package account
...
 import (
 	"github.com/NethermindEth/juno/core/felt"
-	starknetgo "github.com/NethermindEth/starknet.go"
+	"github.com/NethermindEth/starknet.go/curve"
 	"github.com/NethermindEth/starknet.go/rpc"
 	"github.com/NethermindEth/starknet.go/utils"
 )
 
 // computeHashOnElementsFelt hashes the array of felts provided as input
 func computeHashOnElementsFelt(feltArr []*felt.Felt) (*felt.Felt, error) {
```
rianhughes: We can delete computeHashOnElementsFelt and calculateTransactionHashCommon, and just use the functions define in the hash pkg. 

Also, we can maybe move FmtCalldata directly into accounts.go, and delete the hash file.

rianhughes (approved):

Issue #360, opened 2023-09-26
Account - fix failing devnet workflow 


rianhughes:
The devnet workflow test for the accounts package fails because the devnet-account doesnt have enough fee token to submit the transaction. This test runs and passes locally though.

depends https://github.com/NethermindEth/starknet.go/pull/302

PR #398, opened 2023-10-06
fix small bug

rianhughes:


rianhughes:
Test now fails as expected, and should be fixed in https://github.com/NethermindEth/starknet.go/issues/360

cicr99 (approved):

PR #399, opened 2023-10-09
Implement formatCallDAta for Cairo2

### LINKED ISSUE #390: Write example to solve user error
A user in the telegram group is having trouble calling a function in a contract. Write an example to resolve it. Once the contract address and function is shared it should be posted here for more details.

rianhughes:
closes https://github.com/NethermindEth/starknet.go/issues/390

### DIFF
--- a/account/account.go
+++ b/account/account.go
@@ -2,9 +2,7 @@ package account
 
 import (
 	"context"
-	"encoding/json"
 	"errors"
-	"fmt"
 	"time"
 
 	"github.com/NethermindEth/juno/core/felt"
@@ -453,7 +451,7 @@ func FmtCalldataCairo0(fnCalls []rpc.FunctionCall) []*felt.Felt {
 }
 
 /*
-Formats the call data for Cairo 1 contracs
+Formats the call data for Cairo 2 contracs
 */
 func FmtCalldataCairo2(fnCalls []rpc.FunctionCall) []*felt.Felt {
 	execCallData := []*felt.Felt{}
@@ -474,9 +472,4 @@ func FmtCalldataCairo2(fnCalls []rpc.FunctionCall) []*felt.Felt {
 	execCallData = append(execCallData, concatCallData...)
 
 	return execCallData
-}
-
-func PrintTxAsJson(tx any) {
-	qwe, err := json.MarshalIndent(tx, "", "")
-	fmt.Println(string(qwe), err)
 }
--- a/account/account.go
+++ b/account/account.go
@@ -2,7 +2,9 @@ package account
 
 import (
 	"context"
+	"encoding/json"
 	"errors"
+	"fmt"
 	"time"
 
 	"github.com/NethermindEth/juno/core/felt"
@@ -472,4 +474,9 @@ func FmtCalldataCairo2(fnCalls []rpc.FunctionCall) []*felt.Felt {
 	execCallData = append(execCallData, concatCallData...)
 
 	return execCallData
+}
+
+func PrintTxAsJson(tx any) {
+	qwe, err := json.MarshalIndent(tx, "", "")
+	fmt.Println(string(qwe), err)
 }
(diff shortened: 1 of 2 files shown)


account/account.go
```diff
@@ -420,26 +415,68 @@ func (account *Account) TransactionByHash(ctx context.Context, hash *felt.Felt)
 	return account.provider.TransactionByHash(ctx, hash)
 }
 
+func (account *Account) FmtCalldata(fnCalls []rpc.FunctionCall, cairoVersion int) ([]*felt.Felt, error) {
+	switch cairoVersion {
+	case 0:
+		return FmtCalldataCairo0(fnCalls), nil
+	case 2:
```
cicr99, line 420: Don't we need to handle cairo 1 version?

account/account.go
```diff
@@ -420,26 +415,68 @@ func (account *Account) TransactionByHash(ctx context.Context, hash *felt.Felt)
...
+
+	// Legacy : Cairo 0
+	concatCallData := []*felt.Felt{}
+	for _, fnCall := range fnCalls {
+		execCallData = append(
+			execCallData,
+			fnCall.ContractAddress,
+			fnCall.EntryPointSelector,
+			new(felt.Felt).SetUint64(uint64(len(concatCallData))),
+			new(felt.Felt).SetUint64(uint64(len(fnCall.Calldata))+1),
```
cicr99, line 442: Why is it needed to add one to the length? I thought just the length would give you the next position to write the calldata.

account/account.go
```diff
@@ -420,26 +415,68 @@ func (account *Account) TransactionByHash(ctx context.Context, hash *felt.Felt)
...
-			callData = append(callData, &felt.Zero, &felt.Zero)
-			continue
-		}
+	return execCallData
+}
 
-		callData = append(callData, new(felt.Felt).SetUint64(uint64(len(callArray))), new(felt.Felt).SetUint64(uint64(len(tx.Calldata))+1))
-		callArray = append(callArray, tx.Calldata...)
+/*
+Formats the call data for Cairo 1 contracs
```
cicr99: I suppose this should be _Cairo **2** contracts_

account/account.go
cicr99: I see this is the main difference between this implementation and the one for Cairo 2 contracts. Would be nice to know why, though.

account/account_test.go
```diff
@@ -396,15 +475,19 @@ func TestAddInvoke(t *testing.T) {
...
 
-		txHash, err := acnt.TransactionHashInvoke(test.InvokeTx)
+		err = acnt.SignInvokeTransaction(context.Background(), &test.InvokeTx)
 		require.NoError(t, err)
-		require.Equal(t, txHash.String(), test.ExpectedHash.String())
 
 		resp, err := acnt.AddInvokeTransaction(context.Background(), test.InvokeTx)
-		require.Equal(t, err.Error(), test.ExpectedError.Error())
-		require.Nil(t, resp)
+		fmt.Println(resp, err)
```
cicr99: Can this be removed?

account/account.go
```diff
@@ -420,26 +415,68 @@ func (account *Account) TransactionByHash(ctx context.Context, hash *felt.Felt)
...
-	return callData
+	execCallData = append(execCallData, new(felt.Felt).SetUint64(uint64(len(concatCallData))))
+	execCallData = append(execCallData, concatCallData...)
+
+	return execCallData
+}
+
+func PrintTxAsJson(tx any) {
+	qwe, err := json.MarshalIndent(tx, "", "")
+	fmt.Println(string(qwe), err)
```
cicr99: Why is this needed? It's not being used anywhere else

cicr99 (approved):

account/account.go
rianhughes: It's a helper method that allows one to print the transaction in a marshalled format, which is great for debugging. I use this logic all the time (but have to re-add it every time), so figured it would be good to add it to the pkg

account/account.go
rianhughes: Probably yes

account/account.go
rianhughes: We need to get a reference for formatCallData. The current implementation was what I could figure out from looking at the SDKs, and submitting transactions to the network using my wallet etc. I added multiple invoke test cases to make sure this implementation wasn't specific to a single function, but we really need a reference to be sure we have the correct implementation.

PR #402, opened 2023-10-09
Fix GH workflow devnet error

### LINKED ISSUE #360: Account - fix failing devnet workflow

rianhughes:
closes https://github.com/NethermindEth/starknet.go/issues/360

### DIFF
--- a/.bin/devnet.sh
+++ b/.bin/devnet.sh
@@ -2,7 +2,8 @@
 
 i=0
 version=""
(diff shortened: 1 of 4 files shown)


rianhughes:
https://github.com/NethermindEth/starknet.go/pull/413

PR #415, opened 2023-10-10
Code comments (issue #303).

aquental:
Code comments for all functions in Starknet.go library.
Issue [#303](https://github.com/NethermindEth/starknet.go/issues/303)

PR #416, opened 2023-10-11
Implement rpc05 specVersion

### LINKED ISSUE #405: Addition of `starknet_specVersion` method
Method [here](https://github.com/starkware-libs/starknet-specs/blob/c0980ed9aa5a8eb2def8d9b145fba27b4e29b7f9/api/starknet_api_openrpc.json#L11)

Add tests

rianhughes:
closes https://github.com/NethermindEth/starknet.go/issues/405

### DIFF
--- a/rpc/version.go
+++ b/rpc/version.go
@@ -2,6 +2,9 @@ package rpc
 
 import "context"
 
+// SpecVersion returns the version of the Starknet JSON-RPC specification being used
+// Parameters: None
+// Returns: String of the Starknet JSON-RPC specification
 func (provider *Provider) SpecVersion(ctx context.Context) (string, error) {
 	var result string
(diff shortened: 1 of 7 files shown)


rpc/version.go
```diff
@@ -0,0 +1,9 @@
+package rpc
+
+import "context"
+
+func (provider *Provider) SpecVersion(ctx context.Context) (string, error) {
```
cicr99, line 8: Shouldn't this be added to `RpcProvider` interface as well?

cicr99: Could it also be added in the readme as a supported method?

cicr99: It would also be nice if the comments for this method were added (using the same format as in #415) so we don't have to it later

cicr99 (approved):

PR #422, opened 2023-10-11
rpcv05 update errors

### LINKED ISSUE #408: Update errors
- [x] Addition of `data` to CONTRACT_ERROR
- [x] Update `traceBlockTransaction` error (this is being solved as part of https://github.com/NethermindEth/starknet.go/issues/427 issue)

rianhughes:
closes https://github.com/NethermindEth/starknet.go/issues/408

### DIFF
--- a/rpc/errors.go
+++ b/rpc/errors.go
@@ -108,7 +108,7 @@ var (
 	}
 	ErrBlockNotFound = &RPCError{
 		code:    24,
-		message: "Block not found",
+		message: "Invalid block hash",
 	}
 	ErrInvalidTxnHash = &RPCError{
 		code:    25,


cicr99 (commented):
The tasks on #408 are yet to be completed

rianhughes:
One task on #408 has to do with updating the errors, and the other has to do with updating a method. I might split updating the method into a separate pr if that's okay with you
https://github.com/NethermindEth/starknet.go/issues/427

cicr99:
> One task on #408 has to do with updating the errors, and the other has to do with updating a method. I might split updating the method into a separate pr if that's okay with you #427

Ok, I agree that you can handle the update of the method separately. I updated the issue to track that task with #427  issue. On the other hand, there is the data field in CONTRACT_ERROR task, I left a comment there about it. In any case, I'll update your description to only reference issue #408, because the way it is will close the issue as soon as this get merged and I would like all the tasks to be finished before closing it.

cicr99 (approved):

rpc/errors.go
```diff
@@ -108,7 +108,7 @@ var (
 	}
 	ErrBlockNotFound = &RPCError{
 		code:    24,
-		message: "Block not found",
+		message: "Invalid block hash",
```
cicr99, line 111: I don't think this message needed to be changed  https://github.com/starkware-libs/starknet-specs/blob/c0980ed9aa5a8eb2def8d9b145fba27b4e29b7f9/api/starknet_api_openrpc.json#L3345

cicr99 (changes requested):

rpc/errors.go
cicr99: I looked a bit more into it. It appears there is an inconsistency on the spec. BLOCK_NOT_FOUND error is defined twice with different error messages: [here](https://github.com/starkware-libs/starknet-specs/blob/c0980ed9aa5a8eb2def8d9b145fba27b4e29b7f9/api/starknet_api_openrpc.json#L3345) and [here](https://github.com/starkware-libs/starknet-specs/blob/c0980ed9aa5a8eb2def8d9b145fba27b4e29b7f9/api/starknet_trace_api_openrpc.json#L483C14-L483C29). "Block not found" seems like the correct message.

PR #437, opened 2023-10-13
rpcv05 implement broadcast txn types

### LINKED ISSUE #403: Reintroduction of Broadcasted Transactions
See [BROADCASTED_TXN](https://github.com/starkware-libs/starknet-specs/blob/c0980ed9aa5a8eb2def8d9b145fba27b4e29b7f9/api/starknet_api_openrpc.json#L1825) type on the spec

- Implement specific types:
    - BROADCASTED_INVOKE_TXN
    - BROADCASTED_DECLARE_TXN
    - BROADCASTED_DEPLOY_ACCOUNT_TXN
- Update input parameter in `starknet_estimateFee` method (tracked by #451)

rianhughes:
closes https://github.com/NethermindEth/starknet.go/issues/403


joshklop (approved):
I carefully reviewed the types and ensured they matched the spec. Unless I missed something, they should match exactly.

The only read method that accepts the broadcasted txn types appears to be `starknet_estimateFee`. starknet.go's `EstimateFeeInput` is an `interface{}` that will accept any type, so we can't enforce that users pass the correct type. If that is ok, I think we can merge this PR.

cicr99:
> I carefully reviewed the types and ensured they matched the spec. Unless I missed something, they should match exactly.
> 
> The only read method that accepts the broadcasted txn types appears to be `starknet_estimateFee`. starknet.go's `EstimateFeeInput` is an `interface{}` that will accept any type, so we can't enforce that users pass the correct type. If that is ok, I think we can merge this PR.

You are right that `starknet_estimateFee` method should be updated in order to use BroadcastedTxn. I will do this in a separate PR, as this one is just for implementing the types

rpc/types_broadcast_transaction.go
```diff
@@ -0,0 +1,47 @@
...
+type BroadcastInvokev1Txn struct {
+	InvokeTxnV1
+}
+
+type BroadcastDeclareV1Txn struct {
+	Type TransactionType `json:"type"`
+	// SenderAddress the address of the account contract sending the declaration transaction
+	SenderAddress           *felt.Felt         `json:"sender_address"`
+	MaxFee                  *felt.Felt         `json:"max_fee"`
+	Version                 TransactionVersion `json:"version"`
```
cicr99: This property should be of type `NumAsHex` instead of `TransactionVersion`, which is a string (https://github.com/starkware-libs/starknet-specs/blob/c0980ed9aa5a8eb2def8d9b145fba27b4e29b7f9/api/starknet_api_openrpc.json#L1892)

rpc/types_broadcast_transaction.go
```diff
@@ -0,0 +1,47 @@
...
+	Signature               []*felt.Felt       `json:"signature"`
+	Nonce                   *felt.Felt         `json:"nonce"`
+	DeprecatedContractClass `json:"contract_class"`
```
cicr99: This property should have a name

rpc/types_broadcast_transaction.go
```diff
@@ -0,0 +1,47 @@
...
+}
+type BroadcastDeclareV2Txn struct {
+	Type TransactionType `json:"type"`
+	// SenderAddress the address of the account contract sending the declaration transaction
+	SenderAddress     *felt.Felt         `json:"sender_address"`
+	CompiledClassHash *felt.Felt         `json:"compiled_class_hash"`
+	MaxFee            *felt.Felt         `json:"max_fee"`
+	Version           TransactionVersion `json:"version"`
```
cicr99: Same comment about the type here, should be `NumAsHex` instead

rpc/types_broadcast_transaction.go
```diff
@@ -0,0 +1,47 @@
...
+	Signature         []*felt.Felt       `json:"signature"`
+	Nonce             *felt.Felt         `json:"nonce"`
+	ContractClass     `json:"contract_class"`
```
cicr99: Same comment about the name

Issue #381, opened 2023-10-03
Make deploy accounts transactions nice


rianhughes:
We should add a method to let users deploy accounts using the default UDC etc. Check other SDKs for inspiration

danielcdz:
@rianhughes @thiagodeev Hello! With some guidance and context, I can help you guys with this one

thiagodeev:
Hi @danielcdz, sorry for the delay.
Thanks again for the help!

If you take a look at the `examples` folder, you'll see the `deployContractUDC` example, showing how to deploy a contract via the UDC contract ([ref ](https://docs.starknet.io/architecture-and-concepts/accounts/universal-deployer/)about the UDC).
You'll notice that the example shows a somewhat hard and manual way of doing this, so the proposal of this issue is to create a function/method to help the user deploy contracts via UDC, abstracting a lot of the hard work (similar to the [BuildAndSendInvokeTxn ](https://github.com/NethermindEth/starknet.go/blob/a7f3d98da64995bd23e1ea32472b38518c1af36e/account/account.go#L123) method, that abstracts the hard work of building Invoke txns). You can take a look at this starknet.py example to learn more: https://starknetpy.readthedocs.io/en/latest/guide/deploying_contracts.html

danielcdz:
@thiagodeev sorry for the delay here, I'll be working on this this week, hope to open a draft PR asap, again, thanks for the opportunity and the clarifications

HACKER097:
Hi! I noticed PR #706 has been dormant for a while. Would it be alright if I picked this up and opened a fresh PR based on the original intent, or continued from there if preferred?

thiagodeev:
Hey @HACKER097, thanks!
You can create a new one

thiagodeev:
Done in https://github.com/NethermindEth/starknet.go/pull/760

PR #445, opened 2023-10-22
Add DeployAccount Method to account/account.go for Issue #381

Gabulhas:
Introduces a new DeployAccount method in account/account.go, aiming to simplify the process of account deployment on StarkNet as outlined in Issue #381. The DeployAccount method encapsulates multiple functions into a single call, making it easier and more intuitive for users to deploy accounts.

TODO:
- Add tests (I don't know if this is needed)
- Let users deploy accounts using the default UDC


cicr99:
Thanks for this contribution! I see it still has a [WIP] at the beginning of the title, does this mean it still not ready for review? Can we change it into a draft if that's the case? Do you want us to take a look now even if it still not ready?

Gabulhas:
Sorry, was out for a conference for the last week

I will work on it more, I'll request for a review after :)

Gabulhas:
it's something really small by the way :+1: 

Can you review it?

account/account.go
```diff
@@ -824,3 +824,49 @@ func FmtCalldataCairo2(fnCalls []rpc.FunctionCall) []*felt.Felt {
 
 	return execCallData
 }
+
+type DeployOptions struct {
+	ClassHash           *felt.Felt
+	MaxFee              *felt.Felt
+	DeploytWaitTime     time.Duration
+	ConstructorCalldata []*felt.Felt
+}
```
rianhughes: Can you add comments (eg Classhash, ConstructorCallData). There's also a typo in DeploytWaitTime.

account/account.go
```diff
@@ -824,3 +824,49 @@ func FmtCalldataCairo2(fnCalls []rpc.FunctionCall) []*felt.Felt {
...
+
+func (account *Account) DeployAccount(options DeployOptions) (*rpc.AddDeployAccountTransactionResponse, error) {
```
rianhughes: Can we rename this to something like "CreateAndExecuteAddDeployAccount"?

account/account.go
```diff
@@ -824,3 +824,49 @@ func FmtCalldataCairo2(fnCalls []rpc.FunctionCall) []*felt.Felt {
...
+	if err != nil {
+		return nil, err
+	}
+
+	err = account.SignDeployAccountTransaction(context.Background(), &tx, precomputedAddress)
+	if err != nil {
+		return nil, err
+	}
+
+	resp, err := account.AddDeployAccountTransaction(context.Background(), tx)
```
rianhughes: You will need to wrap the tx in an rpc.BroadcastDeployAccountTxn type

account/account.go
```diff
@@ -824,3 +824,49 @@ func FmtCalldataCairo2(fnCalls []rpc.FunctionCall) []*felt.Felt {
...
+	if err != nil {
+		return nil, err
+	}
+
+	account.WaitForTransactionReceipt(context.Background(), resp.TransactionHash, options.DeploytWaitTime)
```
rianhughes: It's good practice to handle the return types, particularly the errors

account/account_test.go
```diff
@@ -955,6 +955,50 @@ func TestAddDeclareTxn(t *testing.T) {
 	}
 }
 
+func TestDeployAccountDevnet(t *testing.T) {
```
rianhughes: This doesn't run for me (eg see wrapping tx type in rpc.BroadcastDeployAccountTxn comment above)

examples/deployAccount/main.go
```diff
@@ -1,9 +1,11 @@
 package main
```
rianhughes: Can you create a new example, I think it would be nice to have examples using both methods

account/account_test.go
```diff
@@ -955,6 +955,50 @@ func TestAddDeclareTxn(t *testing.T) {
...
+
+	precomputedAddress, err := acnt.PrecomputeAddress(&felt.Zero, fakeUserPub, classHash, []*felt.Felt{fakeUserPub})
+
+	_, err = devnet.Mint(precomputedAddress, new(big.Int).SetUint64(10000000000000000000))
+
+	require.NoError(t, err)
+
+	deployOptions := account.DeployOptions{
+		ClassHash:           classHash,
+		MaxFee:              new(felt.Felt).SetUint64(0),
```
rianhughes: I think you might have to increase the MaxFee?

account/account_test.go
```diff
@@ -955,6 +955,50 @@ func TestAddDeclareTxn(t *testing.T) {
...
+		DeploytWaitTime:     2 * time.Second,
+		ConstructorCalldata: []*felt.Felt{},
```
rianhughes: Also, shouldn't this include "pub"?

examples/deployAccount/main.go
rianhughes: Also, this isnt running for me

examples/deployAccount/main.go
```diff
@@ -59,39 +61,30 @@ func main() {
...
 	if err != nil {
-		panic(err)
+		fmt.Printf("Error: %v\n", err)
+		return
 	}
 
-	// Send transaction to the network
-	resp, err := acnt.AddDeployAccountTransaction(context.Background(), rpc.BroadcastDeployAccountTxn{DeployAccountTxn: tx})
+	fmt.Println("Waiting for deployment")
+	deployOptions := account.DeployOptions{
```
rianhughes: I think you need to set ConstructorCallData

rianhughes (changes requested):
I've left some comments.

account/account.go
Gabulhas: Sure! 
The idea of just having "DeployAccount" was to be easier for the developer, since this is just a merge between two functions.

But will do that!

account/account_test.go
Gabulhas: I need to merge the new RPCv0.5 with my branch, this applies to other parts of my code

account/account.go
Gabulhas: Merged the upstream main branch with this PR.

examples/deployAccount/main.go
Gabulhas: Sure!

Gabulhas:
Just made the changes you asked, let me know if there's something more :smile:

account/account.go
```diff
@@ -824,3 +824,55 @@ func FmtCalldataCairo2(fnCalls []rpc.FunctionCall) []*felt.Felt {
...
 }
+
+type DeployOptions struct {
+	ClassHash           *felt.Felt    // ClassHash is the unique identifier of the account class in the network.
+	MaxFee              *felt.Felt    // MaxFee represents the maximum fee the user is willing to pay for the deployment transaction.
+	DeployWaitTime      time.Duration // DeployWaitTime specifies the duration to wait for the deployment transaction to be processed.
+	ConstructorCalldata []*felt.Felt  // ConstructorCalldata contains the calldata to be passed to the constructor of the account contract upon deployment.
+}
+
+func (account *Account) CreateAndExecuteAddDeployAccount(options DeployOptions) (*rpc.AddDeployAccountTransactionResponse, error) {
```
rianhughes, line 835: This works for rpcv05 and lower. However, rpcv06 introduces a new deployAccount transaction version (v3), meaning users won't be able to use this method to submit v3 transactions. It would be great if we could make this compatible with both transaction versions (eg by accepting the transaction as an argument, tx rpc.DeployAccountType)

Gabulhas: sure will fix it

Gabulhas: compatible with both rpcv06 and previous, or both as in v1 and v2 version?

Gabulhas: by "this method" you mean the whole method "CreateAndExecuteAddDeployAccount" or the way you deploy accounts like in the function?

rianhughes: The CreateAndExecuteAddDeployAccount method.

PR #455, opened 2023-10-27
Handle errors with data

### LINKED ISSUE #408: Update errors

cicr99:
Closes #408


rpc/errors.go
```diff
@@ -49,12 +49,15 @@ func Err(code int, data any) *RPCError {
...
 func tryUnwrapToRPCErr(err error, rpcErrors ...*RPCError) error {
-
 	var nodeErr *RPCError
 	if json.Unmarshal([]byte(err.Error()), nodeErr) != nil {
 		return err
 	}
 
+	if dataErr, ok := isErrorWithData(nodeErr); ok {
+		return dataErr
+	}
```
cicr99: The way this is handled now, if the error `err` is one of the listed errors with data in `isErrorWithData` method, it will be caught here even if it is not contained in `rpcErrors`. This is why I removed these errors from the call to `tryUnwrapToRPCErr` method throughout the code. I'm wondering if I should add them back for readability.
Also should I check the error with data in `isErrorWithData` method is contained in `rpcErrors` list? I didn't think it would cause any problems so I didn't add this verification, but wanted to double check

joshklop: > I'm wondering if I should add them back for readability.

I'm still learning the code base and could be wrong, but I think it is ok as it is in this PR.

> should I check the error with data in isErrorWithData method is contained in rpcErrors list?

I agree with what you did: I don't think the extra verification is necessary. To be honest, it might be simplest to define an rpcErrors slice and check them all every time, instead of explicitly passing a subset of errors as arguments. But that can be a separate PR.

rpc/errors.go
```diff
@@ -63,46 +66,28 @@ func tryUnwrapToRPCErr(err error, rpcErrors ...*RPCError) error {
...
-// - bool: a boolean value indicating if the error is an unexpected error
-func isErrUnexpectedError(err error) (*RPCError, bool) {
-	var nodeErr *RPCError
-	if json.Unmarshal([]byte(err.Error()), nodeErr) != nil {
-		return nil, false
-	}
-
+// - *RPCError: a pointer to the RPCError resulting object
+// - bool: a boolean value indicating if the error is of the same error type as `rpcErr`
+func isErrorWithData(nodeErr *RPCError) (*RPCError, bool) {
```
joshklop: nit: I don't think we need to return the `bool`. We can just return `*RPCError` and check if it is `nil`.

joshklop (approved):

Issue #317, opened 2023-09-11
Use Juno utils.Map


rianhughes:
Look into using juno/pkg/slices.go utils.Map in particular for things like HexArrToFelt(...)  etc

Sah314:
Hello! can you please elaborate the issue a little bit. Thanks.

rianhughes:
Sure, in starknet.go's util package we have helper functions like HexArrToFelt which apply a function over each element in an array. However, we could just use something like Junos Map function https://github.com/NethermindEth/juno/blob/main/utils/slices.go#L14.

Bitcoinnoobie:
I would like to work on this issue.

PR #467, opened 2023-11-15
Replaced some for loops with some Map from juno/utils

Gabulhas:
Pull request for #317
Replaced some of the for loops with the Map function.


joshklop (commented):
Thanks for the contribution! Because importing `github.com/NethermindEth/juno/utils` adds a lot of transitive dependencies that starknet.go doesn't otherwise use, I don't think we'll merge this PR. 

Though there is an issue on the Juno repo to split up its `utils` package to make this less of a problem. https://github.com/NethermindEth/juno/issues/1198

PR #473, opened 2023-11-22
rpcv06 update emitted event

### LINKED ISSUE #472: rpcv06 Some Emitted Event files are not required
EMITTED_EVENT

rianhughes:
closes https://github.com/NethermindEth/starknet.go/issues/472


joshklop (approved):
Why do they need to be omitted sometimes?

rianhughes:
> Why do they need to be omitted sometimes?
If it's not classified as "required" then the "omitempty" flag is added
https://github.com/starkware-libs/starknet-specs/blob/v0.6.0-rc1/api/starknet_api_openrpc.json#L924

PR #490, opened 2023-12-03
Fix function call calldata length for Cairo 0

amayvs:
Calldata length for each function call was set
incorrectly in `FmtCalldataCairo0`, resulting in error

```
ASSERT_EQ instruction failed: 0:17 != 0:18
```


amayvs:
@rianhughes What additional details would be required to get this reviewed?

PR #506, opened 2024-01-20
fix account test json

Lawliet-Chan:
seems you made a mistake, you should unmarshal `casm.json` file instead of `sierra.json`


thiagodeev:
Hi @Lawliet-Chan! Sorry for the delay.
Thanks. I'm closing this since the code was already updated.

PR #517, opened 2024-02-08
Rianhughes/07 get block with receipts

### LINKED ISSUE #516: RPC07 implement new method

rianhughes:
closes https://github.com/NethermindEth/starknet.go/issues/516

Edit: This PR incorporates the changes from rc-1 which relate to block receipts


rpc/types_block.go
```diff
@@ -123,6 +123,28 @@ type PendingBlock struct {
...
+	BlockBodyWithReceipts
+}
+
+type BlockBodyWithReceipts struct {
+	Transactions []TransactionWithReceipt `json:"transactions"`
+}
+
+type TransactionWithReceipt struct {
+	Transaction Transaction        `json:"transaction,omitempty"`
+	Receipt     TransactionReceipt `json:"receipt"`
```
omerfirmak: This field's schema is not yet finalized, let's wait that to be finalized.

rianhughes: Receipts were updated in rc1. I will incorporate the rc-1 changes in this PR

rianhughes: Receipts seem to be finalised in rc-1. I'll incorporate them into this PR.

rianhughes:
Todo: update tests when juno supports rpc v07

IronGauntlets:
The PR, in general, looks good, however, it still needs to be updated as per your comment. I can have a look again after you have updated it.

rpc/block.go
```diff
@@ -166,3 +167,25 @@ func (provider *Provider) BlockWithTxs(ctx context.Context, blockID BlockID) (in
...
+	if err := do(ctx, provider.c, "starknet_getBlockWithReceipts", &result, blockID); err != nil {
+		return nil, tryUnwrapToRPCErr(err, ErrBlockNotFound)
+	}
+	var block BlockWithReceipts
+	err := json.Unmarshal(result, &block)
+	if err == nil && block.BlockStatus != "" {
+		return &block, nil
+	}
+
+	// if result.Status == nil it's a pending block
```
IronGauntlets: This is incorrect, please see [here](https://github.com/starkware-libs/starknet-specs/blob/7ae9e7bdcd2126bd30830b96b10932cf59d13ae3/api/starknet_api_openrpc.json#L3201). If the block is pending then the status should be `"PENDING"`.

rpc/types_block_test.go
```diff
@@ -125,3 +127,48 @@ func TestBlock_Unmarshal(t *testing.T) {
...
+			blockCasted := block.(*BlockWithReceipts)
+			txsWithReceipts := blockCasted.BlockBodyWithReceipts.Transactions
+			require.Equal(t, blockCasted.BlockStatus, test.ExpectedBlock.Status)
+			require.Equal(t, blockCasted.BlockNumber, test.ExpectedBlock.BlockNumber)
+			require.NotZero(t, len(blockCasted.Transactions))
+			require.NotNil(t, blockCasted.L1GasPrice)
+			require.NotNil(t, blockCasted.L1DataGasPrice)
+			require.NotNil(t, txsWithReceipts)
+			require.NotNil(t, txsWithReceipts[0].Transaction)
+			require.NotNil(t, txsWithReceipts[0].Receipt)
```
IronGauntlets: If `test` and `blockCasted` have the same fields and type then you can do `require.Equal`

rpc/types_transaction_receipt.go
```diff
@@ -475,6 +475,19 @@ func (tr *UnknownTransactionReceipt) UnmarshalJSON(data []byte) error {
...
+		recInner, err := json.Marshal(dec["TransactionReceipt"])
+		if err != nil {
+			return err
+		}
+		if err := json.Unmarshal(recInner, &decInner); err != nil {
+			return err
+		}
+		dec = decInner
+	}
+
```
IronGauntlets: It seems to me like the only difference between this piece of code and the one lines 257-268 in file `rpc/types_transaction.go` is `Transaction` and `TransactionReceipt`. Please consider removing this duplication.

rpc/block.go
rianhughes: getBlockWithReceipts can return a [BLOCK_WITH_RECEIPTS](https://github.com/starkware-libs/starknet-specs/blob/7ae9e7bdcd2126bd30830b96b10932cf59d13ae3/api/starknet_api_openrpc.json#L120)which has a [BLOCK_STATUS](https://github.com/starkware-libs/starknet-specs/blob/7ae9e7bdcd2126bd30830b96b10932cf59d13ae3/api/starknet_api_openrpc.json#L1634) (which may be "PENDING"), however getBlockWithReceipts can also [return](https://github.com/starkware-libs/starknet-specs/blob/7ae9e7bdcd2126bd30830b96b10932cf59d13ae3/api/starknet_api_openrpc.json#L1679) a distinct type [PENDING_BLOCK_WITH_RECEIPTS](https://github.com/starkware-libs/starknet-specs/blob/7ae9e7bdcd2126bd30830b96b10932cf59d13ae3/api/starknet_api_openrpc.json#L1679) which has no BLOCK_STATUS.

rianhughes: The naming could be clearer tbh (on the spec side), but I think the current implementation is correct.

rianhughes: I've updated the logic to make this a bit clearer, but it's essentially the same

rpc/types_block_test.go
rianhughes: Updated

rpc/types_transaction_receipt.go
rianhughes: I've moved this duplicated logic to a helper

rpc/block.go
IronGauntlets: Yeah looks like the spec is ambiguous

IronGauntlets (approved):

PR #518, opened 2024-02-08
update block header

rianhughes:


rpc/types_block.go
```diff
@@ -150,10 +150,21 @@ type BlockHeader struct {
...
 	L1GasPrice ResourcePrice `json:"l1_gas_price"`
+	// The price of l1 data gas in the block
+	L1DataGasPrice ResourcePrice `json:"l1_data_gas_price"`
+	// Specifies whether the data of this block is published via blob data or calldata
+	L1DAMode L1DAMode `json:"l1_da_mode"`
 	// Semver of the current Starknet protocol
 	StarknetVersion string `json:"starknet_version"`
 }
 
+type L1DAMode string
```
IronGauntlets: If there are a limited number of L1DAMode this should be a `const` with a String method. By keeping it as a string you don't have any type safety.

IronGauntlets (changes requested):

rpc/types_block.go
rianhughes: I feel like this is worse UX for users though. Eg if they print out a blockheader they will see eg 0 or 1 instead of "BLOB" or "CALLDATA"

IronGauntlets (approved):

PR #525, opened 2024-02-12
implement an example to deploy contracts using the UDC

### LINKED ISSUE #513: UDC support for next release
Pick up the UDC [issue](https://github.com/NethermindEth/starknet.go/issues/192) for the next release.

rianhughes:
closes https://github.com/NethermindEth/starknet.go/issues/513


examples/deployAccountUDC/go.work
kirugan, line 1: Why do we need this file?

examples/deployAccountUDC/main.go
kirugan, line 1: In future we should consider https://go.dev/blog/examples this way of creating examples.

examples/deployAccountUDC/main.go
```diff
@@ -0,0 +1,153 @@
...
+	"github.com/joho/godotenv"
+)
+
+// NOTE : Please add in your keys only for testing purposes, in case of a leak you would potentially lose your funds.
+var (
+	name                  string = "testnet"                                                            // env."name"
+	account_addr          string = "0x06f36e8a0fc06518125bbb1c63553e8a7d8597d437f9d56d891b8c7d3c977716" // Replace it with your account address
+	account_cairo_version        = 0                                                                    // Replace  with the cairo version of your account
+	privateKey            string = "0x0687bf84896ee63f52d69e6de1b41492abeadc0dc3cb7bd351d0a52116915937" // Replace it with your account private key
+	public_key            string = "0x58b0824ee8480133cad03533c8930eda6888b3c5170db2f6e4f51b519141963"  // Replace it with your account public key
```
kirugan, line 24: suggestion: 
1. move all variables that need to be replaced in .env.template file (all values are set with comment to replace it user .env file)
2. add .gitignore in this directory with this content:
```
.env.*
!.env.template
```

this way user don't have to modify anything in code and is safe for comitting private data to some repo

rianhughes: Mind it I push this to a separate issue that we can include in the next release (all the examples will need updated)?

examples/deployAccountUDC/go.work
rianhughes: It allows us to create and run examples using the code from the main directory, without having to create a release first.

examples/deployAccountUDC/main.go
kirugan: Sure

kirugan (approved):

PR #531, opened 2024-02-29
update RPC methods to return RPCErrors

### LINKED ISSUE #530: 0.7.1 RPC methods should return RPC errors
The errors data field currently isn't returned.

rianhughes:
closes https://github.com/NethermindEth/starknet.go/issues/530


rpc/errors.go
```diff
@@ -46,205 +45,144 @@ func Err(code int, data any) *RPCError {
 // - rpcErrors: variadic list of *RPCError objects to be checked
 // Returns:
 // - error: the original error
-func tryUnwrapToRPCErr(err error, rpcErrors ...*RPCError) error {
-	var nodeErr *RPCError
-	if json.Unmarshal([]byte(err.Error()), nodeErr) != nil {
-		return err
+func tryUnwrapToRPCErr(err error, rpcErrors ...*RPCError) *RPCError {
+	nodeErr, ok := err.(*RPCError)
```
archseer: Will this work with `ethrpc.jsonError`? Because of struct layout compatibility? In our code I've had to cast to `DataError`: `err.(ethrpc.DataError)`

rianhughes: Nice catch. I've updated tryUnwrapToRPCErr so that it should work with any json error similar to RPCError (like [ethrpc.jsonErr](https://github.com/ethereum/go-ethereum/blob/master/rpc/json.go#L139))

rpc/errors.go
```diff
@@ -46,205 +46,151 @@ func Err(code int, data any) *RPCError {
 // - rpcErrors: variadic list of *RPCError objects to be checked
 // Returns:
 // - error: the original error
-func tryUnwrapToRPCErr(err error, rpcErrors ...*RPCError) error {
-	var nodeErr *RPCError
-	if json.Unmarshal([]byte(err.Error()), nodeErr) != nil {
-		return err
+func tryUnwrapToRPCErr(err error, rpcErrors ...*RPCError) *RPCError {
+	errBytes, errIn := json.Marshal(err)
```
IronGauntlets, line 50: `json.Marshal()` is not equivalent to `[]byte(err.Error())`, was this intentional?

archseer: Yeah it's intentional (but it could use a comment). The old code was incorrect because CallContext would return a `ethrpc.DataError`, and `Error()` would just return the main part of the message (e.g. "Contract error" but no context). This code seems to try and remap from `DataError` to a RPCError. We could do this a bit more explicitly by casting to `DataError` and remapping the types

rpc/errors.go
archseer: I would use the `Error` and `DataError` interfaces since `jsonErr` and other errors follow this interface: https://github.com/ethereum/go-ethereum/blob/0a2f33946b95989e8ce36e72a88138adceab6a23/rpc/errors.go#L36-L46

rpc/errors.go
rianhughes: The reason I took the marshalling approach was so that it would work with any type with a {Code, Message, Data} structure (DataError, RPCError, any other error that has a similar structure), whereas explicit casting seemed a bit more constrained.

archseer (approved):

PR #534, opened 2024-03-05
RPCv07 rc2 getTransactionReceipt

rianhughes:
Pending receipts removed
blockhash and blockNumber removed from receipts
getTransactionReceipt returns a concrete new type, TXN_RECEIPT_WITH_BLOCK_INFO
Unneeded tests removed (duplicates existing test)

### DIFF
--- a/rpc/transaction_test.go
+++ b/rpc/transaction_test.go
@@ -2,12 +2,10 @@ package rpc
 
 import (
 	"context"
-	"regexp"
 	"testing"
 
 	"github.com/NethermindEth/juno/core/felt"
 	"github.com/NethermindEth/starknet.go/utils"
-	"github.com/google/go-cmp/cmp"
 	"github.com/test-go/testify/require"
 )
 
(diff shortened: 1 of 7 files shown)


rpc/transaction_test.go
```diff
@@ -261,154 +248,33 @@ func TestTransactionReceipt_MatchesCapturedTransaction(t *testing.T) {
...
-		if txReceiptInterface == nil {
-			t.Fatal("transaction receipt should exist")
-		}
-		txnDeclareReceipt, ok := txReceiptInterface.(DeclareTransactionReceipt)
-		if !ok {
-			t.Fatalf("transaction receipt should be Deploy or Declare, instead %T", txReceiptInterface)
-		}
-		if !cmp.Equal(test.ExpectedTxnReceipt, txnDeclareReceipt) {
-			t.Fatalf("the expected transaction blocks to match, instead: %s", cmp.Diff(test.ExpectedTxnReceipt, txnDeclareReceipt))
-		}
```
IronGauntlets, line 411: I am not sure why you are deleting these

rpc/types_transaction_receipt.go
```diff
@@ -271,108 +269,6 @@ func (tr L1HandlerTransactionReceipt) GetExecutionStatus() TxnExecutionStatus {
 	return tr.ExecutionStatus
 }
 
-type PendingL1HandlerTransactionReceipt struct {
```
IronGauntlets, line 274: The pending transaction receipt is still valid as you can provide a tx hash which is pending.

rianhughes: Sure, but pending receipts are no longer explicit types in rc-2.

[starknet_getTransactionReceipt](https://github.com/starkware-libs/starknet-specs/blob/v0.7.0-rc2/api/starknet_api_openrpc.json#L367) explicitly returns [TXN_RECEIPT_WITH_BLOCK_INFO](https://github.com/starkware-libs/starknet-specs/blob/v0.7.0-rc2/api/starknet_api_openrpc.json#L3074) which returns [TXN_RECEIPT](https://github.com/starkware-libs/starknet-specs/blob/v0.7.0-rc2/api/starknet_api_openrpc.json#L3049), and not a [PENDING_TXN_RECEIPT](https://github.com/starkware-libs/starknet-specs/blob/v0.7.0-rc1/api/starknet_api_openrpc.json#L391) like it used to

rpc/transaction_test.go
rianhughes: I figured having (and maintaining) a test for each Receipt type might be overkill. I can add them back in (and update them) if you think it's necessary

rianhughes: We do have a test for TransactionReceipt in TestTransactionReceipt

IronGauntlets: Makes sense

rpc/types_transaction_receipt.go
IronGauntlets: I see

IronGauntlets (approved):

PR #542, opened 2024-03-15
RPC methods to return error

rianhughes:


rianhughes:
Note: errors_test.go shows that rpc methods return an error type (so err!=nil executes as expected), but retains the rich RPC error information.

account/account.go
```diff
@@ -525,19 +525,19 @@ func (account *Account) PrecomputeAddress(deployerAddress *felt.Felt, salt *felt
...
 		select {
 		case <-ctx.Done():
 			return nil, rpc.Err(rpc.InternalError, ctx.Err())
 		case <-t.C:
-			receiptWithBlockInfo, rpcErr := account.TransactionReceipt(ctx, transactionHash)
-			if rpcErr != nil {
-				if rpcErr.Message == rpc.ErrHashNotFound.Message {
+			receiptWithBlockInfo, err := account.TransactionReceipt(ctx, transactionHash)
+			if err != nil {
+				if err.Error() == rpc.ErrHashNotFound.Message {
```
kirugan: can we compare it with `err == rpc.ErrHashNotFound` or  use `errors.Is` ?

examples/deployAccount/main.go
```diff
@@ -93,9 +93,9 @@ func main() {
 	}
 
 	// Send transaction to the network
-	resp, rpcErr := acnt.AddDeployAccountTransaction(context.Background(), rpc.BroadcastDeployAccountTxn{DeployAccountTxn: tx})
-	if rpcErr != nil {
-		panic(fmt.Sprint("Error returned from AddDeployAccountTransaction: ", rpcErr))
+	resp, err := acnt.AddDeployAccountTransaction(context.Background(), rpc.BroadcastDeployAccountTxn{DeployAccountTxn: tx})
+	if err != nil {
+		panic(fmt.Sprint("Error returned from AddDeployAccountTransaction: ", err))
```
kirugan: just curious why `fmt.Sprint` not `fmt.Sprintf`?

rpc/block_test.go
```diff
@@ -134,7 +134,7 @@ func TestBlockWithTxHashes(t *testing.T) {
 
 	type testSetType struct {
 		BlockID                          BlockID
-		ExpectedError                    *RPCError
+		ExpectedError                    error
```
kirugan: nit: rename it to `ExpectedErr` since in other tests type that looks like this one has this naming

rpc/contract.go
```diff
@@ -55,7 +55,7 @@ func (provider *Provider) ClassAt(ctx context.Context, blockID BlockID, contract
 // Returns:
 // - ClassOutput: a ClassOutput interface
 // - error: an error if any
-func typecastClassOutput(rawClass *map[string]any) (ClassOutput, *RPCError) {
+func typecastClassOutput(rawClass *map[string]any) (ClassOutput, error) {
```
kirugan: Just curious why it's ptr to map? (default value of map is nil, could it work without ptr? )

rpc/errors_test.go
```diff
@@ -0,0 +1,24 @@
...
+
+	"github.com/test-go/testify/require"
+)
+
+func TestRPCError(t *testing.T) {
+	if testEnv == "mock" {
+		testConfig := beforeEach(t)
+		_, err := testConfig.provider.ChainID(context.Background())
+		require.NoError(t, err)
+		require.Nil(t, err)
```
kirugan: One of these checks are redundant: if return type is ptr to type you can just leave only `require.Nil` if it's an `error` - leave only `require.NoError(...)`

rpc/mock_test.go
```diff
@@ -499,10 +499,13 @@ func mock_starknet_getEvents(result interface{}, method string, args ...interfac
 	if len(args) != 1 {
 		return errWrongArgs
 	}
-	_, ok = args[0].(EventsInput)
+	ei, ok := args[0].(EventsInput)
 	if !ok {
 		return errWrongArgs
 	}
+	if ei.ChunkSize ==0{
```
kirugan: formatting

rpc/errors_test.go
```diff
@@ -0,0 +1,24 @@
...
+
+import (
+	"context"
+	"testing"
+
+	"github.com/test-go/testify/require"
+)
+
+func TestRPCError(t *testing.T) {
+	if testEnv == "mock" {
```
kirugan, line 11: what is `testEnv` variable? wdyt about chaing it to more go way: using go tags for files

examples/deployAccount/main.go
rianhughes: No reason :) I updated it to use fmt.Sprintf

rpc/errors_test.go
rianhughes: It's set via a flag
	flag.StringVar(&testEnv, "env", "mock", "set the test environment")
    
Happy to open another issue fr a refactor (could bea good first issue?)

rpc/contract.go
rianhughes: Can't remember, maybe I thought passing by reference would be more efficient. Updated it to use a non-ptr

account/account_test.go
```diff
@@ -458,7 +458,7 @@ func TestSignMOCK(t *testing.T) {
 func TestAddInvoke(t *testing.T) {
 
 	type testSetType struct {
-		ExpectedError        *rpc.RPCError
+		ExpectedError        error
```
kirugan: ExpectedErr ?

rpc/errors_test.go
kirugan: > Happy to open another issue fr a refactor (could bea good first issue?)

Yep, created the issue - https://github.com/NethermindEth/starknet.go/issues/543

kirugan (approved):

PR #549, opened 2024-04-01
Update SpecVersion to stop returning InternalError for nil errors

AryanGodara:
add check for nil error in function SpecVersion, instead of returning a wrapped error every time.

### DIFF
--- a/rpc/version.go
+++ b/rpc/version.go
@@ -8,5 +8,8 @@ import "context"
 func (provider *Provider) SpecVersion(ctx context.Context) (string, error) {
 	var result string
 	err := do(ctx, provider.c, "starknet_specVersion", &result)
-	return result, Err(InternalError, err)
+	if err != nil {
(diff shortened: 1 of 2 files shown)


AryanGodara:
@rianhughes This one's ready. I need you approval to run the CI ig

rianhughes:
Would you mind updating the [test](https://github.com/AryanGodara/starknet.go/blob/fix-testSpecVersion/rpc/version_test.go#L23) to use 0.7.0, so that it passes on testnet? Thanks

AryanGodara:
> Would you mind updating the [test](https://github.com/AryanGodara/starknet.go/blob/fix-testSpecVersion/rpc/version_test.go#L23) to use 0.7.0, so that it passes on testnet? Thanks

Done, sorry I missed this 😬

rianhughes (approved):

rianhughes:
Nice!

PR #555, opened 2024-04-11
Update rpc: Call, Contract, Event and mock tests

### LINKED ISSUE #544: goerli depreceation
Any tests and examples that currently use Goerli should use another test network, like Sepolia. 

1. Enter the rpc directory
2. rename .env.template to .env.testnet, and enter the Sepolia endpoint
3. go test ./... -env testnet

You will see a bunch of tests that currently depend on Goerli, but should be updated to depend on Sepolia instead.

Free RPC endpoints https://data.voyager.online/

AryanGodara:
Partially resolves #544


rpc/contract_test.go
```diff
@@ -62,7 +64,7 @@ func TestClassAt(t *testing.T) {
 	for _, test := range testSet {
 		spy := NewSpy(testConfig.provider.c)
 		testConfig.provider.c = spy
-		resp, err := testConfig.provider.ClassAt(context.Background(), WithBlockTag("latest"), test.ContractAddress)
+		resp, err := testConfig.provider.ClassAt(context.Background(), WithBlockHash(utils.TestHexToFelt(t, "0x561eeb100ad42aedc8810cce883caccc77eda75a9af58b24aabb770c027d249")), test.ContractAddress)
```
rianhughes: we shouldn't hardcode the block hash parameter here. Lets push it to the testSet (eg use test.Block)

rpc/contract_test.go
```diff
@@ -203,13 +205,13 @@ func TestClass(t *testing.T) {
 		},
 		"testnet": {
 			{
-				BlockID:         WithBlockTag("pending"),
-				ClassHash:       utils.TestHexToFelt(t, "0x493af3546940eb96471cf95ae3a5aa1286217b07edd1e12d00143010ca904b1"),
+				BlockID:         WithBlockNumber(15329),
+				ClassHash:       utils.TestHexToFelt(t, "0x079b7ec8fdf40a4ff6ed47123049dfe36b5c02db93aa77832682344775ef70c6"),
```
rianhughes: Can we update this test? Currently it seems to imply that we expect v0 Contracts, but the class you are returning in this test is actually a v2. They have different field types (see DeprecatedContractClass and ContractClass), so it would be great if we could handle both of these cases, and assert equality on some of the fields (eg the first few felts in SierraProgram if it's a ContractClass type)

AryanGodara: I added testcases with v2 contract and classes for this function, `TestClassHash` and `TestClassAt`, and the return type seems to be working properly (currently checking if `contract_class_version` field is present to check the type of the returned result (`depcreatedContractClass` or `contractClass`)

`ssert equality on some of the fields (eg the first few felts in SierraProgram if it's a ContractClass type)` Just need to add this check now.

And also `TestClassAt` is pending in implementation

thiagodeev:
Call and Contract tests passing on mock, sepolia, devnet, and mainnet.

rpc/call_test.go
```diff
@@ -82,21 +81,18 @@ func TestCall(t *testing.T) {
...
-			t.Fatal("should return an output")
-		}
-		require.Equal(t, test.ExpectedPatternResult, output[0])
+		require.NoError(err)
+		require.NotEmpty(output, "should return an output")
+
+		diff, err := spy.Compare(output, false)
+		require.NoError(err)
+		require.Equal(diff, "FullMatch", "expecting to match")
+
```
rianhughes: We should be able to remove these lines (no need to use spy)

rpc/contract_test.go
```diff
@@ -131,27 +171,17 @@ func TestClassHashAt(t *testing.T) {
...
-				log.Fatal(err)
-			}
-			t.Fatal("structure expecting to be FullMatch, instead", diff)
-		}
-		if classhash == nil {
-			t.Fatalf("should return a class, instead %v", classhash)
-		}
-		require.Equal(t, test.ExpectedClassHash, classhash)
+		require.NoError(err, "expecting to match")
+		require.Equal(diff, "FullMatch", "structure expecting to be FullMatch")
```
rianhughes: What is this doing? Does it test anything that require.Equal(test.ExpectedClassHash, classhash) doesn't?

rpc/contract_test.go
```diff
@@ -202,47 +232,61 @@ func TestClass(t *testing.T) {
...
+			require.NoError(err, "expecting to match")
+			require.Equal(diff, "FullMatch", "structure expecting to be FullMatch")
 
 			if !strings.HasPrefix(class.Program, test.ExpectedProgram) {
 				t.Fatal("code should exist")
 			}
-		case ContractClass:
-			require.Equal(t, class.EntryPointsByType.Constructor, test.ExpectedEntryPointConstructor)
+		case *ContractClass:
+			diff, err := spy.Compare(class, false)
```
rianhughes: Is thi spy doing anything? can we remove it?

rpc/contract_test.go
```diff
@@ -298,25 +350,16 @@ func TestStorageAt(t *testing.T) {
...
-			if _, err := spy.Compare(value, true); err != nil {
-				log.Fatal(err)
-			}
-			t.Fatal("structure expecting to be FullMatch, instead", diff)
-		}
-		if value != test.ExpectedValue {
-			t.Fatalf("expecting value %s, got %s", test.ExpectedValue, value)
-		}
+		require.NoError(err, "expecting to match")
+		require.Equal(diff, "FullMatch", "structure expecting to be FullMatch")
```
rianhughes: Same with this spy

rpc/contract_test.go
```diff
@@ -337,46 +380,53 @@ func TestNonce(t *testing.T) {
...
-			t.Fatal("expecting to match", err)
-		}
-		if diff != "FullMatch" {
-			if _, err := spy.Compare(nonce, true); err != nil {
-				log.Fatal(err)
-			}
-			t.Fatal("structure expecting to be FullMatch, instead", diff)
-		}
+		require.NoError(err, "expecting to match")
+		require.Equal(diff, "FullMatch", "structure expecting to be FullMatch")
```
rianhughes: and this diff :)

rpc/call_test.go
thiagodeev: Done

rpc/contract_test.go
thiagodeev: I have the same thought. Removed in the new commit

rpc/contract_test.go
thiagodeev: Agree. Done

rpc/contract_test.go
thiagodeev: Done

rpc/contract_test.go
thiagodeev: Done

rianhughes (approved):

Issue #551, opened 2024-04-10
rpc testnet tests fix


rianhughes:
There are a couple of failing testnet tests in the rpc package which need fixed

PR #563, opened 2024-05-03
provider_test.go upgraded to Sepolia

thiagodeev:
Started upgrading the RPC tests to Sepolia network instead of Goerli.
This PR upgrades the provider_test.go tests to Sepolia.
Partially resolves the issue https://github.com/NethermindEth/starknet.go/issues/551.

### DIFF
--- a/devnet/devnet.go
+++ b/devnet/devnet.go
@@ -60,7 +60,9 @@ func (devnet *DevNet) api(uri string) string {
 // the list of accounts and any error that occurred during the process.
 //
 // Parameters:
-//   none
+//
+//	none
+//
 // Returns:
 // - []TestAccount: a slice of TestAccount structs
 func (devnet *DevNet) Accounts() ([]TestAccount, error) {
@@ -87,9 +89,12 @@ func (devnet *DevNet) Accounts() ([]TestAccount, error) {
 // and false otherwise.
 //
 // Parameters:
-//  none
+//
+//	none
+//
 // Returns:
-//  bool: true if the DevNet is alive, false otherwise
+//
+//	bool: true if the DevNet is alive, false otherwise
 func (devnet *DevNet) IsAlive() bool {
 	req, err := http.NewRequest(http.MethodGet, devnet.api("/is_alive"), nil)
 	if err != nil {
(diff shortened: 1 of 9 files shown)


thiagodeev:
The fixed test is breaking with the 'devnet' param, the devnet is returning 'goerli' as the chainId. I didn't test with devnet, my bad.
I will switch the current legacy devnet container to the new recommended devnet-rs container.

rianhughes:
Hey, the PR looks good, you just need to fix the TestCall that is failing on devnet

devnet/devnet.go
```diff
@@ -107,8 +112,9 @@ func (devnet *DevNet) IsAlive() bool {
 }
 
 type MintResponse struct {
-	NewBalance *big.Int `json:"new_balance"`
-	Unit       string   `json:"unit"`
+	NewBalance      string `json:"new_balance"`
```
thiagodeev, line 115: The new devnet version returns a string instead a number, and a new field. The compiler was unable to convert to bigInt

devnet/devnet_test.go
```diff
@@ -44,26 +48,6 @@ func TestDevnet_Accounts(t *testing.T) {
 	}
 }
 
-// TestDevnet_FeeToken tests the FeeToken function of the Devnet struct.
```
rianhughes, line 47: Why was this removed? I think we can add it back in without breaking tests

thiagodeev: I removed this tests because the[ new devnet doesn't have the "/fee_token" endpoint anymore](https://github.com/0xSpaceShard/starknet-devnet-rs/blob/main/crates/starknet-devnet-server/src/server.rs#L34). That being said, I believe the "Mint" and "TestCall" tests are enough to prove the token existing, address correctness, and functionalities.

rpc/call_test.go
```diff
@@ -38,12 +38,12 @@ func TestCall(t *testing.T) {
 			{
 				FunctionCall: FunctionCall{
 					// ContractAddress of predeployed devnet Feetoken
-					ContractAddress:    utils.TestHexToFelt(t, "0x49d36570d4e46f48e99674bd3fcc84644ddd6b96f7c741b1562b82f9e004dc7"),
-					EntryPointSelector: utils.GetSelectorFromNameFelt("name"),
+					ContractAddress:    utils.TestHexToFelt(t, DevNetETHAddress),
```
thiagodeev, line 41: Using an existing variable instead of hardcoding the address

rpc/call_test.go
```diff
@@ -38,12 +38,12 @@ func TestCall(t *testing.T) {
...
+					EntryPointSelector: utils.GetSelectorFromNameFelt("decimals"),
```
thiagodeev, line 42: I had some issues related to felt.Felt comparison using the "name" method. Using "decimals" was simpler to compare

rianhughes (approved):

PR #566, opened 2024-05-10
RPC/block tests upgraded to sepolia, fixed not running correct RPC url passed from env flag on CLI bug

thiagodeev:
- Upgraded block_test.go tests to Sepolia network.
- Implemented TestBlockWithTxsAndInvokeTXNV1 test.
- Fixed the bug that ignored the correct RPC URL passed by the 'env' flag in the CLI during testing (e.g. "go test -v . -env devnet" did not load the Devnet RPC), instead getting the Sepolia URL.

Partially resolves the issue https://github.com/NethermindEth/starknet.go/issues/551.

### DIFF
--- a/rpc/block_test.go
+++ b/rpc/block_test.go
@@ -645,8 +645,8 @@ func TestBlockTransactionCount(t *testing.T) {
 		},
 		"testnet": {
 			{
-				BlockID:       WithBlockNumber(300000),
-				ExpectedCount: 23,
+				BlockID:       WithBlockNumber(30000),
+				ExpectedCount: 4,
 			},
 		},
 		"mainnet": {},
--- a/rpc/block_test.go
+++ b/rpc/block_test.go
@@ -700,8 +700,6 @@ func TestCaptureUnsupportedBlockTxn(t *testing.T) {
 // Returns:
 //
 //	none
-//
-// TODO: Find a block with such a Txn
 func TestBlockWithTxsAndInvokeTXNV1(t *testing.T) {
 	testConfig := beforeEach(t)
 
@@ -759,17 +757,6 @@ func TestBlockWithTxsAndInvokeTXNV1(t *testing.T) {
 		"mock": {},
 		"testnet": {
 			{
-				BlockID:       WithBlockTag("latest"),
-				ExpectedError: nil,
-			},
-			{
-				BlockID:                     WithBlockHash(utils.TestHexToFelt(t, "0x6df565874b2ea6a02d346a23f9efb0b26abbf5708b51bb12587f88a49052964")),
-				ExpectedError:               nil,
-				want:                        &fullBlockSepolia64159,
-				LookupTxnPositionInExpected: 0,
-				LookupTxnPositionInOriginal: 4,
-			},
-			{
 				BlockID:                     WithBlockNumber(64159),
 				ExpectedError:               nil,
 				want:                        &fullBlockSepolia64159,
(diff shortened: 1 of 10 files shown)


rpc/provider_test.go
```diff
@@ -101,7 +101,6 @@ func beforeEach(t *testing.T) *testConfiguration {
 		return &testConfig
 	}
 
-	testConfig.base = "https://free-rpc.nethermind.io/sepolia-juno"
```
thiagodeev, line 104: This line was overriding the default base URLs coming from testConfig and via the "env" flag in the CLI.
Now run `go test -v . -env ENVIROMENT` really get the corrent URL from the config, no needing to edit the "env.template" file if the URL already exists on the testConfig variable.

rpc/block_test.go
```diff
@@ -735,24 +703,128 @@ func TestCaptureUnsupportedBlockTxn(t *testing.T) {
...
 		"mainnet": {},
 	}[testEnv]
+
 	for _, test := range testSet {
-		if test.check {
-			t.Fatalf("error running test: %v", ErrNotImplemented)
+		spy := NewSpy(testConfig.provider.c)
+		testConfig.provider.c = spy
+		blockWithTxsInterface, err := testConfig.provider.BlockWithTxs(context.Background(), test.BlockID)
+		if err != test.ExpectedError {
```
rianhughes: Can we do something like require.Equal(t, test.ExpectedErr,err)?

rianhughes: We could also so and if-else statement so that we only check the response if there is no error etc

rpc/block_test.go
```diff
@@ -735,24 +703,128 @@ func TestCaptureUnsupportedBlockTxn(t *testing.T) {
...
+			t.Fatal("BlockWithTxHashes match the expected error:", err)
+		}
+		if test.ExpectedError != nil && blockWithTxsInterface == nil {
+			continue
+		}
+		blockWithTxs, ok := blockWithTxsInterface.(*Block)
+		if !ok {
```
rianhughes: require.True(t,ok)

rpc/block_test.go
```diff
@@ -735,24 +703,128 @@ func TestCaptureUnsupportedBlockTxn(t *testing.T) {
...
+			t.Fatalf("expecting *rpv02.Block, instead %T", blockWithTxsInterface)
+		}
+		_, err = spy.Compare(blockWithTxs, false)
+		if err != nil {
+			t.Fatal("expecting to match", err)
```
rianhughes: require.NoError(t,err)

rpc/block_test.go
```diff
@@ -735,24 +703,128 @@ func TestCaptureUnsupportedBlockTxn(t *testing.T) {
...
+		if !strings.HasPrefix(blockWithTxs.BlockHash.String(), "0x") {
+			t.Fatal("Block Hash should start with \"0x\", instead", blockWithTxs.BlockHash)
+		}
+
+		if len(blockWithTxs.Transactions) == 0 {
+			t.Fatal("the number of transaction should not be 0")
 		}
+
+		if test.want != nil {
+			if (*test.want).BlockHash == &felt.Zero {
```
rianhughes: ?

rpc/block_test.go
thiagodeev: Yes, I agree.
This test, inlcuding all code lines commented in this review, was based on the existing [TestBlockWithTxsAndInvokeTXNV0](https://github.com/NethermindEth/starknet.go/blob/a4cfd1c94bf723cfb2a4a72251fcf3ed65ac9756/rpcv02/block_test.go#L253) test. I chose to make as few modifications as possible, as my main goal was just to update the tests to use the Sepolia network. I apologies.
From now on,  I'll make some adjustments when I see something that could be improved!

Implemented in the [new commit](https://github.com/NethermindEth/starknet.go/pull/566/commits/b6b1c26212540990fb75ecfd744a6eb7681b5caa).

rpc/block_test.go
thiagodeev: Implemented in the [new commit](https://github.com/NethermindEth/starknet.go/pull/566/commits/b6b1c26212540990fb75ecfd744a6eb7681b5caa).

rpc/block_test.go
thiagodeev: This code compares the spy object with the block object. The two objects are clearly different (I tested getting the result, it was "NoMach"). I couldn't see any utility for this, so I removed it in the [new commit](https://github.com/NethermindEth/starknet.go/pull/566/commits/b6b1c26212540990fb75ecfd744a6eb7681b5caa).

rpc/block_test.go
thiagodeev: Explained in the first comment of this review, I got this code from the existing TestBlockWithTxsAndInvokeTXNV0 test.

Removed in the [new commit](https://github.com/NethermindEth/starknet.go/pull/566/commits/b6b1c26212540990fb75ecfd744a6eb7681b5caa).

rianhughes (approved):

PR #574, opened 2024-06-12
Move PrecomputeAddress function to 'contract' package

thiagodeev:
The PrecomputeAddress function precomputes the address of a contract instance thus it's not directly related to the 'account' package, but to the 'contract' package. The only use of this function in the 'account' package is to calculate the account contract address for deploying an account, and to satisfy that I created the PrecomputeAccountAddress function in the 'account' package that calls the PrecomputeAddress function passing the proper parameters.

This will prevent developers from generating wrong account addresses when deploying an account as this process requires the 'deployer_address' to be '0', but the previous function was not intuitive when asking for the 'deployerAddress' parameter without describing that it should be '0' in such case. The new account.PrecomputeAccountAddress function passes '0' as 'deployer_address' automatically to the contract.PrecomputeAccount without prompting the user.

https://docs.starknet.io/architecture-and-concepts/smart-contracts/contract-address/


contracts/contracts_test.go
```diff
@@ -50,3 +54,41 @@ func TestUnmarshalCasmClass(t *testing.T) {
...
+	classHash := utils.TestHexToFelt(t, "0xdeadbeef")
+	constructorCalldata := []*felt.Felt{
+		deployerAddress,
+	}
+
+	expectedPrecomputedAddress := []*felt.Felt{
+		utils.TestHexToFelt(t, "0x3791840c6746577f27f4a5766a6862b7486f0a254d34e0d2f0a5d238160cf10"),
+		utils.TestHexToFelt(t, "0x479b5192d4af71279a92eddb09e4dcfdc00751ccb27da0e9c4f9b2312318242"),
+		utils.TestHexToFelt(t, "0x7a824044120ac54527d701e18b2eea6e11fb0a8f070a21acd13a7a4062065b4"),
+	}
```
rianhughes: Where did you get these? It would also be nice to have a test case that generates an address that exists on-chain

thiagodeev: I generated them from starknet.js. Nice idea, I'll do it

contracts/contracts_test.go
```diff
@@ -50,3 +54,65 @@ func TestUnmarshalCasmClass(t *testing.T) {
...
+		{ //https://sepolia.voyager.online/tx/0x7a4458b402a172e730c947b293a499d310a7ae6cfb18b5d9774fc10625927e5
+			DeployerAddress: "0",
+			Salt:            "0x023a851e8aeba201772098e1a1db3448f6238b20f928527242eb383905d91a87",
+			ClassHash:       "0x061dac032f228abef9c6626f995015233097ae253a7f72d68552db02f2971b8f",
+			ConstructorCalldata: []*felt.Felt{
+				utils.TestHexToFelt(t, "0x023a851e8aeba201772098e1a1db3448f6238b20f928527242eb383905d91a87"),
+			},
+			ExpectedPrecomputedAddress: "0x28771beb7a2522a07d2ae6fc1fa5af942e8e863f70e6d7d74f9600ea3d5c242",
+		},
+		{ //https://sepolia.voyager.online/tx/0x2fd8282cbf6cd496b0bd896722c92b5118cae13fdcb96250c9a2e188adb7c86
```
rianhughes: I think this should be 
https://sepolia.voyager.online/tx/0x2419a80d80045dd08cdb2606850c4eaf0ed8e705ee07bb1837d8daf12263bc0

thiagodeev: You're right. Sorry

thiagodeev: Done

rianhughes (approved):

PR #583, opened 2024-06-28
Adding support for Byte array serialization

krishnateja262:
Utility function which takes in a string and converts it to array of felt objects


krishnateja262:
@thiagodeev Please check and let me know if any changes are required

thiagodeev:
Nice code @krishnateja262!
I'll make some comments

utils/Felt.go
```diff
@@ -87,3 +90,54 @@ func FeltArrToBigIntArr(f []*felt.Felt) []*big.Int {
...
+const SHORT_LENGTH = 31
+
+// ByteArrToFelt converts ByteArray to array of Felt objects.
+//
+// Parameters:
+// - s: string/bytearray to convert
+// Returns:
+// - []*felt.Felt: the array of felt.Felt objects
+// - error: an error, if any
+func ByteArrToFelt(s string) ([]*felt.Felt, error) {
```
thiagodeev: Since we are converting a golang string to the equivalent Cairo ByteArray type, let's change this function name to something like "StringToByteArrFelt" and adapt the description.

utils/Felt.go
```diff
@@ -87,3 +90,54 @@ func FeltArrToBigIntArr(f []*felt.Felt) []*big.Int {
...
+	harr, err := HexArrToFelt(hexarr)
+	if err != nil {
+		return nil, err
+	}
+
+	harr = append(harr, new(felt.Felt).SetUint64(size))
+	return append([]*felt.Felt{new(felt.Felt).SetUint64(count)}, harr...), nil
+}
+
+func splitLongString(s string) ([]string, error) {
```
thiagodeev: The Single-Responsibility Principle is a good thing, but since we are creating just a utility function, and as this function is being used only once, let's add the content of the "splitLongString" function to the body of the "ByteArrToFelt" function for the sake of simplicity.

utils/Felt.go
```diff
@@ -87,3 +90,54 @@ func FeltArrToBigIntArr(f []*felt.Felt) []*big.Int {
...
+	exp := fmt.Sprintf(".{1,%d}", SHORT_LENGTH)
+	r, err := regexp.Compile(exp)
```
thiagodeev: Let's use [MustCompile](https://pkg.go.dev/regexp#MustCompile) instead since we already know it's a valid regex expression. It'll s simplify the code

utils/Felt.go
```diff
@@ -87,3 +90,54 @@ func FeltArrToBigIntArr(f []*felt.Felt) []*big.Int {
 	}
 	return bigArr
 }
+
+const SHORT_LENGTH = 31
```
thiagodeev: As we will unify both functions, let's put this const inside the "ByteArrToFelt" function

thiagodeev (changes requested):

utils/Felt.go
krishnateja262: Done

utils/Felt.go
krishnateja262: Done

utils/Felt.go
krishnateja262: Done

utils/Felt.go
krishnateja262: Done

utils/Felt.go
```diff
@@ -87,3 +90,43 @@ func FeltArrToBigIntArr(f []*felt.Felt) []*big.Int {
 	}
 	return bigArr
 }
+
+// StringToByteArrFelt converts string to array of Felt objects.
+//
+// Parameters:
+// - s: string/bytearray to convert
+// Returns:
```
rianhughes, line 106: Can we add a comment about the structure of the returned felts? eg something along the lines of [number of felts, felts..., size of last felt]

utils/Felt.go
```diff
@@ -87,3 +90,43 @@ func FeltArrToBigIntArr(f []*felt.Felt) []*big.Int {
...
+// - []*felt.Felt: the array of felt.Felt objects
+// - error: an error, if any
+func StringToByteArrFelt(s string) ([]*felt.Felt, error) {
```
rianhughes, line 113: What is the motivation for this function? Currently it's not being used so it's unclear why it should be in the codebase

utils/Felt.go
```diff
@@ -87,3 +90,43 @@ func FeltArrToBigIntArr(f []*felt.Felt) []*big.Int {
...
+		size = uint64(len(val))
+		hexarr = append(hexarr, hex.EncodeToString([]byte(val)))
+	}
+
+	harr, err := HexArrToFelt(hexarr)
+	if err != nil {
+		return nil, err
+	}
+
+	harr = append(harr, new(felt.Felt).SetUint64(size))
```
rianhughes, line 144: why are we interested in the size of the last felt?

utils/Felt_test.go
```diff
@@ -0,0 +1,39 @@
...
+)
+
+func TestByteArrToFelt(t *testing.T) {
+	var tests = []struct {
+		in  string
+		out []string
+	}{
+		{
+			in:  "hello",
+			out: []string{"0x0", "0x68656c6c6f", "0x5"},
```
rianhughes, line 16: I would have expected the first string to be 0x1 (similarly for the next test case, I would have expected 0x2)

utils/Felt.go
krishnateja262: This utility method is useful for interacting with contracts that accept `ByteArray` as an input parameter, similar to the `HexToFelt` `HexArrToFelt` helper methods. The `FunctionCall` can only take an array of felts as its payload, and this function assists in the serialization process, eliminating the need to find and go through the documentation.

utils/Felt_test.go
krishnateja262: Based on the example as mentioned in the [doc](https://docs.starknet.io/architecture-and-concepts/smart-contracts/serialization-of-cairo-types/#serialization_of_byte_arrays)

First element in the array indicates number of words with 31 bytes of length

First test case `hello` does not have any subwords with size 31
```
    0, // Number of 31-byte words in the data array.
    0x68656c6c6f, // Pending word
    5 // Length of the pending word, in bytes
 ```
Second test case has 37 characters, with one 31 bytes word and the pending word is of 6 bytes long
```
    1, // Number of 31-byte words in the array construct.
    0x4c6f6e6720737472696e672c206d6f7265207468616e203331206368617261, // 31-byte word.
    0x63746572732e, // Pending word
    6 // Length of the pending word, in bytes
 ```

utils/Felt.go
krishnateja262: Based on the example as mentioned in the [doc](https://docs.starknet.io/architecture-and-concepts/smart-contracts/serialization-of-cairo-types/#serialization_of_byte_arrays)

Last element in the array indicates the size of pending word

First test case `hello` does not have any subwords with size 31. The pending word(hello) is of size 5
```
    0, // Number of 31-byte words in the data array.
    0x68656c6c6f, // Pending word
    5 // Length of the pending word, in bytes
 ```

File: utils/Felt.go (hunk 5 above)
Comment (rianhughes) [reply utils/Felt.go:106]: Would you mind adding this link to the function documentation? Thanks

https://docs.starknet.io/architecture-and-concepts/smart-contracts/serialization-of-cairo-types/

File: utils/Felt.go (hunk 7 above)
Comment (rianhughes) [reply utils/Felt.go:144]: Great, thanks!

**Review by thiagodeev — APPROVED** [review]

File: utils/Felt.go [hunk 9]
```diff
@@ -87,3 +90,58 @@ func FeltArrToBigIntArr(f []*felt.Felt) []*big.Int {
...
+	exp := fmt.Sprintf(".{1,%d}", SHORT_LENGTH)
+	r := regexp.MustCompile(exp)
+
+	arr := r.FindAllString(s, -1)
+	if len(arr) == 0 {
+		return []*felt.Felt{}, fmt.Errorf("invalid string no matches found, s: %s", s)
+	}
+
+	hexarr := []string{}
+	var (
```
thiagodeev: Let's simplify this to `var count, size uint64`

utils/Felt.go
```diff
@@ -87,3 +90,58 @@ func FeltArrToBigIntArr(f []*felt.Felt) []*big.Int {
...
+		}
+		hexarr = append(hexarr, hex.EncodeToString([]byte(val)))
+	}
+
+	harr, err := HexArrToFelt(hexarr)
+	if err != nil {
+		return nil, err
+	}
+
+	if size > 0 {
```
thiagodeev: I believe this condition is not necessary. Since the [doc](https://docs.starknet.io/architecture-and-concepts/smart-contracts/serialization-of-cairo-types/#serialization_of_byte_arrays) says that this is the "Length of the pending word, in bytes", if the length is 0 we should set it to 0, right? And since 0 is the default uint64 value, we can just set it without any condition.
We can see this with the "Number of 31-byte words in the data array." value in the doc, they push a 0 when there is no 31-byte word, so I believe that this should also be the case for pending word length.

thiagodeev (changes requested):

thiagodeev:
Very good @krishnateja262!
Now we just need to make some adjustments to the test function

utils/Felt_test.go
```diff
@@ -0,0 +1,47 @@
+package utils
+
+import (
+	"testing"
+)
+
+func TestByteArrToFelt(t *testing.T) {
```
thiagodeev: Update the test name to the current function name

utils/Felt_test.go
```diff
@@ -0,0 +1,47 @@
...
+		},
+		{
+			in:  "Long string, more than 31 charaLong string, more than 31 chara",
+			out: []string{"0x2", "0x4c6f6e6720737472696e672c206d6f7265207468616e203331206368617261", "0x4c6f6e6720737472696e672c206d6f7265207468616e203331206368617261", "0x0"},
+		},
+	}
+
+	for _, tc := range tests {
+		res, err := StringToByteArrFelt(tc.in)
+		if err != nil {
```
thiagodeev: We can change it to "require.NoError"

utils/Felt_test.go
```diff
@@ -0,0 +1,47 @@
...
+			t.Fatalf("error in byte array conversion, err: %v", err)
+		}
+
+		if len(res) != len(tc.out) {
```
thiagodeev: We can change it to "require.Len" or "require.Equal"

krishnateja262:
> Very good @krishnateja262! Now we just need to make some adjustments to the test function

Hi @thiagodeev the last 2 test cases have exactly 31 and 62 characters respectively with required output felt array having 0x0 as the ending value. These tests are passing. Are there any other changes you are looking for?

utils/Felt_test.go
```diff
@@ -0,0 +1,47 @@
...
+			t.Fatalf("error in byte array conversion, invalid length")
+		}
+
+		out, _ := HexArrToFelt(tc.out)
+		for i, cmp := range out {
```
thiagodeev: We can change it to "require.Equal", "require.EqualValues", "require.Exactly"...

thiagodeev (changes requested):

thiagodeev (approved):

thiagodeev:
One last thing: have you tested this function interacting with a Cairo method that requires a ByteArray as input? Did it work?

thiagodeev (changes requested):
> One last thing: have you tested this function interacting with a Cairo method that requires a ByteArray as input? Did it work?

Tested locally and working normally. Great work @krishnateja262!  
To finish the feature, could you do the reverse? ByteArray to String? So the user can convert the value to a string again

utils/Felt_test.go
```diff
@@ -0,0 +1,40 @@
...
+		{
+			in:  "Long string, more than 31 characters.",
+			out: []string{"0x1", "0x4c6f6e6720737472696e672c206d6f7265207468616e203331206368617261", "0x63746572732e", "0x6"},
+		},
+		{
+			in:  "Long string, more than 31 chara",
+			out: []string{"0x1", "0x4c6f6e6720737472696e672c206d6f7265207468616e203331206368617261", "0x0"},
+		},
+		{
+			in:  "Long string, more than 31 charaLong string, more than 31 chara",
```
rianhughes: typo: chara

utils/Felt_test.go
```diff
@@ -0,0 +1,40 @@
...
+			out: []string{"0x2", "0x4c6f6e6720737472696e672c206d6f7265207468616e203331206368617261", "0x4c6f6e6720737472696e672c206d6f7265207468616e203331206368617261", "0x0"},
+		},
+	}
+
+	for _, tc := range tests {
+		res, err := StringToByteArrFelt(tc.in)
+		require.NoError(t, err, "error returned from StringToByteArrFelt")
+		require.Len(t, res, len(tc.out), "invalid conversion: array sizes do not match")
+
+		out, _ := HexArrToFelt(tc.out)
```
rianhughes: we should handle the error

rianhughes:
Just those two small comments, happy to approve otherwise. Thanks!

krishnateja262:
@rianhughes have done the changes as suggested

@thiagodeev pushed an implementation of Felt Byte array to string conversion, please check and let me know for any changes

krishnateja262:
@thiagodeev deployed a simple bytearray storage contract in sepolia here: 0x0080fc7dbc9adfe5f4edc3af816f722cc102f86786c053a0f76860d80afdd0a7

I made an observation regarding strings with 31 characters or multiples of 31 characters when testing with the contract. The chain expects both pending word and the size of the pending word. In the initial implementation, only the length of the pending word was pushed into the array, causing transactions to revert.

To resolve this, I modified the code to also add `0x0` for the pending word since it is empty. Now, the array typically ends with `[0x0, 0x0]` for strings with 31 or multiples of 31 characters.

thiagodeev (approved):
Nice catch @krishnateja262. Well done.
Let's also wait for @rianhughes review

krishnateja262:
@thiagodeev let me know if we can do this for U256 data type as well(serialisation and deserialisation). I can create a new issue and raise PR

thiagodeev:
Of course @krishnateja262! It will be very helpful

krishnateja262:
@rianhughes Please let me know if there's anything I need to do to help get this merged

PR #585, opened 2024-07-03
chore::> Constrain BroadcastDeclare and BroadcastDeployAccount 

PsychoPunkSage:
Here
I implemented:
1. `GetContractClass` ::> `BroadcastDeclareTxnType` interface
2. `GetConstructorCalldata` and `GetContractAddressSalt` ::> `BroadcastAddDeployTxnType` interface

As these are the best distinguishing factors for objects that implements respective interfaces


PsychoPunkSage:
I have a question...
Shouldn't [THIS PART](https://github.com/PsychoPunkSage/starknet.go/blob/main/rpc/write_test.go#L20) be assigned to `BroadcastDeclareTxnType` instead of `BroadcastAddDeployTxnType`.
Cause its a test for `DeclareTransaction` not `AddDeployTxn`...

rianhughes:
>Shouldn't [THIS PART](https://github.com/PsychoPunkSage/starknet.go/blob/main/rpc/write_test.go#L20) be assigned to BroadcastDeclareTxnType instead of BroadcastAddDeployTxnType.

Yes, you're correct, nice catch!

You'll also need to remove the switch-case logic in the AddDeclareTransaction method. After that you should be able to fix the tests (which will require updating mock_starknet_addDeclareTransaction (since this function is called in the mock tests))

PsychoPunkSage:
Done, Updated everything, ready to be merged!!

rpc/types_broadcast_transaction.go
```diff
@@ -21,15 +21,20 @@ var (
...
 var (
 	_ BroadcastDeclareTxnType = BroadcastDeclareTxnV1{}
 	_ BroadcastDeclareTxnType = BroadcastDeclareTxnV2{}
 	_ BroadcastDeclareTxnType = BroadcastDeclareTxnV3{}
 )
 
-type BroadcastAddDeployTxnType interface{}
+type BroadcastAddDeployTxnType interface {
+	GetConstructorCalldata() []*felt.Felt
+	GetContractAddressSalt() *felt.Felt
```
rianhughes: nit: since the purpose of these methods are to contrain the interface, I think we only need one of them, so imo we can remove (eg) GetContractAddressSalt

PsychoPunkSage: Sure,
Done that and rebased the PR...
ready to be merged

rianhughes (approved):

Issue #567, opened 2024-05-29
Starknet transaction types supports behind the mainnet 


gilfoyle2021asc:
the code is too old to support V3 transactions of DECLARE and INVOKE etc.

thiagodeev:
Hi @gilfoyle2021asc.
Please give more details about the problem you are facing. The Declare and Invoke V3 transactions do exist in the starknet-go v0.7.0

thiagodeev:
Closing due to lack of activity. If the problem persists, please open another issue.

PR #588, opened 2024-07-04
chore ::> Merging all TransactionReceipt types into a single type

### LINKED ISSUE #576: Merge TransactionReceipt types into a single type
starknet go splits the receipts into distinct types. However, we can create a better user experience by having a single transaction receipt type (super set) and only populating the fields as necessary etc. This will also reduce the amount of code we need to maintain which is nice.

PsychoPunkSage:
Closes #576 

Here I created a `superset` type that can accommodate all the fields required by the different transaction receipts. This include includes all the fields from the various receipt types. Then I update the `unmarshalling` logic to populate this new superset type based on the fields present in the JSON data.


PsychoPunkSage:
hi @rianhughes 
I tried to formulate a way to solve #567 
Can you please give some suggestion on the approach before I start working on its proper integration with rest of the repo?
My method is in a temporary file called `rpc/types_transaction_receipt_pps.go`.... I will later on put the logic inside `rpc/types_transaction_receipt.go` once the approach is finalized.

rpc/types_transaction_receipt.go
```diff
@@ -99,115 +77,23 @@ func (tt *TransactionType) UnmarshalJSON(data []byte) error {
...
-	return tr.TransactionHash
-}
-
-// GetExecutionStatus returns the execution status of the L1HandlerTransactionReceipt.
-func (tr L1HandlerTransactionReceipt) GetExecutionStatus() TxnExecutionStatus {
-	return tr.ExecutionStatus
-}
-
 type ComputationResources struct {
-	// The number of Cairo steps used
```
rianhughes, line 188: We should keep these doc strings imo

rpc/types_transaction_receipt.go
```diff
@@ -217,189 +103,83 @@ func (er *ComputationResources) Validate() bool {
...
-type TransactionReceipt interface {
-	Hash() *felt.Felt
-	GetExecutionStatus() TxnExecutionStatus
-}
-
-type OrderedMsg struct {
-	// The order of the message within the transaction
-	Order   int `json:"order"`
-	MsgToL1 MsgToL1
-}
```
rianhughes, line 242: You'll need to keep this type

rpc/types_transaction_receipt.go
```diff
@@ -217,189 +103,83 @@ func (er *ComputationResources) Validate() bool {
...
-type MsgFromL1 struct {
-	// FromAddress The address of the L1 contract sending the message
-	FromAddress string `json:"from_address"`
-	// ToAddress The target L2 address the message is sent to
-	ToAddress *felt.Felt `json:"to_address"`
-	// EntryPointSelector The selector of the l1_handler in invoke in the target contract
-	Selector *felt.Felt `json:"entry_point_selector"`
-	//Payload  The payload of the message
-	Payload []*felt.Felt `json:"payload"`
-}
```
rianhughes, line 262: You'll need to keep this type

rpc/types_transaction_receipt.go
PsychoPunkSage: Done

rpc/types_transaction_receipt.go
PsychoPunkSage: Done

rpc/types_transaction_receipt.go
PsychoPunkSage: Done

PsychoPunkSage:
@rianhughes 
Integrated everything with a common `TransactionReceipt`.
All test are passing.
Please have a look whenever you are free.

rpc/types_transaction_receipt.go
```diff
@@ -21,32 +46,29 @@ const (
...
+	RevertReason       string             `json:"revert_reason,omitempty"`
 	Events             []Event            `json:"events"`
 	ExecutionResources ExecutionResources `json:"execution_resources"`
+	ContractAddress    *felt.Felt         `json:"contract_address,omitempty"`
+	MessageHash        NumAsHex           `json:"message_hash,omitempty"`
 }
 
-// Hash returns the transaction hash associated with the CommonTransactionReceipt.
-func (tr CommonTransactionReceipt) Hash() *felt.Felt {
+func (tr TransactionReceipt) Hash() *felt.Felt {
```
thiagodeev: Since the [Transaction Receipt interface](https://github.com/NethermindEth/starknet.go/pull/588/files#diff-3efe168df0042abb897e9f9a4fb39becbca4bd49b6c3575b957e3748cf1a38faL233) no longer exists, we can remove the Hash and GetExecutionStatus methods as they were just used to fit the interface.

rpc/types_transaction_receipt.go
```diff
@@ -358,48 +189,66 @@ type TxnStatusResp struct {
...
+		TransactionReceipt: t.TransactionReceipt,
+		BlockHash:          t.BlockHash.String(),
+		BlockNumber:        t.BlockNumber,
+	}
+
+	return json.Marshal(aux)
+}
+
+func (tr *TransactionReceiptWithBlockInfo) UnmarshalJSON(data []byte) error {
+	var aux struct {
```
thiagodeev, line 204: We can change it to something like this to avoid unnecessary code:

```
var aux struct {
  TransactionReceipt
  BlockHash   string `json:"block_hash,omitempty"`
  BlockNumber uint   `json:"block_number,omitempty"`
}
```

rpc/types_transaction_receipt.go
```diff
@@ -358,48 +189,66 @@ type TxnStatusResp struct {
...
+		MessageHash        NumAsHex           `json:"message_hash,omitempty"`
+		BlockHash          string             `json:"block_hash,omitempty"`
+		BlockNumber        uint               `json:"block_number,omitempty"`
+	}
 
 	if err := json.Unmarshal(data, &aux); err != nil {
 		return err
 	}
+
+	tr.TransactionReceipt = TransactionReceipt{
```
thiagodeev: And then, we can also change it to something like this to avoid unnecessary code:

`tr.TransactionReceipt = aux.TransactionReceipt`

thiagodeev (changes requested):
Hey @PsychoPunkSage, good job. I left few comments

rpc/types_transaction_receipt.go
PsychoPunkSage: The thing is those `getter function` are being used in `account/account_test.go`. So I decided not to remove it earlier as it provide more clarity about what the test is doing.

rpc/types_transaction_receipt.go
PsychoPunkSage: Done

rpc/types_transaction_receipt.go
PsychoPunkSage: Done

rpc/types_transaction_receipt.go
PsychoPunkSage: Done

rianhughes (approved):

thiagodeev (approved):
Good job!

PR #589, opened 2024-07-05
Update tests to Sepolia

### LINKED ISSUE #573: Fix RPC tests
TestDeclareTransaction
TestSpecVersion
TestTransactionReceipt
Also, remove integration data from rpc/tests/, and update any tests that depend on them to use Seploia data

thiagodeev:
Fix #573


rpc/types_block_test.go
```diff
@@ -130,40 +132,431 @@ func TestBlock_Unmarshal(t *testing.T) {
...
+				require.NotEmpty(pBlock.ParentHash, "Error in PendingBlockWithReceipts ParentHash")
+				require.NotEmpty(pBlock.SequencerAddress, "Error in PendingBlockWithReceipts SequencerAddress")
+				require.NotEmpty(pBlock.Timestamp, "Error in PendingBlockWithReceipts Timestamp")
+				require.NotEmpty(pBlock.Transactions, "Error in PendingBlockWithReceipts Transactions")
+			}
+
+		default:
+			t.Fatalf("unexpected block type, found: %T\n", resultType)
+		}
 	}
```
rianhughes: Can we do a final require.Equal(block, expectedBlock) at the end, just to make sure we don't miss anything?

thiagodeev: Done

rpc/errors.go
```diff
@@ -61,7 +60,11 @@ func tryUnwrapToRPCErr(err error, rpcErrors ...*RPCError) *RPCError {
 			return &nodeErr
 		}
 	}
-	return Err(InternalError, fmt.Sprintln(nodeErr.Code, nodeErr.Message, nodeErr.Data))
+
+	if nodeErr.Code == 0 {
+		return Err(InternalError, err.Error())
+	}
+	return Err(nodeErr.Code, nodeErr.Data)
```
rianhughes, line 67: Why the change here?

rpc/types_block_test.go
```diff
@@ -130,40 +133,156 @@ func TestBlock_Unmarshal(t *testing.T) {
...
+				require.Equal(pBlock.ParentHash, test.ExpectedPendingBlockWithReceipts.ParentHash, "Error in PendingBlockWithReceipts ParentHash")
+				require.Equal(pBlock.SequencerAddress, test.ExpectedPendingBlockWithReceipts.SequencerAddress, "Error in PendingBlockWithReceipts SequencerAddress")
+				require.Equal(pBlock.Timestamp, test.ExpectedPendingBlockWithReceipts.Timestamp, "Error in PendingBlockWithReceipts Timestamp")
+				require.Equal(pBlock.Transactions, test.ExpectedPendingBlockWithReceipts.Transactions, "Error in PendingBlockWithReceipts Transactions")
+				require.Exactly(pBlock, test.ExpectedPendingBlockWithReceipts)
+			} else {
+				require.NotEmpty(pBlock.ParentHash, "Error in PendingBlockWithReceipts ParentHash")
+				require.NotEmpty(pBlock.SequencerAddress, "Error in PendingBlockWithReceipts SequencerAddress")
+				require.NotEmpty(pBlock.Timestamp, "Error in PendingBlockWithReceipts Timestamp")
+				require.NotEmpty(pBlock.Transactions, "Error in PendingBlockWithReceipts Transactions")
```
rianhughes: Do you think we can do something like this:

        resultBlockWithReceipts, ok := (result).(*BlockWithReceipts)
		require.True(ok)
		require.Equal(resultBlockWithReceipts, test.ExpectedBlockWithReceipts)

rianhughes:
Let's also update the rpc/test/contracts to use sepolia classes

rpc/errors.go
thiagodeev: Because the previous version was like "If this error isn't one of these passed by the parameter, I'll set it to internal error without checking if it is a valid error of another type".

The code unmarshals the error to an RPCError in the `nodeErr` var so if the "nodeErr.Code" is zero (default int value, meaning that it hasn't a code) that means that it is not a valid RPCError, so we consider as "Internal error". But, if it is not zero, that means that this error is an existing error, so let's at least pass it to the "Err" function (in this function, the default value is InternalError anyway).

I did this because I faced a situation where the error was "InvalidParams", a valid case in the switch of the "Err" function but the code was hardcoding "InternalError" without even validating it.

rpc/types_block_test.go
thiagodeev: Yes, makes sense :). But I left some lines as there are some cases where the `test.ExpectedBlockWithReceipts` is nil

rianhughes (approved):

PR #591, opened 2024-07-11
chore::> Change input type of `SignInvokeTransaction`, `SignDeployAccountTransaction`, and `SignDeclareTransaction`

### LINKED ISSUE #590: Improve Transaction Handling: `WaitForTransactionReceipt`, Signing, & Keystore
Following improvements were raised by @thiagodeev 

1. The `WaitForTransactionReceipt` wasn't working as intended in the examples. Is there any potential issue in the method?

2. The `SignInvokeTransaction`, `SignDeployAccountTransaction`, and `SignDeclareTransaction` should receive a `Broadcast<type>Transaction` instead of a `<type>transaction` since is the Broadcast tx that will be sent with the Add<type>Transaction methods. We can also think about having an `account.SignTransaction` method that under the root calls dynamically one of these methods...

3. We can think about having an `account.AddTransaction` method that under the root calls dynamically one of `Add<type>Transaction` methods.

4. The `MemKeystore` needs both public and private keys as parameters to be initialized, but sometimes it is a bit difficult to find the public key. Why not just require the private key from the user and use some methods available by the **curve** package to dynamically derive the public key?

PsychoPunkSage:
For issue #590 

This PR will Solve #590 **pointer 2**

### DIFF
--- a/account/account.go
+++ b/account/account.go
@@ -30,10 +30,10 @@ var (
 type AccountInterface interface {
 	Sign(ctx context.Context, msg *felt.Felt) ([]*felt.Felt, error)
 	TransactionHashInvoke(invokeTxn rpc.BroadcastInvokeTxnType) (*felt.Felt, error)
-	TransactionHashDeployAccount(tx rpc.DeployAccountType, contractAddress *felt.Felt) (*felt.Felt, error)
+	TransactionHashDeployAccount(tx rpc.BroadcastAddDeployTxnType, contractAddress *felt.Felt) (*felt.Felt, error)
 	TransactionHashDeclare(tx rpc.BroadcastDeclareTxnType) (*felt.Felt, error)
 	SignInvokeTransaction(ctx context.Context, tx rpc.BroadcastInvokeTxnType) error
-	SignDeployAccountTransaction(ctx context.Context, tx *rpc.BroadcastDeployAccountTxn, precomputeAddress *felt.Felt) error
+	SignDeployAccountTransaction(ctx context.Context, tx rpc.BroadcastAddDeployTxnType, precomputeAddress *felt.Felt) error
 	SignDeclareTransaction(ctx context.Context, tx rpc.BroadcastDeclareTxnType) error
 	PrecomputeAccountAddress(salt *felt.Felt, classHash *felt.Felt, constructorCalldata []*felt.Felt) (*felt.Felt, error)
 	WaitForTransactionReceipt(ctx context.Context, transactionHash *felt.Felt, pollInterval time.Duration) (*rpc.TransactionReceiptWithBlockInfo, error)
@@ -140,9 +140,9 @@ func (account *Account) SignInvokeTransaction(ctx context.Context, invokeTx rpc.
 // - precomputeAddress: the precomputed address for the transaction
 // Returns:
 // - error: an error if any
-func (account *Account) SignDeployAccountTransaction(ctx context.Context, deployAccountTx *rpc.BroadcastDeployAccountTxn, precomputeAddress *felt.Felt) error {
+func (account *Account) SignDeployAccountTransaction(ctx context.Context, deployAccountTx rpc.BroadcastAddDeployTxnType, precomputeAddress *felt.Felt) error {
 
-	hash, err := account.TransactionHashDeployAccount(*deployAccountTx, precomputeAddress)
+	hash, err := account.TransactionHashDeployAccount(deployAccountTx, precomputeAddress)
 	if err != nil {
 		return err
 	}
@@ -150,7 +150,15 @@ func (account *Account) SignDeployAccountTransaction(ctx context.Context, deploy
 	if err != nil {
 		return err
 	}
-	deployAccountTx.Signature = signature
+
+	switch tx := (deployAccountTx).(type) {
+	case rpc.BroadcastDeployAccountTxn:
+		tx.Signature = signature
+	case rpc.BroadcastDeployAccountTxnV3:
+		tx.Signature = signature
+	default:
+		return errors.New("unsupported (invoke) transaction type")
+	}
 	return nil
 }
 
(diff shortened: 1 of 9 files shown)


account/account.go
```diff
@@ -32,9 +32,9 @@ type AccountInterface interface {
 	TransactionHashInvoke(invokeTxn rpc.InvokeTxnType) (*felt.Felt, error)
 	TransactionHashDeployAccount(tx rpc.DeployAccountType, contractAddress *felt.Felt) (*felt.Felt, error)
 	TransactionHashDeclare(tx rpc.DeclareTxnType) (*felt.Felt, error)
-	SignInvokeTransaction(ctx context.Context, tx *rpc.InvokeTxnV1) error
-	SignDeployAccountTransaction(ctx context.Context, tx *rpc.DeployAccountTxn, precomputeAddress *felt.Felt) error
-	SignDeclareTransaction(ctx context.Context, tx *rpc.DeclareTxnV2) error
+	SignInvokeTransaction(ctx context.Context, tx *rpc.BroadcastInvokev1Txn) error
```
thiagodeev: Let's change that.
The previous and this new versions set a specific broadcasTx version for both SignInvoke and SignDeclare methods (i.e., 'BroadcastInvokev1Txn' and 'BroadcastDeclareTxnV2'), let's require a general BroadcastInvokeTxnType and BroadcastDeclareTxnType instead.

rpc/types_broadcast_transaction.go
```diff
@@ -82,15 +82,8 @@ func (tx BroadcastDeclareTxnV1) GetContractClass() interface{} {
...
-	Type TransactionType `json:"type"`
-	// SenderAddress the address of the account contract sending the declaration transaction
-	SenderAddress     *felt.Felt         `json:"sender_address"`
-	CompiledClassHash *felt.Felt         `json:"compiled_class_hash"`
-	MaxFee            *felt.Felt         `json:"max_fee"`
-	Version           TransactionVersion `json:"version"`
-	Signature         []*felt.Felt       `json:"signature"`
-	Nonce             *felt.Felt         `json:"nonce"`
-	ContractClass     ContractClass      `json:"contract_class"`
+	DeclareTxnV2
```
thiagodeev: Unfortunately, we can't do that because the `BroadcastDeclareTxnV2` doesn't have the 'ClassHash' field required by the `DeclareTxnV2` tx. Let's move it back.
Ref: https://github.com/starkware-libs/starknet-specs/blob/e5ce8c6f99672d9ddf85a8e1d6ef8e87169d6e8d/api/starknet_api_openrpc.json#L2068

thiagodeev (changes requested):
Nice @PsychoPunkSage.
I left some observations

thiagodeev:
Also, the code is failing in some CI test

rpc/types_broadcast_transaction.go
PsychoPunkSage: Done

account/account.go
```diff
@@ -387,42 +416,19 @@ func dataAvailabilityMode(feeDAMode, nonceDAMode rpc.DataAvailabilityMode) (uint
 // - error: an error, if any
 //
 // If the `tx` parameter is not one of the supported types, the function returns an error `ErrTxnTypeUnSupported`.
-func (account *Account) TransactionHashDeclare(tx rpc.DeclareTxnType) (*felt.Felt, error) {
+func (account *Account) TransactionHashDeclare(tx rpc.BroadcastDeclareTxnType) (*felt.Felt, error) {
 
 	switch txn := tx.(type) {
-	case rpc.DeclareTxnV0:
-		// Due to inconsistencies in version 0 hash calculation we don't calculate the hash
```
thiagodeev, line 394: Why did you do that? The description says about the V0 and you removed the V0 and applied the comment to the V1, you also removed the V1 hash calculation. Let's put them back

account/account.go
```diff
@@ -387,42 +416,19 @@ func dataAvailabilityMode(feeDAMode, nonceDAMode rpc.DataAvailabilityMode) (uint
...
-			&felt.Zero,
-			calldataHash,
-			txn.MaxFee,
-			account.ChainId,
-			[]*felt.Felt{txn.Nonce},
-		)
-	case rpc.DeclareTxnV2:
-		if txn.CompiledClassHash == nil || txn.SenderAddress == nil || txn.Version == "" || txn.ClassHash == nil || txn.MaxFee == nil || txn.Nonce == nil {
+	case rpc.BroadcastDeclareTxnV2:
+		if txn.CompiledClassHash == nil || txn.SenderAddress == nil || txn.Version == "" || txn.MaxFee == nil || txn.Nonce == nil {
```
thiagodeev: Let's validate if the ContractClass is present with something like this: `txn.ContractClass.SierraProgram == nil`

account/account.go
```diff
@@ -387,42 +416,19 @@ func dataAvailabilityMode(feeDAMode, nonceDAMode rpc.DataAvailabilityMode) (uint
...
 			return nil, ErrNotAllParametersSet
 		}
 
-		calldataHash, err := hash.ComputeHashOnElementsFelt([]*felt.Felt{txn.ClassHash})
+		calldataHash, err := hash.ComputeHashOnElementsFelt([]*felt.Felt{txn.CompiledClassHash})
```
thiagodeev: This is wrong, we [still need](https://docs.starknet.io/architecture-and-concepts/network-architecture/transactions/#v2_deprecated_hash_calculation) the hash of the ClassHash. As we don't have the ClassHash field in the BroadcastDeclareTx, we will have to compute it from the ContractClass; we can use the hash.ClassHash function to do this

rpc/types_broadcast_transaction.go
```diff
@@ -107,15 +107,15 @@ func (tx BroadcastDeclareTxnV3) GetContractClass() interface{} {
 }
 
 type BroadcastDeployAccountTxn struct {
-	DeployAccountTxn
+	*DeployAccountTxn
```
thiagodeev, line 110: Let's do this with BroadcastInvokes and BroadcastDeclareV3 too

thiagodeev (changes requested):
Hey @PsychoPunkSage, there are a few things that need fixing

thiagodeev:
As discussed internally, we'll need first to create a function to calculate the class hash of legacy contract classes. It'll be done soon

thiagodeev:
Thanks @PsychoPunkSage!
Closing this since something similar has been implemented in the new starknet.go version.

PR #592, opened 2024-07-11
chore::> Introduce common `account.SendTransaction` function to dynamically call `Add<Type>Transaction`.

PsychoPunkSage:
Addresses #590 pointer 3


account/account.go
```diff
@@ -533,6 +533,27 @@ func (account *Account) WaitForTransactionReceipt(ctx context.Context, transacti
 	}
 }
 
+// AddTransaction calls dynamically one of Add<type>Transaction methods
```
thiagodeev: Let's make a more didactic description for better user understanding.
It's a function that can send Invoke, Declare, and Deploy transactions. Explain it your way.

account/account.go
```diff
@@ -533,6 +533,27 @@ func (account *Account) WaitForTransactionReceipt(ctx context.Context, transacti
...
+//
+// Parameters:
+// - ctx: the context.Context object for the transaction.
+// - txn: this Broadcast Transaction to be called.
```
thiagodeev: grama: "the Broadcast Transaction to be sent."**

account/account.go
```diff
@@ -533,6 +533,27 @@ func (account *Account) WaitForTransactionReceipt(ctx context.Context, transacti
...
+// Returns:
+// - interface{} returns required Broadcast txn return value.
+// - error: an error if any.
+func (account *Account) AddTransaction(ctx context.Context, txn rpc.BroadcastTxn) (interface{}, error) {
```
thiagodeev: Let's change the function name to "SendTransaction", it's more didactic.

account/account.go
```diff
@@ -533,6 +533,27 @@ func (account *Account) WaitForTransactionReceipt(ctx context.Context, transacti
...
+	switch tx := txn.(type) {
+	case rpc.BroadcastInvokeTxnType:
+		return account.AddInvokeTransaction(ctx, tx)
```
thiagodeev: Let's remove these three methods from the "account" package. Call them directly from the "account.provider" instead. You can also remove [this ](https://github.com/NethermindEth/starknet.go/blob/main/account/account.go#L43)line

account/account.go
thiagodeev: Let's improve the return value. Creating a general "Transaction Response" just like we did with the "Transaction Receipt" might be the solution...

examples/deployAccount/main.go
```diff
@@ -121,13 +121,14 @@ func main() {
 	fmt.Scan(&input)
 
 	// Send transaction to the network
-	resp, err := accnt.AddDeployAccountTransaction(context.Background(), tx)
+	// resp, err := accnt.AddDeployAccountTransaction(context.Background(), tx)
```
thiagodeev: Lost comment

account/account_test.go
```diff
@@ -526,10 +526,10 @@ func TestAddInvoke(t *testing.T) {
 		err = acnt.SignInvokeTransaction(context.Background(), &test.InvokeTx.InvokeTxnV1)
 		require.NoError(t, err)
 
-		resp, err := acnt.AddInvokeTransaction(context.Background(), test.InvokeTx)
+		resp, err := acnt.AddTransaction(context.Background(), test.InvokeTx)
 		if err != nil {
 			require.Equal(t, test.ExpectedErr.Error(), err.Error(), "AddInvokeTransaction returned an unexpected error")
-			require.Nil(t, resp)
+			require.Nil(t, resp.(*rpc.AddInvokeTransactionResponse))
```
thiagodeev: This and all other similiar type assertions could be removed when the general Transaction Response type is ready

thiagodeev (changes requested):
Nice @PsychoPunkSage.
I left some comments

account/account.go
PsychoPunkSage: Done

account/account.go
PsychoPunkSage: Done

account/account.go
PsychoPunkSage: Done

account/account.go
PsychoPunkSage: Done

PsychoPunkSage: Done

PsychoPunkSage:
Hey @thiagodeev 
Some documentation changes are left...
Rest everything is done..
Have a look at it.

examples/deployAccount/main.go
PsychoPunkSage: Done

account/account.go
```diff
@@ -533,40 +532,37 @@ func (account *Account) WaitForTransactionReceipt(ctx context.Context, transacti
...
-// AddInvokeTransaction generates an invoke transaction and adds it to the account's provider.
+// SendTransaction can send Invoke, Declare, and Deploy transactions. It provides a unified way to send different transactions.
 //
 // Parameters:
 // - ctx: the context.Context object for the transaction.
-// - invokeTx: the invoke transaction to be added.
+// - txn: this Broadcast Transaction to be sent.
 // Returns:
-// - *rpc.AddInvokeTransactionResponse: The response for the AddInvokeTransactionResponse
+// - interface{} returns required Broadcast txn return value.
```
thiagodeev: 1 - ~~this~~ **the** Broadcast Transaction to be sent.
2 - Wrong type description, it's not "interface{}". Also, let's follow the pattern= returnType: description

rpc/types_transaction_response.go
```diff
@@ -18,3 +18,30 @@ type AddDeployAccountTransactionResponse struct {
 type AddInvokeTransactionResponse struct {
 	TransactionHash *felt.Felt `json:"transaction_hash"`
 }
+
+type TransactionResponse struct {
+	TransactionHash *felt.Felt `json:"transaction_hash"`
+	ClassHash       *felt.Felt `json:"class_hash"`
+	ContractAddress *felt.Felt `json:"contract_address"`
```
thiagodeev: Let's add the `omitempty` tag to the ClassHash and ContractAddress fields, that way we don't need the "ConvertToTransactionResponse" function nor the AddInvokeTransactionResponse, AddDeployAccountTransactionResponse, and the AddInvokeTransactionResponse types; we can remove them.

thiagodeev (changes requested):
Great @PsychoPunkSage! We're almost there.
Left a few comments

account/account.go
PsychoPunkSage: done

rpc/types_transaction_response.go
PsychoPunkSage: done

PsychoPunkSage:
@thiagodeev 
I have made required changes...
But for now we shouldn't remove AddInvokeTransactionResponse, AddDeployAccountTransactionResponse, and the AddInvokeTransactionResponse type cause it is used in multiple places...
Plus I don't want this PR to get any longer for now :)
What do you think?

rpc/types_transaction_response.go
```diff
@@ -18,3 +18,30 @@ type AddDeployAccountTransactionResponse struct {
...
 	TransactionHash *felt.Felt `json:"transaction_hash"`
 }
+
+type TransactionResponse struct {
+	TransactionHash *felt.Felt `json:"transaction_hash"`
+	ClassHash       *felt.Felt `json:"class_hash,omitempty"`
+	ContractAddress *felt.Felt `json:"contract_address,omitempty"`
+}
+
+// func ConvertToTransactionResponse(resp interface{}) *TransactionResponse {
```
thiagodeev: Let's remove this comment

thiagodeev (changes requested):

thiagodeev:
Okay @PsychoPunkSage, I agree with you. Since we have these TxResponse being used by provider methods, let's leave them for another time ;)

rpc/types_transaction_response.go
PsychoPunkSage: Ouch; 
forgot to remove it :)

PsychoPunkSage:
Done @thiagodeev

account/account_test.go
thiagodeev, line 1: The best would be to have a dedicated TestSendTransaction test case, but for now, let's at least rename the three tests to SendInvokeTxn, SendDeployAccountTxn, and SendDeclareTxn and update their descriptions.

thiagodeev (changes requested):
It's almost completed!
Just one thing before merging:

account/account_test.go
PsychoPunkSage: Done @thiagodeev

thiagodeev (approved):
Great @PsychoPunkSage! LGTM

account/account.go
```diff
@@ -40,7 +40,6 @@ type AccountInterface interface {
 }
 
 var _ AccountInterface = &Account{}
-var _ rpc.RpcProvider = &Account{}
```
rianhughes, line 44: Imo we should keep this in. If a new method is added to the RPC spec then this helps ensure that the account type implements it. What do you think @thiagodeev

rianhughes: Otherwise the PR looks good

thiagodeev: This line was removed because as the AddDeclare/Invoke/DeployAcc methods were removed from the Account type it no longer implements the RPC interface.
IMO, the Account type doesn't need to implement the RPC interface because they are separate abstractions. If the user wants to use an RPC method it can just access the 'provider' Account type property. I see this approach in the Ethers.js library https://docs.ethers.org/v6/api/providers/#Signer. 
We can adapt some provider methods to the context of the Account (e.g. Nonce could return the account nonce instead of being a duplicated version of the provider.Nonce) but no need to have all of them IMO.

PsychoPunkSage:
Hi @rianhughes @thiagodeev 
I think this PR is ready tobe merged

PR #594, opened 2024-07-11
chore::> Derive public key from private key using `crypto` package

PsychoPunkSage:
Addresses #590 pointer 4


PsychoPunkSage:
@thiagodeev 
were you talking about something like this??

account/keystore.go
```diff
@@ -38,14 +40,21 @@ func NewMemKeystore() *MemKeystore {
 // SetNewMemKeystore returns a new instance of MemKeystore and sets the given public key and private key in it.
 //
 // Parameters:
-// - pub: a string representing the public key
 // - priv: a pointer to a big.Int representing the private key
 // Returns:
 // - *MemKeystore: a pointer to the newly created MemKeystore instance
-func SetNewMemKeystore(pub string, priv *big.Int) *MemKeystore {
+// - error: if any error occurs during conversion
+func SetNewMemKeystore(priv *big.Int) (*MemKeystore, error) {
```
thiagodeev: Let's revert this commit and all these changes, we'll instead create a new file called "signer.go" in the same "account" package.
Inside it, create a struct like this:

```
type Signer struct {
	keystore *MemKeystore
	publicKey string
}
```

In this file, create a NewSigner method that requires only the `privateKey string` parameter. Inside this function:
- get the publicKey with `curve.Curve.PrivateToPoint(privateKey)`
- create a new MemKeyStore passing the pub and priv keys
- instantiate a Signer struct passing these values above and return it

Also, create these methods for the Signer struct type:
- PublicKey() returning it
- MemKeyStore() returning the pointer to it
- Put(privateKey, *MemKeyStore) to put these values into it
- Sign(ctx, msgHash) calling the MemKeyStore.Sign passing the appropriate values

thiagodeev (changes requested):
Yes @PsychoPunkSage! That's exactly what I had in mind.
But now, I believe we can do better. Sorry about that, but I'm going to ask you to take another approach.

PsychoPunkSage:
Hmm,
Okay... I thought about this approach when I started working on the issue.. but it was not that appealing to me to create all the fn again. Instead I was thinking to modify the existing func and check whether user is passing a `pub_key` or not.. if not then do the required computation...

This approach will surely make things clean but don't you think it will be redundant??
What do you think @thiagodeev ??

thiagodeev:
This was also my first intention on this feature, but I realized that the MemKeyStore is not a simple privateKey pointer, is literally a "key store", that can store multiple keys and have multiple use cases. If we change it directly, users who are using this feature could be in trouble.
Being so, creating a new abstraction specific to our intention (a Signer), having a new or existing MemKeyStore linked (covering both use cases) and specific methods to it it's a valid approach IMO. What do you think?

PsychoPunkSage:
Mmmm,
Yeah I think for now we can do that, well, it can always be modified in future by any new contributor.
So, I will separate it...

account/keystore.go
PsychoPunkSage: Done

PsychoPunkSage:
@thiagodeev have a look

account/signer.go
```diff
@@ -0,0 +1,54 @@
...
+	"github.com/NethermindEth/starknet.go/curve"
+	"github.com/NethermindEth/starknet.go/utils"
+)
+
+type Signer struct {
+	keystore  *MemKeystore
+	publicKey string
+}
+
+func NewSigner(privateKey *big.Int) (*Signer, error) {
```
thiagodeev: Let's receive the privateKey as a string and convert it to biting internally. This will be simpler for the user as we usually have it as hexString in env variables

account/signer.go
```diff
@@ -0,0 +1,54 @@
...
+
+func (s *Signer) PublicKey() string {
+	return s.publicKey
+}
+
+func (s *Signer) MemKeyStore() *MemKeystore {
+	return s.keystore
+}
+
+func (s *Signer) Put(priv *big.Int) {
```
thiagodeev: Same as above, let's request privateKey as a string. Also, this method should allow pointing an existing MemKeyStore to this signer, that way the dev could create only one keystore map and have multiple signers using it. It's like creating a new signer but also passing an existing MemKeyStore pointer to it

thiagodeev (changes requested):
Great job @PsychoPunkSage!
I'll ask one general thing: please put descriptions in the code (package, functions, and methods).
Besides that, I left a few comments

account/signer.go
PsychoPunkSage: done

account/signer.go
PsychoPunkSage: Hey @thiagodeev 
didn't get what u said about existing MemKeyStore...
I pushed the changes. please look whether you talking about it or something else??

thiagodeev: If you take a look at the MemKeyStore struct you'll notice that it is a map[string]*big.Int, so when we instantiate a new Signer here we're just creating a new map and setting our pub and priv key as key/values in that map.
So, if the user already has a MemKeyStore instantiated storing other keys, why not use this existing map to store new keys instead of creating a new one?
Makes sense to you now?

account/signer.go
```diff
@@ -13,13 +14,23 @@ type Signer struct {
...
+		return nil, fmt.Errorf("invalid private key format")
+	}
+
 	pubKey, err := getPublicKey(privateKey)
 	if err != nil {
 		return nil, err
 	}
 
-	keyStore := SetNewMemKeystore(pubKey, privateKey)
+	var keyStore *MemKeystore
```
thiagodeev: I believe you confused something here, the existing MemKeyStore was for the Put method. The Put method should accept a privKey and a MemKeyStore pointer as parameter. Let's remove this code from this method

account/signer.go
```diff
@@ -13,13 +14,23 @@ type Signer struct {
 	publicKey string
 }
 
-func NewSigner(privateKey *big.Int) (*Signer, error) {
+func NewSigner(privateKeyHex string, existingKeystore *MemKeystore) (*Signer, error) {
+	privateKey, ok := new(big.Int).SetString(privateKeyHex, 16)
```
thiagodeev: Instead of "privateKeyHex" let's go back to the "privateKey" param name, because if we pass 0 to the "SetString" function as the base parameter instead of 16 the function dynamically assumes the current base based on the string prefix. No need to be only a hex value. Ref: https://pkg.go.dev/math/big#Int.SetString

account/signer.go
```diff
@@ -13,13 +14,23 @@ type Signer struct {
...
+	if !ok {
+		return nil, fmt.Errorf("invalid private key format")
```
thiagodeev: nit: "invalid private key **value**"

account/signer.go
```diff
@@ -35,8 +46,13 @@ func (s *Signer) MemKeyStore() *MemKeystore {
 	return s.keystore
 }
 
-func (s *Signer) Put(priv *big.Int) {
-	s.keystore.Put(s.publicKey, priv)
+func (s *Signer) Put(priv string) error {
+	privateKey, ok := new(big.Int).SetString(priv, 16)
+	if !ok {
+		return fmt.Errorf("invalid private key format")
```
thiagodeev: nit: "invalid private key value"

thiagodeev (changes requested):
Hey @PsychoPunkSage, left some comments.
An observation: you need to put golang doc code comments

account/signer.go
PsychoPunkSage: done

account/signer.go
PsychoPunkSage: done

account/signer.go
PsychoPunkSage: done

account/signer.go
PsychoPunkSage: done

account/signer.go
PsychoPunkSage: done

PsychoPunkSage:
Hi @thiagodeev 
Did required changes. Please have a look whenever you find time.

account/signer.go
```diff
@@ -0,0 +1,115 @@
...
+// Put stores a new private key in the keystore.
+//
+// Parameters:
+// - priv: a string representing the private key in hexadecimal format.
+// - existingKeystore: a pointer to an existing MemKeystore to use. If nil, the Signer's keystore is used.
+//
+// Returns:
+// - error: an error if the private key is invalid.
+func (s *Signer) Put(priv string, existingKeystore *MemKeystore) error {
+	privateKey, ok := new(big.Int).SetString(priv, 16)
```
thiagodeev: Let's change the base to 0 to select it based on the string prefix

account/signer.go
```diff
@@ -0,0 +1,115 @@
...
+	}
+
+	var keystoreToUse *MemKeystore
+	if existingKeystore != nil {
+		keystoreToUse = existingKeystore
+	} else {
+		keystoreToUse = s.keystore
+	}
+
+	keystoreToUse.Put(s.publicKey, privateKey)
```
thiagodeev, line 88: The previous publicKey is incorrect here since the user is adding a new privateKey. We must first derive the publicKey from it, assign it to "s.publicKey", and then assign the pubK & privK with the "keystoreToUse.Put"

thiagodeev (changes requested):
Hey @PsychoPunkSage, left a few comments, otherwise looks good!

account/signer.go
PsychoPunkSage: oops Forgot to change it.
Done

account/signer.go
PsychoPunkSage: done

PsychoPunkSage:
@thiagodeev 
Did you find time to review the PR?

thiagodeev:
Thanks @PsychoPunkSage!
Sorry for the delay in reviewing it.
Are you still available? There's a next phase for this feature and it would be great if you could do it

PsychoPunkSage:
> Thanks @PsychoPunkSage! Sorry for the delay in reviewing it. Are you still available? There's a next phase for this feature and it would be great if you could do it

Hi @thiagodeev 
Yes I'm available. I can try to help...

PR #595, opened 2024-07-17
pedersen x range bug fix

rianhughes:
There seems to be a bug in the pedersen hash func 

ref: https://github.com/starkware-libs/cairo-lang/blob/de741b92657f245a50caab99cfaef093152fd8be/src/starkware/crypto/signature/fast_pedersen_hash.py#L27


thiagodeev (approved):
LGTM
https://github.com/starkware-libs/cairo-lang/blob/de741b92657f245a50caab99cfaef093152fd8be/src/starkware/crypto/signature/fast_pedersen_hash.py#L27

PR #598, opened 2024-07-18
`Makefile` and new linters for starknet.go

### LINKED ISSUE #597: need of Makefile
A **`makefile`** might be useful, we could have a commands like `make format`, `make lint`,  `make install-deps`, and `make test`.
Something like this [Junos](https://github.com/NethermindEth/juno/blob/main/Makefile)

PsychoPunkSage:
Basic template is ready for `Makefile`
Closes #597


PsychoPunkSage:
@rianhughes 
Please have a look at Makefile

Makefile
```diff
@@ -0,0 +1,68 @@
...
+test-verbose:
+	@go test ./... -v
+
+bench:
+	@go test -bench=.
+
+rpc-test:
+	@go test -v ./rpc -env [mainnet|devnet|testnet|mock]
+
+simple-call:
```
rianhughes: I don't think we should run examples in the make file

Makefile
```diff
@@ -0,0 +1,68 @@
+test:
```
rianhughes: Lets have a test command to run all mock tests in (verbose), and another for devnet tests (also verbose)

Makefile
PsychoPunkSage: I think It would be a good option to have it.
I have made sure that if the `developer` haven't set the environment correctly, then the correct procedure to run the examples will be printed in the terminal
like this::>
image

Makefile
PsychoPunkSage: Mmmm..
can you tell how to run `devnet test`??

PsychoPunkSage:
Hi @rianhughes 
I tried to implement `golangci-lint.yml` file
I am facing few problems
`make lint` gives me this error 
```
golangci-lint not installed. Run 'make install-golangci-lint' to install it.
make: *** [Makefile:26: lint] Error 1
```
Even though I have installed golangci-lint properly.
I found a way to solve it

```bash
ls $(go env GOPATH)/bin
```` 
Gives me list of go env paths which includes `golangci-lint`

So, on doing this.
```bash
export PATH=$PATH:$(go env GOPATH)/bin
```
`make lint` work and gives me desired results...
So I'm kind of lost here... as I cannot do this in makefile.

rianhughes:
> Hi @rianhughes I tried to implement `golangci-lint.yml` file I am facing few problems `make lint` gives me this error
> 
> ```
> golangci-lint not installed. Run 'make install-golangci-lint' to install it.
> make: *** [Makefile:26: lint] Error 1
> ```
> 
> Even though I have installed golangci-lint properly. I found a way to solve it
> 
> ```shell
> ls $(go env GOPATH)/bin
> ```
> 
> Gives me list of go env paths which includes `golangci-lint`
> 
> So, on doing this.
> 
> ```shell
> export PATH=$PATH:$(go env GOPATH)/bin
> ```
> 
> `make lint` work and gives me desired results... So I'm kind of lost here... as I cannot do this in makefile.

Maybe including the following in the Makefile would help?

GOPATH := $(shell go env GOPATH)
GOBIN := $(GOPATH)/bin

install-golangci-lint:
	@export PATH=$(GOBIN):$$PATH && \
	go install github.com/golangci/golangci-lint/cmd/golangci-lint@v1.59.1

PsychoPunkSage:
Hi @rianhughes 
Its not working for me...
I have to manually run `export PATH=$PATH:$(go env GOPATH)/bin` in terminal to make it (`make lint`) work. Cause `makefile` is unable to add `GOBIN` to `PATH`
Can you please see whether its the same in your end or not??

PsychoPunkSage:
Hi @rianhughes 
Did you find time to review this PR??

thiagodeev (approved):

thiagodeev:
Thanks @PsychoPunkSage!

Issue #622, opened 2024-09-17
update Juno version at go.mod file


thiagodeev:
Currently using Juno v0.3, but Juno is currently at v0.12. Let's update it

lucky777strike:
Juno v0.3 uses github.com/sourcegraph/sourcegraph/lib v0.0.0-20221216004406-749998a2ac74 which moved to other repo and now for build i need use 
replace github.com/sourcegraph/sourcegraph/lib v0.0.0-20221216004406-749998a2ac74 => github.com/sourcegraph/sourcegraph-public-snapshot/lib v0.0.0-20221216004406-749998a2ac74

PR #625, opened 2024-09-25
Upgrades Juno and Go version

thiagodeev:
Solves https://github.com/NethermindEth/starknet.go/issues/622


account/account_test.go
```diff
@@ -1195,6 +1194,7 @@ func TestAddDeclareTxn(t *testing.T) {
 // - []devnet.TestAccount: a slice of test accounts
 // - error: an error, if any
 func newDevnet(t *testing.T, url string) (*devnet.DevNet, []devnet.TestAccount, error) {
+	t.Helper()
```
thiagodeev, line 1197: Just to remove the "unused param" warning

rpc/types_block_test.go
```diff
@@ -269,7 +269,6 @@ func TestBlockWithReceipts(t *testing.T) {
 				require.NotEmpty(pBlock.ParentHash, "Error in PendingBlockWithReceipts ParentHash")
 				require.NotEmpty(pBlock.SequencerAddress, "Error in PendingBlockWithReceipts SequencerAddress")
 				require.NotEmpty(pBlock.Timestamp, "Error in PendingBlockWithReceipts Timestamp")
-				require.NotEmpty(pBlock.Transactions, "Error in PendingBlockWithReceipts Transactions")
```
thiagodeev, line 272: When created, the pending block may not have any transactions yet, so this line was failing sometimes when that was the case

.github/workflows/linter.yml
```diff
@@ -25,4 +25,4 @@ jobs:
         uses: golangci/golangci-lint-action@v3
         with:
           # Optional: version of golangci-lint to use in form of v1.2 or v1.2.3 or `latest` to use the latest version
-          version: v1.56.2
\ No newline at end of file
+          version: v1.61
```
thiagodeev, line 28: After changing the go version the linter was failing, but with this version it worked

rianhughes (approved):

Issue #623, opened 2024-09-24
Implement RPCv08-rc0


rianhughes:
https://github.com/starkware-libs/starknet-specs/tree/v0.8.0-rc0

TODO: 
- merge v0.8.0 with main to pass CI tests, and then, merge some contributor's PR
- make a PR to change code docs to include an extra space (e.g. `// - asdf...` to `//    - asdf...`. This will make the doc to be rendered as a markdown list.
- before merge [667](https://github.com/NethermindEth/starknet.go/pull/667), create a copy of main to be released as something like `v.0.8.0:SNIP-12`

#### Features
- [x] Implement websocket support

#### Methods
- [x] [starknet_getMessagesStatus](https://github.com/NethermindEth/starknet.go/issues/624)
- [x] starknet_getStorageProof
- [x] starknet_getCompiledCasm

websocket methods
- [x] starknet_subscribeNewHeads
- [x] starknet_subscriptionNewHeads
- [x] starknet_subscribeEvents
- [x] starknet_subscriptionEvents
- [x] starknet_subscribeTransactionStatus
- [x] starknet_subscriptionTransactionsStatus
- [x] starknet_subscribePendingTransactions
- [x] starknet_subscriptionPendingTransactions
- [x] starknet_subscriptionReorg
- [x] starknet_unsubscribe

#### Components
- [x] ~EVENT_KEYS~ (no need to change the code, just a doc improvement from their side)
- [x] ~L1_TXN_HASH~ (no need to change the code, just a doc improvement from their side)
- [x] TXN_STATUS_RESULT
- [x] MERKLE_NODE
- [x] NODE_HASH_TO_NODE_MAPPING
- [x] CONTRACT_EXECUTION_ERROR
- [x] CASM_COMPILED_CONTRACT_CLASS
- [x] CASM_ENTRY_POINT
- [x] CellRef
- [x] Deref
- [x] DoubleDeref
- [x] Immediate
- [x] BinOp
- [x] ResOperand
- [x] HINT
- [x] DEPRECATED_HINT
- [x] CORE_HINT
- [x] STARKNET_HINT
- [x] NEW_TXN_STATUS
- [x] SUBSCRIPTION_ID
- [x] REORG_DATA

#### Errors
- [x] COMPILATION_ERROR
- [x] INVALID_SUBSCRIPTION_ID
- [x] TOO_MANY_ADDRESSES_IN_FILTER
- [x] TOO_MANY_BLOCKS_BACK
[...]

PR #628, opened 2024-09-30
Thiagodeev/rpcv08 txn status

thiagodeev:

**Methods**
starknet_getTransactionStatus

**Components**
TXN_STATUS_RESULT
FEE_ESTIMATE

_Note_: as the RPC v0.8.0 has not been released yet, we can't fully test these changes as the nodes haven't implemented them yet. They will be updated before merge into main

### DIFF
--- a/rpc/block_test.go
+++ b/rpc/block_test.go
@@ -655,5 +655,4 @@ func validatePendingBlockHeader(t *testing.T, pBlock *PendingBlockHeader) {
 	require.NotZero(t, pBlock.L1GasPrice)
 	require.NotZero(t, pBlock.StarknetVersion)
 	require.NotZero(t, pBlock.L1DataGasPrice)
-	require.NotZero(t, pBlock.L1DAMode)
 }
--- a/rpc/contract_test.go
+++ b/rpc/contract_test.go
@@ -397,6 +397,7 @@ func TestNonce(t *testing.T) {
 //
 //	none
 func TestEstimateMessageFee(t *testing.T) {
+	//TODO: upgrade the testnet test case before merge
 	testConfig := beforeEach(t)
 
 	type testSetType struct {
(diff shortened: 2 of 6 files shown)


rpc/block_test.go
```diff
@@ -655,5 +655,4 @@ func validatePendingBlockHeader(t *testing.T, pBlock *PendingBlockHeader) {
 	require.NotZero(t, pBlock.L1GasPrice)
 	require.NotZero(t, pBlock.StarknetVersion)
 	require.NotZero(t, pBlock.L1DataGasPrice)
-	require.NotZero(t, pBlock.L1DAMode)
```
thiagodeev: This was failing because some blocks actually had "BLOB" (zero value of L1DAMode) as the `L1DAMode`, it's a valid option

rpc/contract_test.go
```diff
@@ -475,6 +478,8 @@ func TestEstimateMessageFee(t *testing.T) {
 }
 
 func TestEstimateFee(t *testing.T) {
+	//TODO: upgrade the mainnet and testnet test cases before merge
```
thiagodeev, line 481: As the nodes have not implemented v0.8.0, we can't test it with real data yet

rpc/block_test.go
rianhughes: Could we use 	require.NotNil(t, pBlock.L1DAMode)?

rpc/transaction_test.go
```diff
@@ -179,6 +179,8 @@ func TestTransactionReceipt(t *testing.T) {
 
 // TestGetTransactionStatus tests starknet_getTransactionStatus
 func TestGetTransactionStatus(t *testing.T) {
+	//TODO: implement a test case to 'failure_reason' before merge
```
rianhughes, line 182: t.Skip("todo: implement test for rpcv08-rc0 when full nodes support this rpc version")

rpc/contract_test.go
```diff
@@ -397,6 +397,7 @@ func TestNonce(t *testing.T) {
 //
 //	none
 func TestEstimateMessageFee(t *testing.T) {
+	//TODO: upgrade the testnet test case before merge
```
rianhughes, line 400: t.Skip with a note here as well

rpc/contract_test.go
```diff
@@ -475,6 +478,8 @@ func TestEstimateMessageFee(t *testing.T) {
 }
 
 func TestEstimateFee(t *testing.T) {
+	//TODO: upgrade the mainnet and testnet test cases before merge
+
```
rianhughes, line 482: Maybe we can just add another t.Skip("insert note here") instead of commenting most of the tests out

rianhughes:
Might be good to leave a note in the PR description that tests for these changes need implemented once full nodes support this rpc version

rpc/block_test.go
thiagodeev: Sounds good, I'll do it

rpc/contract_test.go
thiagodeev: I chose to keep the tests instead of skipping them because the mock environment is running without any problem; what do you think?

rianhughes: I was thinking the same, but thought t.Skip() might have been useful when most the tests are commented out. It also helps keep track of which tests we have turned off, and need updating. Also long as we are consistent and leave a note, it should be fine.

rianhughes (approved):

Issue #630, opened 2024-09-30
Bug - incorrect invoke transaction trace structure


dkatzan:
The go struct defined for the trace of an invoke transaction is incorrect, causing to missing data while parsing API results

specifically, the following field `ExecInvocation` struct defined [here](https://github.com/NethermindEth/starknet.go/blob/e171fe1153916745cd960bcdd3587c201c1f5a66/rpc/types_trace.go#L135) and used [here](https://github.com/NethermindEth/starknet.go/blob/e171fe1153916745cd960bcdd3587c201c1f5a66/rpc/types_trace.go#L48)
contains 2 inner fields, while the node result is slightly different, resulting with the inner `FnInvocation` to always be empty

Also, the relevant `TestSimulateTransaction` in mainnet is broken due to multiple reasons, part of them discussed [here](https://github.com/NethermindEth/starknet.go/issues/608)

I already have a draft work in progress, and will be happy to open a PR to fix these issues soon

PR #631, opened 2024-09-30
fix definition of invoke transaction trace

dkatzan:
This PR fixes a bug in invoke transaction trace described [here](https://github.com/NethermindEth/starknet.go/issues/630)


rianhughes (approved):
Changes LGTM, just need to fix the two failing testnet tests (we can do this on our end). Thanks for the PR!

dkatzan:
Thx @rianhughes , will it be possible to create a new tag version from the main branch?

PR #637, opened 2024-10-14
Implements SNIP-12

thiagodeev:
## Overview
This PR implements the [SNIP-12](https://github.com/starknet-io/SNIPs/blob/main/SNIPS/snip-12.md) specification for typed structured data hashing and signing in StarkNet. The implementation follows SNIP-12 which defines a standard way to encode structured data for signing, similar to EIP-712 in Ethereum.

### Key Features
- Full implementation of TypedData structure and encoding
- Support for both revision 0 and 1 of the specification
- Comprehensive type system including:
- Basic types (felt, shortstring, bool, etc.)
- Complex types (arrays, structs, enums)
- Special types (merkletree, string handling)
- Domain separation for message signing
- Structured hash calculation following the specification
- JSON unmarshaling support for easy data input

### Technical Details
- Implements type encoding and hashing according to SNIP-12 rules
- Supports nested structures and recursive type definitions
- Handles both StarkNet Domain and custom message types
- Includes comprehensive parameter validation
- Maintains backward compatibility with existing StarkNet message signing
- Testing
- Unit tests cover all major functionality
- Test vectors from the SNIP-12 specification
- Edge cases and error conditions

### Documentation
- Added detailed comments explaining the implementation
- Included references to the SNIP-12 specification
- Documented all public methods and types

This implementation enables Starknet.go users to work with typed structured data in a standardized way, matching the behavior of other StarkNet implementations.


utils/Felt.go
```diff
@@ -92,6 +102,20 @@ func FeltArrToBigIntArr(f []*felt.Felt) []*big.Int {
...
+// - f: the array of Felt objects to convert
+// Returns:
+// - []string: the array of string objects
+func FeltArrToStringArr(f []*felt.Felt) []string {
+	var stringArr []string
+	for _, felt := range f {
+		stringArr = append(stringArr, felt.String())
+	}
+	return stringArr
+}
```
rianhughes:
```suggestion
func FeltArrToStringArr(f []*felt.Felt) []string {
	stringArr := make([]string, len(f))
	for i, felt := range f {
		stringArr[i] = felt.String()
	}
	return stringArr
}
```

utils/Felt.go
```diff
@@ -117,7 +141,7 @@ func StringToByteArrFelt(s string) ([]*felt.Felt, error) {
 
 	arr := r.FindAllString(s, -1)
 	if len(arr) == 0 {
-		return []*felt.Felt{}, fmt.Errorf("invalid string no matches found, s: %s", s)
+		return []*felt.Felt{&felt.Zero, &felt.Zero, &felt.Zero}, nil
```
rianhughes, line 134: Why does this function now return a valid result instead of an error?

utils/keccak.go
```diff
@@ -46,6 +46,27 @@ func StrToBig(str string) *big.Int {
...
+	bigNum, ok := new(big.Int).SetString(str, 0)
+	if ok {
+		hex = bigNum.Text(16)
+		hex = "0x" + hex
+	} else {
+		hex = fmt.Sprintf("0x%x", str)
+	}
+
+	return hex
+}
```
rianhughes:
```suggestion
func StrToHex(str string) string {
	if strings.HasPrefix(str, "0x") {
		return str
	}

	if bigNum, ok := new(big.Int).SetString(str, 0); ok {
		return "0x" + bigNum.Text(16)
	}

	return "0x" + fmt.Sprintf("%x", str)
}
```

typedData/tests/example_array.json
```diff
@@ -0,0 +1,34 @@
+{
```
rianhughes, line 1: where did example_array.json come from?

typedData/tests/v1Nested.json
```diff
@@ -0,0 +1,66 @@
+{
```
rianhughes, line 1: where did v1Nested.json come from?

examples/typedData/main.go
```diff
@@ -0,0 +1,99 @@
...
+	privKeyBI, ok := new(big.Int).SetString(setup.GetPrivateKey(), 0)
+	if !ok {
+		panic("Fail to convert privKey to bitInt")
+	}
+	x, y, err := curve.Curve.PrivateToPoint(privKeyBI)
+	if err != nil {
+		panic(fmt.Errorf("fail to get point: %w", err))
+	}
+	isValid := curve.Curve.Verify(messageHash.BigInt(new(big.Int)), signature[0].BigInt(new(big.Int)), signature[1].BigInt(new(big.Int)), x, y)
+	fmt.Println("Verification result:", isValid)
```
rianhughes: What do you think about creating a helper function for this?

typedData/typedData.go
```diff
@@ -0,0 +1,905 @@
...
+}
+
+// Message returns a copy of the TypedData's message map
+func (td *TypedData) Message() map[string]any {
+	copyMap := make(map[string]any, len(td.message))
+	for k, v := range td.message {
+		copyMap[k] = v
+	}
+	return copyMap
+}
```
rianhughes: This will perform a shallow copy, which is fine for primitive types (ints, string, etc), but not fine for reference types (eg slices, structs that contains slices or pointers etc). At least if you're trying to prevent the user from modifying the td fields.

Why are we preventing the user from modifying these fields?

typedData/revision.go
```diff
@@ -0,0 +1,226 @@
...
+		hashMerkleMethod: curve.Poseidon,
+		types: RevisionTypes{
+			Basic:  append(revision_1_basic_types, revision_0_basic_types...),
+			Preset: presetMap,
+		},
+	}
+}
+
+type revision struct {
+	//TODO: create a enum
```
rianhughes, line 76: Is this still relevant?

typedData/typedData.go
```diff
@@ -0,0 +1,905 @@
...
+
+// Revision returns the revision value of the TypedData
+func (td *TypedData) Revision() revision {
+	return *td.revision
+}
```
rianhughes: I believe this is also performing a shallow copy of a reference type, see the above comment.

typedData/typedData.go
```diff
@@ -0,0 +1,905 @@
...
+// (ref: https://github.com/starknet-io/SNIPs/blob/5d5a42c654c27b377d8b7f90b453065fd19ec2eb/SNIPS/snip-12.md#specification)
+//
+// Parameters:
+// - account: A string representing the account.
+// Returns:
+// - hash: A pointer to a felt.Felt representing the calculated hash.
+func (td *TypedData) GetMessageHash(account string) (hash *felt.Felt, err error) {
+	//signed_data = encode(PREFIX_MESSAGE, Enc[domain_separator], account, Enc[message])
+
+	elements := []*felt.Felt{}
```
rianhughes: nit: we know the size of elements beforehand, so cna pre-define its size. We could also just get rid of it and pass in the paramters directly.

utils/Felt.go
thiagodeev: Because if "" is a valid shortstring, why wouldn't it be a valid byte array? No need to return an error, it's just the zero value of the type.
I realized this in an example where starknet.js was hashing correctly the empty string but our code was returning an error just because it was empty.

typedData/tests/example_array.json
thiagodeev: from starknet.js test files
https://github.com/starknet-io/starknet.js/tree/develop/__mocks__/typedData

typedData/tests/v1Nested.json
thiagodeev: from starknet.js test files
https://github.com/starknet-io/starknet.js/tree/develop/__mocks__/typedData

examples/typedData/main.go
thiagodeev: In the account package, right? I think it's a good idea

typedData/typedData.go
thiagodeev: I did that because when creating a Typed Data the code automatically encodes the types and assigns the result to some parameters; if the user changes any value this will not be reflected to previous computed values and so will lead to a wrong message hash. 
If the user really wants to change the message, just change the original JSON and create a new TypedData from it. As the JSON is the source of truth, there's no advantage in being able to modify a TypedData generated from it since this new TypedData no longer represents the original JSON file.

>but not fine for reference types (eg slices, structs that contains slices or pointers etc). At least if you're trying to prevent the user from modifying the td fields.

Hmm. Any good idea about that?

typedData/revision.go
thiagodeev: I believe it's a good thing to make the code more clear in the future, that's why I left it there. WDYT?

typedData/typedData.go
thiagodeev: Same reason as above, but it also prevents (at least was my intention) the user to change the original `revision` value that is created at the `revision.go` file and is used in the entire code

typedData/typedData.go
thiagodeev: makes total sense haha. I'll do it

utils/Felt.go
```diff
@@ -31,6 +31,16 @@ func HexToFelt(hex string) (*felt.Felt, error) {
...
+// StrToFelt generates a felt.Felt from a string representation containing ASCII characters, not hex values.
+//
+// Parameters:
+// - str: The string to convert to a felt.Felt
+// Returns:
+// - *felt.Felt: a felt.Felt representing the converted value
+func StrToFelt(str string) *felt.Felt {
+	return new(felt.Felt).SetBytes([]byte(str))
+}
+
```
rianhughes: I think this can be deleted as it's not being used. Plus, I feel like it's safer to user SetString when going from string to felt

typedData/revision.go
rianhughes: I don't understand the motivation for having it here, and if it's not going to be addressed in the near future then I lean towards removing it. I'll leave the decision up to you though

typedData/tests/example_array.json
rianhughes: I couldn't find it there, was it removed? Not a big issue though

typedData/typedData.go
rianhughes: We could write logic to perform a deep copy, but imo this is a bit overkill atm. I agree it's fair that the JSON is the source of truth, but imo if the user modifies the message then they should also expect the hash to be modified as well.

However, I think we should be consistent, and either perform a hard copy of each field, or allow the user to modify the types.

typedData/typedData.go
rianhughes: Performing a deep copy of this should be straight forward given we know all the types ahead of time. But we should decide what to do in the above comment first

typedData/typedData_test.go
```diff
@@ -0,0 +1,346 @@
...
+		ExpectedMessageHash string
+	}
+	testSet := []testSetType{
+		{
+			TypedData:           typedDataExamples["baseExample"],
+			Address:             "0xCD2a3d9F938E13CD947Ec05AbC7FE734Df8DD826",
+			ExpectedMessageHash: "0x6fcff244f63e38b9d88b9e3378d44757710d1b244282b435cb472053c8d78d0",
+		},
+		{
+			TypedData:           typedDataExamples["example_array"],
```
rianhughes, line 96: Where did the expected hahes for: example_array, v1Nested, example_enum and allInOne, come from? ( can find the others in starknet.js, but not those)

typedData/typedData_test.go
```diff
@@ -0,0 +1,346 @@
...
+		},
+		{
+			TypedData:    typedDataExamples["example_presetTypes"],
+			TypeName:     "Example",
+			ExpectedHash: "0x1a25a8bb84b761090b1fadaebe762c4b679b0d8883d2bedda695ea340839a55",
+		},
+		{
+			TypedData:    typedDataExamples["session_MerkleTree"],
+			TypeName:     "Session",
+			ExpectedHash: "0x1aa0e1c56b45cf06a54534fa1707c54e520b842feb21d03b7deddb6f1e340c",
```
rianhughes, line 208: where did this come from?

typedData/typedData_test.go
```diff
@@ -0,0 +1,346 @@
...
+		},
+		{
+			TypedData:    typedDataExamples["v1Nested"],
+			TypeName:     "TransferERC721",
+			ExpectedHash: "0x11b5fb80dd88c3d8b6239b065def4ac9a79e6995b117ed5940a3a0734324b79",
+		},
+		{
+			TypedData:    typedDataExamples["example_enum"],
+			TypeName:     "Example",
+			ExpectedHash: "0x1551dc992033e2256a2f7ec849495d90f9759ebb535e3006d16e2b9e3b57b4c",
```
rianhughes, line 337: Where did ExpectedHash come from for these?

typedData/typedData_test.go
```diff
@@ -0,0 +1,346 @@
...
+		},
+		{
+			TypedData:      typedDataExamples["example_baseTypes"],
+			TypeName:       "Example",
+			ExpectedEncode: `"Example"("n0":"felt","n1":"bool","n2":"string","n3":"selector","n4":"u128","n5":"i128","n6":"ContractAddress","n7":"ClassHash","n8":"timestamp","n9":"shortstring")`,
+		},
+		{
+			TypedData:      typedDataExamples["example_presetTypes"],
+			TypeName:       "Example",
+			ExpectedEncode: `"Example"("n0":"TokenAmount","n1":"NftId")"NftId"("collection_address":"ContractAddress","token_id":"u256")"TokenAmount"("token_address":"ContractAddress","amount":"u256")"u256"("low":"u128","high":"u128")`,
```
rianhughes, line 261: If a "non-basic" type is used, does it get explicitly included in the encoding?

typedData/typedData.go
```diff
@@ -0,0 +1,896 @@
...
+
+	chainId, err := getField("chainId")
+	if err != nil {
+		var err2 error
+		// ref: https://community.starknet.io/t/signing-transactions-and-off-chain-messages/66
+		chainId, err2 = getField("chain_id")
+		if err2 != nil {
+			return err
+		}
+	}
```
rianhughes: Maybe something like

value, err1 := getField("chainId")
    if err1 == nil {
        return value, nil
    }

    value, err2 := getField("chain_id")
    if err2 == nil {
        return value, nil
    }

    // Return both errors if both fail
    combinedError := fmt.Errorf("primary error: %v; fallback error: %v", err1, err2)
    return "", combinedError

typedData/tests/example_array.json
thiagodeev: Sorry, you are right!
I take it from the SNIP: https://github.com/starknet-io/SNIPs/blob/main/SNIPS/snip-12.md#json-example

typedData/typedData.go
thiagodeev: It's indeed overkill, but only if the method is called; but why would the user need to call it? Since checking the hashes of types, structs, and the message can be done with the TypedData methods, there's no need for the user to use these values; so it'll not affect the overall code performance. 
IMO, the TypedData after being created should not be able to be modified as it is a representation of something, so we need to take steps to make it not possible for the users, helping them to avoid accidents

typedData/typedData_test.go
thiagodeev: I created a script using starknet.js were I just needed to fill the `typedData.json` file with the JSON that I wanted to check, change some variables and the revision version, and the code would return me the result I wanted. Then, I compared them with the ones returned by starknet.go until they matched.
image

typedData/typedData_test.go
thiagodeev: Explained in my previous comment
https://github.com/NethermindEth/starknet.go/pull/637#discussion_r1879436991

typedData/typedData_test.go
thiagodeev: Explained in my previous comment
https://github.com/NethermindEth/starknet.go/pull/637#discussion_r1879436991

typedData/typedData_test.go
thiagodeev: Yes. Ref: https://github.com/starknet-io/SNIPs/blob/main/SNIPS/snip-12.md#encode_type

typedData/typedData.go
thiagodeev: This was intentional.
The SNIP says the correct is `chainId` but I [found ](https://github.com/starknet-io/SNIPs/issues/122)that previously the `chain_id` name was also used. So it will first look for the 'chainId' field, if not found, it'll then look for the 'chain_id'. If neither 'chainId' or 'chain_id' were found, the error will only say about the `chainId` field since it's the currently used field, the 'chain_id' is verified only for backward compatibility. WDYT?

Now that you have mentioned, I believe it'll be good to before looking for the 'chain_id' field it'll first verify if the Revision version is 0 since the only acceptable field for Revision 1 is 'chainId', so it would be invalid in this case.

typedData/typedData_test.go
thiagodeev: 'allInOne' was created by me

typedData/typedData.go
rianhughes: Yup, I think we should check the revision, and if it's 0 then return both errors, and if it's 1 just check if it's `chainId` (assuming revision 1 only allows chainId)

examples/typedData/main.go
rianhughes: I think the utils pkg might be a good place for it because users might just want to verify a signature that has nothing to do with their own account (or they might not even have an account). Also, I just realised, I think this should verify the signature using the public key, not the private key

thiagodeev: Done. I put the function in the `curve` package because when I tried to put it in `utils` an error was returned about "circle importing"

utils/Felt.go
thiagodeev: Done.
This function was to `StrToFelt generates a felt.Felt from a string representation containing ASCII characters, not hex values.`.  

If you try to put "helloMan!" using SetString it will return `can't parse into a big.Int: helooMann!!`. As you may think, 'SetString()' will first cast it to big.Int; it does that because it is intended to be used with numbers as string, not string characters. Also, even if it succeeds in converting it, the final hex result will differ from the hex representation derived from 'SetBytes()'. SetBytes is the correct one to be used with string representations, but as the users probably will not know about that I created this helper function with the description.

typedData/typedData.go
thiagodeev: Done

typedData/typedData.go
thiagodeev: Done, changed to public fields

rianhughes (approved):

PR #639, opened 2024-10-17
Fix trace/simulation tests for sepolia

wojciechos:
This PR updates the expected test data, which was previously based on invalid responses from Juno. The bug in Juno caused an overestimation leading to incorrect test expectations. The bug has been fixed in [this PR](https://github.com/NethermindEth/juno/pull/2118), and this update corrects the test data to reflect the accurate responses after the fix.

### DIFF
--- a/rpc/tests/trace/sepoliaSimulateInvokeTxResp.json
+++ b/rpc/tests/trace/sepoliaSimulateInvokeTxResp.json
@@ -182,7 +182,7 @@
 					"range_check_builtin_applications": 121,
 					"ec_op_builtin_applications": 3,
 					"data_availability": {
-						"l1_gas": 4668,
+						"l1_gas": 4646,
 						"l1_data_gas": 0
 					}
 				}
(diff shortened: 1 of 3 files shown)


rianhughes (approved):
The Juno fix isn't reflected in the testnet endpoint that starknet.go uses. However, I tested it locally with an endpoint that does contain the Juno fix, and all tests pass.

rpc/tests/trace/sepoliaSimulateInvokeTxResp.json
```diff
@@ -182,7 +182,7 @@
 					"range_check_builtin_applications": 121,
 					"ec_op_builtin_applications": 3,
 					"data_availability": {
-						"l1_gas": 4668,
+						"l1_gas": 4646,
```
derrix060, line 185: looks like it should indeed be 4668: https://github.com/NethermindEth/starknet.go/actions/runs/11381954197/job/31664465265?pr=639#step:8:68

derrix060: oh I just saw Rian's comment now.

In this case we need to update the endpoint that starknet.go is using

wojciechos: The endpoint under test is using a buggy version of Juno that doesn’t yet include the [fix](https://github.com/NethermindEth/juno/pull/2118).

wojciechos: We decided to hold off on merging it until the endpoint is updated with the new Juno version.

rianhughes (approved):

PR #641, opened 2024-10-24
Update expected starknet version to 0.8.0

derrix060:
As of latest main Juno build, we use 0.8.0: https://github.com/NethermindEth/juno/actions/runs/11462050493/job/31893277337#step:5:82


derrix060:
Needs https://github.com/NethermindEth/starknet.go/issues/623 first

PR #642, opened 2024-10-29
starknet_getCompiledCasm

thiagodeev:

Methods
starknet_getCompiledCasm

Components
 CASM_COMPILED_CONTRACT_CLASS
 CASM_ENTRY_POINT
 CellRef
 Deref
 DoubleDeref
 Immediate
 BinOp
 ResOperand
 HINT
 DEPRECATED_HINT
 CORE_HINT
 STARKNET_HINT

Errors
 COMPILATION_ERROR

### DIFF
--- a/rpc/types_executables.go
+++ b/rpc/types_executables.go
@@ -1,9 +1,6 @@
 package rpc
 
 import (
-	"encoding/json"
-	"fmt"
-
 	"github.com/NethermindEth/juno/core/felt"
 )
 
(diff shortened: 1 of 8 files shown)


rpc/types_executables.go
```diff
@@ -0,0 +1,362 @@
+package rpc
+
+import (
+	"github.com/NethermindEth/juno/core/felt"
+)
+
+type CasmCompiledContractClass struct {
+	EntryPointsByType CasmEntryPointsByType `json:"entry_points_by_type"`
+	ByteCode          []*felt.Felt          `json:"bytecode"`
+	Prime             []*felt.Felt          `json:"prime"`
```
rianhughes: Why not NumAsHex for prime?

rpc/types_executables.go
```diff
@@ -0,0 +1,362 @@
...
+	CompilerVersion   string                `json:"compiler_version"`
+	Hints             Hints                 `json:"hints"`
```
rianhughes: I _think_ this should be [2]Hints

rpc/types_executables.go
```diff
@@ -0,0 +1,362 @@
...
+type felt252Dict struct {
+	DictPtr ResOperand `json:"dict_ptr"`
+	Key     ResOperand `json:"key"`
+}
+
+type Felt252DictRead struct {
+	felt252Dict
+	ValueDst CellRef `json:"value_dst"`
+}
+
```
rianhughes, line 133: Why seperate ValueDst from DictPtr and Key?

rianhughes: Ah, because the read and write both use the same types for dict_ptr and key?

utils/data.go
```diff
@@ -16,3 +19,18 @@ func UnwrapJSON(data map[string]interface{}, tag string) (map[string]interface{}
 	}
 	return data, nil
 }
+
+func GetTypedFieldFromJSON[T any](data map[string]interface{}, tag string) (T, error) {
```
rianhughes: Is this used anywhere?

rpc/types_executables.go
```diff
@@ -0,0 +1,362 @@
...
+	HintArr []Hint
+}
+
+func (hints *Hints) Values() (int, []Hint) {
+	return hints.Int, hints.HintArr
+}
+
+func (hints *Hints) Tuple() [2]any {
+	return [2]any{hints.Int, hints.HintArr}
+}
```
rianhughes, line 41: What are these functions for?

rpc/types_executables.go
thiagodeev: Nice catch, thanks!

rpc/types_executables.go
thiagodeev: I did that because `hints` is a [tuple](https://github.com/starkware-libs/starknet-specs/blob/19ab9d4df4ae3acc1a52cde5a43a7cace08bcc4b/api/starknet_executables.json#L92) of [intenger, []hint] and Golang doesn't allow an array of multiple types. So I created a struct having these two fields and two helper methods for getting the values: the `Values()` that returns both values the way Golang usually does (int, []Hint) and the `Tuple()` that returns an `[2]any` array in case someone needs it in this format

rpc/types_executables.go
thiagodeev: Yup, [Felt252DictEntryInit](https://github.com/NethermindEth/starknet.go/pull/642/files#diff-a5d4b8f184d307383cc60d205757b876abb2e7cf9c45a4a49dfb845871fa2910R139) too

utils/data.go
thiagodeev: I was using it in previous commits but not now, I'll remove it then

rpc/types_executables.go
thiagodeev: https://github.com/NethermindEth/starknet.go/pull/642#discussion_r1824423182

Answered here

PR #644, opened 2024-11-06
Thiagodeev/rpcv08 final changes

thiagodeev:

Components
EXECUTION_RESOURCES
COMPUTATION_RESOURCES (remotion)


rpc/errors.go
```diff
@@ -199,3 +257,51 @@ var (
...
+
+	return fmt.Errorf("failed to unmarshal ContractExecutionError")
+}
+
+type ContractExecutionErrorStruct struct {
+	ContractAddress *felt.Felt              `json:"contract_address"`
+	ClassHash       *felt.Felt              `json:"class_hash"`
+	Selector        *felt.Felt              `json:"selector"`
+	Error           *ContractExecutionError `json:"error"`
+}
```
rianhughes, line 307: Are we missing the "error message" field (seems like the spec is missing the property field?), and these should be omitempty since there's a oneOf condition?

rpc/errors.go
```diff
@@ -62,19 +65,74 @@ func tryUnwrapToRPCErr(err error, rpcErrors ...*RPCError) *RPCError {
...
+	return e.Message + ": " + e.Data.Message
+}
+
+type RPCData struct {
+	Message                       string                         `json:",omitempty"`
+	ContractErrorData             *ContractErrorData             `json:",omitempty"`
+	TransactionExecutionErrorData *TransactionExecutionErrorData `json:",omitempty"`
+}
+
+func (rpcData *RPCData) UnmarshalJSON(data []byte) error {
```
rianhughes, line 92: note to self to come back to this

rianhughes: I'm not sure we can/should constrain the `data` field to be `string` or these other concrete types. In practice, I've seen `error`s (message, code, data) returned in the `data` field, so I think we should replace `Message` with `any`, something like


```
type RPCData struct {
	any                           `json:",omitempty"`
	ContractErrorData             *ContractErrorData             `json:",omitempty"`
	TransactionExecutionErrorData *TransactionExecutionErrorData `json:",omitempty"`
}

// Custom unmarshal function for RPCData
func (r *RPCData) UnmarshalJSON(data []byte) error {
	// Initialize a map to hold raw JSON data for conditional decoding
	var rawMap map[string]json.RawMessage
	if err := json.Unmarshal(data, &rawMap); err != nil {
		return err
	}

	// Check and decode ContractErrorData if it exists
	if rawMsg, exists := rawMap["revert_error"]; exists && len(rawMsg) > 0 {
		r.ContractErrorData = &ContractErrorData{}
		return json.Unmarshal(rawMsg, r.ContractErrorData)
	}

	// Check and decode TransactionExecutionErrorData if it exists
	if rawMsg, exists := rawMap["execution_error"]; exists && len(rawMsg) > 0 {
		r.TransactionExecutionErrorData = &TransactionExecutionErrorData{}
		return json.Unmarshal(rawMsg, r.TransactionExecutionErrorData)
	}
	r.any = data
	return nil
}
```

rpc/errors.go
```diff
@@ -62,19 +65,74 @@ func tryUnwrapToRPCErr(err error, rpcErrors ...*RPCError) *RPCError {
...
+		return e.Message
+	}
+	return e.Message + ": " + e.Data.Message
+}
+
+type RPCData struct {
+	Message                       string                         `json:",omitempty"`
+	ContractErrorData             *ContractErrorData             `json:",omitempty"`
+	TransactionExecutionErrorData *TransactionExecutionErrorData `json:",omitempty"`
+}
```
rianhughes: I'm not sure we can/should constrain the data field to be string or these other concrete types. In practice, I've seen errors (message, code, data) returned in the data field, so I think we should replace Message with any, something like
```

type RPCData struct {
	any                           `json:",omitempty"`
	ContractErrorData             *ContractErrorData             `json:",omitempty"`
	TransactionExecutionErrorData *TransactionExecutionErrorData `json:",omitempty"`
}

// Custom unmarshal function for RPCData
func (r *RPCData) UnmarshalJSON(data []byte) error {
	// Initialize a map to hold raw JSON data for conditional decoding
	var rawMap map[string]json.RawMessage
	if err := json.Unmarshal(data, &rawMap); err != nil {
		return err
	}

	// Check and decode ContractErrorData if it exists
	if rawMsg, exists := rawMap["revert_error"]; exists && len(rawMsg) > 0 {
		r.ContractErrorData = &ContractErrorData{}
		return json.Unmarshal(rawMsg, r.ContractErrorData)
	}

	// Check and decode TransactionExecutionErrorData if it exists
	if rawMsg, exists := rawMap["execution_error"]; exists && len(rawMsg) > 0 {
		r.TransactionExecutionErrorData = &TransactionExecutionErrorData{}
		return json.Unmarshal(rawMsg, r.TransactionExecutionErrorData)
	}
	r.any = data
	return nil
}

```

rianhughes (approved):

PR #649, opened 2024-12-19
New rpcdata + RPCv08 errors

thiagodeev:

Components
CONTRACT_EXECUTION_ERROR

Errors
CONTRACT_ERROR
TRANSACTION_EXECUTION_ERROR

Also, it improves the RPCError type. Now we have a way of returning the error message contained in the Data error field. We can also access it just by checking if it's nil or not.

### DIFF
--- a/rpc/errors.go
+++ b/rpc/errors.go
@@ -46,16 +46,16 @@ func Err(code int, data *RPCData) *RPCError {
 // - rpcErrors: variadic list of *RPCError objects to be checked
 // Returns:
 // - error: the original error
-func tryUnwrapToRPCErr(err error, rpcErrors ...*RPCError) *RPCError {
-	errBytes, errIn := json.Marshal(err)
-	if errIn != nil {
-		return Err(InternalError, &RPCData{Message: errIn.Error()})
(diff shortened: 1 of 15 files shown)


rpc/errors.go
```diff
@@ -62,19 +65,75 @@ func tryUnwrapToRPCErr(err error, rpcErrors ...*RPCError) *RPCError {
...
+	return e.Message + ": " + e.Data.Message
+}
+
+type RPCData struct {
+	Message                       string                         `json:",omitempty"`
+	ContractErrorData             *ContractErrorData             `json:",omitempty"`
+	TransactionExecutionErrorData *TransactionExecutionErrorData `json:",omitempty"`
+}
+
+func (rpcData *RPCData) UnmarshalJSON(data []byte) error {
```
rianhughes, line 94: Are we missing the logic to handle the data field in the COMPILATION_ERROR?

rianhughes: Same with MarshalJSON

thiagodeev: Sorry, thanks for noticing! Done

PR #651, opened 2024-12-24
thiagodeev/rpcv08-websocket

thiagodeev:

New methods:
starknet_subscribeNewHeads
starknet_subscribeEvents
starknet_subscribeTransactionStatus
starknet_subscribePendingTransactions
starknet_unsubscribe

Components
NEW_TXN_STATUS
REORG_DATA
SUBSCRIPTION_BLOCK_ID

Errors:
INVALID_SUBSCRIPTION_ID
TOO_MANY_ADDRESSES_IN_FILTER
TOO_MANY_BLOCKS_BACK

In order to implement websocket connection with Starknet in starknet.go I had to make some adaptations to the client code being currently used, and to do that, I had to copy the entire client code.


utils/Felt.go
```diff
@@ -31,6 +31,20 @@ func HexToFelt(hex string) (*felt.Felt, error) {
...
+// Not recommended for production use. Always handle errors correctly.
+//
+// Parameters:
+// - hex: the input hexadecimal string to be converted.
+// Returns:
+// - *felt.Felt: a *felt.Felt object
+func HexToFeltNoErr(hex string) *felt.Felt {
+	felt, _ := new(felt.Felt).SetString(hex)
+	return felt
+}
```
rianhughes: Given this is only used in tests we could do something like this,
```
func HexToFelt(t testing.TB, hex string) *felt.Felt {
	t.Helper()
	f, err := new(felt.Felt).SetString(hex)
	require.NoError(t, err)
	return f
}
```
In general I would advise against ignoring errors in functions like this

rpc/provider.go
```diff
@@ -22,16 +23,26 @@ type Provider struct {
 	chainID string
 }
 
-// NewProvider creates a new rpc Provider instance.
-func NewProvider(url string, options ...ethrpc.ClientOption) (*Provider, error) {
+// WsProvider provides the provider for websocket starknet.go/rpc implementation.
+type WsProvider struct {
+	c wsConn
+}
+
```
rianhughes, line 30: We should probably implement an interface defining the set of functions that the websocket provider must implement, basically something like 
```var _ RpcProvider = &Provider{}```
for the websocket provider

File: account/account_test.go [hunk 3]
```diff
@@ -43,18 +41,12 @@ var (
...
 	}
-	base = os.Getenv("INTEGRATION_BASE")
-	if base == "" {
-		if err := godotenv.Load(fmt.Sprintf(".env.%s", testEnv)); err != nil {
-			panic(fmt.Sprintf("Failed to load .env.%s, err: %s", testEnv, err))
-		}
-		base = os.Getenv("INTEGRATION_BASE")
-	}
+	base = "http://localhost:5050"
 	os.Exit(m.Run())
```
rianhughes: Why change the default testEnv to devnet?

examples/websocket/main.go
```diff
@@ -0,0 +1,48 @@
...
+//	none
+//
+// Returns:
+//
+//	none
+func main() {
+	fmt.Println("Starting simpleCall example")
+
+	// Initialize connection to RPC provider
+	client, err := rpc.NewWebsocketProvider("ws://localhost:6061") //local juno node for testing
```
rianhughes: We should set this as an ENV variable to be consistent with the other examples

examples/websocket/main.go
```diff
@@ -0,0 +1,48 @@
...
+	fmt.Println("Established connection with the client")
+
+	ch := make(chan *rpc.BlockHeader)
+	sub, err := client.SubscribeNewHeads(context.Background(), ch)
+	if err != nil {
+		rpcErr := err.(*rpc.RPCError)
+		panic(fmt.Sprintf("Error subscribing: %s", rpcErr.Error()))
+	}
+
+	for {
```
rianhughes: If there was no error, we should unsubscribe when exiting main,
```defer sub.Unsubscribe()```

File: examples/websocket/README.md [hunk 6]
```diff
@@ -0,0 +1,10 @@
+This example calls two contract functions, with and without calldata. It uses an ERC20 token, but it can be any smart contract.
+
+Steps:
+1. Rename the ".env.template" file located at the root of the "examples" folder to ".env"
+1. Uncomment, and assign your Sepolia testnet endpoint to the `RPC_PROVIDER_URL` variable in the ".env" file
+1. Uncomment, and assign your account address to the `ACCOUNT_ADDRESS` variable in the ".env" file
+1. Make sure you are in the "simpleCall" directory
+1. Execute `go run main.go`
```
rianhughes, line 8: I think this readme needs updated (no need for account address, and we should be in the websocket dir :) )

examples/websocket/README.md
```diff
@@ -0,0 +1,10 @@
...
+
+The calls outuputs will be returned at the end of the execution.
```
rianhughes: I think this README needs updated

rpc/types_event.go
```diff
@@ -45,3 +45,10 @@ type EventsInput struct {
 	EventFilter
 	ResultPageRequest
 }
+
+type EventSubscriptionInput struct {
+	FromAddress *felt.Felt     `json:"from_address,omitempty"` // Optional. Filter events by from_address which emitted the event
+	Keys        [][]*felt.Felt `json:"keys,omitempty"`         // Optional. Per key (by position), designate the possible values to be matched for events to be returned. Empty array designates 'any' value
+	BlockID     BlockID        `json:"block,omitempty"`        // Optional. The block to get notifications from, default is latest, limited to 1024 blocks back
+	// TODO: change 'block' to 'block_id' as soon as Juno fixes the issue with the 'block' field
```
rianhughes: imo we should leave this as block_id for now

rpc/websocket.go
```diff
@@ -0,0 +1,93 @@
...
+// Creates a WebSocket stream which will fire events for new block headers
+//
+// Parameters:
+// - ctx: The context.Context object for controlling the function call
+// - headers: The channel to send the new block headers to
+// - blockID (optional): The block to get notifications from, default is latest, limited to 1024 blocks back
+// Returns:
+// - clientSubscription: The client subscription object, used to unsubscribe from the stream and to get errors
+// - error: An error, if any
+func (provider *WsProvider) SubscribeNewHeads(ctx context.Context, headers chan<- *BlockHeader, blockID ...BlockID) (*client.ClientSubscription, error) {
```
rianhughes: why ```blockID ...BlockID```? Should we not accept a single ```blockID```?

rianhughes:
``` 
"params": [
        {
          "name": "block_id",
          "summary": "The block to get notifications from, default is latest, limited to 1024 blocks back",
          "required": false,
          "schema": {
            "$ref": "./api/starknet_api_openrpc.json#/components/schemas/BLOCK_ID"
          }
        }
      ],
```

rpc/websocket.go
```diff
@@ -0,0 +1,93 @@
...
+
+import (
+	"context"
+
+	"github.com/NethermindEth/juno/core/felt"
+	"github.com/NethermindEth/starknet.go/client"
+)
+
+// New block headers subscription.
+// Creates a WebSocket stream which will fire events for new block headers
```
rianhughes: nit:
// Creates a WebSocket stream which will emit events for new block headers.

rianhughes: Same for the other functions

rpc/websocket.go
```diff
@@ -0,0 +1,93 @@
...
+		sub, err = provider.c.Subscribe(ctx, "starknet", "_subscribeEvents", events, tempInput)
+	} else {
+		sub, err = provider.c.Subscribe(ctx, "starknet", "_subscribeEvents", events, input)
+	}
+
+	if err != nil {
+		return nil, tryUnwrapToRPCErr(err, ErrTooManyKeysInFilter, ErrTooManyBlocksBack, ErrBlockNotFound, ErrCallOnPending)
+	}
+	return sub, nil
+}
```
rianhughes: suggestion
```
func (provider *WsProvider) SubscribeEvents(ctx context.Context, events chan<- *EmittedEvent, input EventSubscriptionInput) (*client.ClientSubscription, error) {
	if input.BlockID.IsZero() {
		input.BlockID = BlockID{Tag: "latest"}
	}
	sub, err := provider.c.Subscribe(ctx, "starknet", "_subscribeEvents", events, input)
	if err != nil {
		return nil, tryUnwrapToRPCErr(err, ErrTooManyKeysInFilter, ErrTooManyBlocksBack, ErrBlockNotFound, ErrCallOnPending)
	}
	return sub, nil
}
```
Also, add the ```IsZero()``` method
```
func (b BlockID) IsZero() bool {
	return b == BlockID{}
}
```
Maybe, also just make the default ```BlockID``` a global varialbe (in ```types_block.go```) and use that instead (probably cleaner)
```
var	DefaultBlockID = BlockID{Tag: "latest"}

```

rpc/websocket.go
```diff
@@ -0,0 +1,93 @@
...
+	sub, err := provider.c.SubscribeWithSliceArgs(ctx, "starknet", "_subscribeTransactionStatus", newStatus, transactionHash, WithBlockTag("latest"))
+	if err != nil {
+		return nil, tryUnwrapToRPCErr(err, ErrTooManyBlocksBack, ErrBlockNotFound)
+	}
+	// TODO: wait for Juno to implement this. This is the correct implementation by the spec
+	// 	sub, err := provider.c.SubscribeWithSliceArgs(ctx, "starknet", "_subscribeTransactionStatus", newStatus, transactionHash)
+	// 	if err != nil {
+	// 		return nil, tryUnwrapToRPCErr(err)
+	// 	}
+	return sub, nil
```
rianhughes: maybe just return a "waiting for Juno to implement this method" error?

rpc/websocket.go
```diff
@@ -0,0 +1,93 @@
...
+//
+// Parameters:
+// - ctx: The context.Context object for controlling the function call
+// - newStatus: The channel to send the new transaction status to
+// - transactionHash: The transaction hash to fetch status updates for
+// Returns:
+// - clientSubscription: The client subscription object, used to unsubscribe from the stream and to get errors
+// - error: An error, if any
+func (provider *WsProvider) SubscribeTransactionStatus(ctx context.Context, newStatus chan<- *NewTxnStatusResp, transactionHash *felt.Felt) (*client.ClientSubscription, error) {
+	sub, err := provider.c.SubscribeWithSliceArgs(ctx, "starknet", "_subscribeTransactionStatus", newStatus, transactionHash, WithBlockTag("latest"))
```
rianhughes: Why do we pass in the BlockID? Also, maybe we can just use a global BlockID as mentioned in the above comment?

utils/Felt.go
thiagodeev: Me too.
Also, for tests there's already the `TestHexToFelt` func that does exactly what you mentioned.

I thought about this function when we need to put some values that we have sure are valid felts (like addresses, hashes...) into (for e.g.) an object, and as the felt (for good reasons) returns an error in such operations we need to declare and set the variable previously. With this function we can set it directly. WDYT?

rpc/provider.go
thiagodeev: Agree.
It is in my plans

account/account_test.go
thiagodeev: Sorry, I always do this while testing but it seems I forgot to revert this time 😅.
I do this to make the VSCode testing extension run tests in the selected environment

examples/websocket/main.go
thiagodeev: Agree.
I'm currently using this file to test each endpoint I'm creating, but I'm not committing the changes.
Once I finish the features I'll make a proper example for websocket

examples/websocket/README.md
thiagodeev: Yes. I'll do this at the end

rpc/types_event.go
thiagodeev: I'm letting this way only to pass the tests.
I'll change it at the end

rpc/websocket.go
thiagodeev: Yes. I was implementing this way because this is the only way to make a parameter really optional in golang. But I agree now that this is counterintuitive, so I'll change it

rpc/websocket.go
thiagodeev: I'm taking descriptions directly from the spec whenever possible. This came from the spec: [ref](https://github.com/starkware-libs/starknet-specs/blob/v0.8.0/api/starknet_ws_api.json#L12).
WDYT?

rpc/websocket.go
thiagodeev: Yeah, I know it defaults to "latest" even if we don't pass any blockID, but a json request with an empty param is different from a request passing values. The json will be different from the one requested by the user, even if the result is the same.
I just wanted to make the final json request exactly the one the user is creating in the code. WDYT?

rpc/websocket.go
thiagodeev: Because it'll be easier for me at the end to just uncomment the correct code and delete the other hahaha.
Hope you don't mind. I'll fix it anyway at the end

rpc/websocket.go
thiagodeev: This field is no longer accepted by this method according to the spec, but Juno hasn't implemented it yet when I wrote this. So I made this change to make it work normally and test it, but I'll revert it later

rpc/websocket.go
rianhughes: Yeah I seen that, idk, personally I prefer `emit events`, but I'll leave it up to you

rpc/websocket.go
rianhughes: Yeah that's a good point. Although, given an absent `blockId` is the same as the default value, I think overriding it is okay. Plus this approach is a bit neater, and less error prone imo

thiagodeev: Please take a look at my new approach and let me know your opinion

client/subscription.go
```diff
@@ -255,6 +265,12 @@ func (sub *ClientSubscription) Err() <-chan error {
 	return sub.err
 }
 
+// Reorg returns a channel that notifies the subscriber of a reorganization of the chain.
+// A reorg event could be received only from subscribing to NewHeads, Events, and TransactionStatus
+func (sub *ClientSubscription) Reorg() <-chan *ReorgEvent {
```
thiagodeev, line 270: The Reorg notification was tested locally using a Juno node with a small modification.
To simulate reorg notifications, I changed the Juno code so that instead of waiting for the feeder to return the Reorg event, it'll send one every 5 seconds, along with the main subscription.

So, I subscribed to the `SubscribeNewHeads` method and the code handled perfectly both the main subscription and the ReorgEvents sents.
<table>
<tr>
    <td></td>
    <td></td>
</tr>
</table>

examples/websocket/README.md
```diff
@@ -0,0 +1,9 @@
+This example demonstrates how to subscribe to new block headers using WebSocket. It can be adapted to subscribe to other methods as well.
+
+Steps:
+1. Rename the ".env.template" file located at the root of the "examples" folder to ".env"
+1. Uncomment, and assign your Sepolia WebSocket testnet endpoint to the `WS_PROVIDER_URL` variable in the ".env" file
+1. Make sure you are in the "websocket" directory
+1. Execute `go run main.go`
+
+The calls outuputs will be returned at the end of the execution.
```
rianhughes: typo

examples/websocket/README.md
```diff
@@ -0,0 +1,9 @@
+This example demonstrates how to subscribe to new block headers using WebSocket. It can be adapted to subscribe to other methods as well.
+
+Steps:
+1. Rename the ".env.template" file located at the root of the "examples" folder to ".env"
+1. Uncomment, and assign your Sepolia WebSocket testnet endpoint to the `WS_PROVIDER_URL` variable in the ".env" file
```
rianhughes, line 5: This variable doesn't exist in the .env file

examples/websocket/main.go
```diff
@@ -0,0 +1,102 @@
...
+			break loop1 // Let's exit the loop after receiving the first block header
+		case err := <-sub.Err():
+			// This case will be triggered when an error occurs.
+			setup.PanicRPC(err)
+		}
+	}
+
+	// We can also use the subscription returned by the WS methods to unsubscribe from the stream when we're done
+	sub.Unsubscribe()
+
```
rianhughes, line 64: Imo this should always be executed if we had no error subscribing in the first place, so I would suggest using defer sub.Unsubscribe() between line 40 and 41

examples/websocket/main.go
```diff
@@ -0,0 +1,102 @@
...
+		setup.PanicRPC(err)
+	}
+	fmt.Println()
+	fmt.Println("Successfully subscribed to the node. Subscription ID:", sub.ID())
+
+	go func() {
+		time.Sleep(20 * time.Second)
+		// Unsubscribe from the subscription after 20 seconds
+		sub.Unsubscribe()
+	}()
```
rianhughes, line 82: similarly with this (ie just use a defer statement)

examples/websocket/main.go
rianhughes: Also, I think we can simplify this example by removing the code below this line, and explicitly exit after we receive eg 5 blocks form the ws endpoint

utils/Felt.go
rianhughes: Currently this function is only used in tests, and should be replaced with `TestHexToFelt`. 

I see the appeal of having some method to create felts that we know are valid, but imo this should be pushed to a separate PR for now

rpc/websocket.go
rianhughes: lgtm

rianhughes: Although, we seem to repeat the same logic multiple times in different ways, I feel like we should unify it.
Eg in SubscribeEvents we do
```
var emptyBlockID BlockID
	if options.BlockID == emptyBlockID {
		options.BlockID = WithBlockTag("latest")
	}
```

but in SubscribeNewHeads we have 
```
if blockID == nil {
		blockID = &BlockID{Tag: "latest"}
	}
```
and again in SubscribeTransactionStatus

rpc/websocket.go
rianhughes: Juno now supports this method, so it should be updated

rpc/client.go
```diff
@@ -4,14 +4,20 @@ import (
...
 type callCloser interface {
 	CallContext(ctx context.Context, result interface{}, method string, args ...interface{}) error
 	Close()
 }
 
+type wsConn interface {
+	callCloser
+	Subscribe(ctx context.Context, namespace string, methodSuffix string, channel interface{}, args interface{}) (*client.ClientSubscription, error)
+	SubscribeWithSliceArgs(ctx context.Context, namespace string, methodSuffix string, channel interface{}, args ...interface{}) (*client.ClientSubscription, error)
+}
```
rianhughes, line 19: Do we need SubscribeWithSliceArgs? Can we not just update Subscribe to be
Subscribe(ctx context.Context, namespace string, methodSuffix string, channel interface{}, args ...interface{})

a quick check suggests it works

rpc/websocket_test.go
```diff
@@ -0,0 +1,543 @@
...
+			require.NoError(t, err)
+			defer wsProvider.Close()
+
+			var sub *client.ClientSubscription
+			if test.blockID == nil {
+				sub, err = wsProvider.SubscribeNewHeads(context.Background(), test.headers, nil)
+			} else {
+				sub, err = wsProvider.SubscribeNewHeads(context.Background(), test.headers, test.blockID)
+			}
+
```
rianhughes: I think this can just be simpified by just setting blockId to nil in the testcase

```
{ // normal
				headers:         make(chan *BlockHeader),
				blockID:         nil,
				isErrorExpected: false,
			}
```

rpc/websocket_test.go
```diff
@@ -0,0 +1,543 @@
...
+			if test.isErrorExpected {
+				require.Error(t, err)
+				return
+			} else {
+				require.NoError(t, err)
+			}
```
rianhughes: We should also be able to simplify this, while explicitly testing we get the expected error
eg 
```
if test.ExepctedError !=nil{
    require.EqualError(t, err, test.ExpectedError)
	return
}
require.NoError(t, err)
```

rpc/websocket_test.go
```diff
@@ -0,0 +1,543 @@
...
+
+			require.NotNil(t, sub)
+			defer sub.Unsubscribe()
```
rianhughes: suggestion, we should move the defer above line 68 and do something like this
if sub!=nil{
    defer sub.Unsubscribe()
}

rpc/websocket.go
thiagodeev: But it's still outdated. It needs to remove the "block_id" param.
Juno code: https://github.com/NethermindEth/juno/blob/main/rpc/subscriptions.go#L111
Spec: https://github.com/starkware-libs/starknet-specs/blob/v0.8.0/api/starknet_ws_api.json#L134

rpc/websocket_test.go
```diff
@@ -0,0 +1,543 @@
...
+			{ // invalid, with block number out of the range of 1024 blocks
+				headers:         make(chan *BlockHeader),
+				blockID:         &blockIdEx2,
+				isErrorExpected: true,
+			},
+		},
+	}[testEnv]
+
+	for index, test := range testSet {
+		t.Run(fmt.Sprintf("test %d", index+1), func(t *testing.T) {
```
rianhughes: Maybe we could add a "description" field to the testSet, and just print that? It would really help with understanding which tests are running / failing etc

rpc/websocket_test.go
```diff
@@ -0,0 +1,543 @@
...
+	fromAddress := utils.HexToFeltNoErr("0x049d36570d4e46f48e99674bd3fcc84644ddd6b96f7c741b1562b82f9e004dc7") // sepolia StarkGate: ETH Token
+	key := utils.HexToFeltNoErr("0x99cd8bde557814842a3121e8ddfd433a539b8c9f14bf31ebf108d12e6196e9")
+
+	t.Run("normal call, with empty args", func(t *testing.T) {
+		t.Parallel()
+
+		wsProvider, err := NewWebsocketProvider(testConfig.wsBase)
+		require.NoError(t, err)
+		defer wsProvider.Close()
+
```
rianhughes: We should be able to reuse the same wsProvider for all tests, no need to keep setting up new websockets

rpc/websocket_test.go
```diff
@@ -0,0 +1,543 @@
...
+				if !differentKeyFound {
+					if !slices.Contains(resp.Keys, key) {
+						differentKeyFound = true
+					}
+				}
+
+				if differentFromAddressFound && differentKeyFound {
+					return
+				}
```
rianhughes: What are you trying to check for here??

rpc/websocket_test.go
```diff
@@ -0,0 +1,543 @@
...
+		defer sub.Unsubscribe()
+
+		for {
+			select {
+			case resp := <-events:
+				require.IsType(t, &EmittedEvent{}, resp)
+				require.Less(t, resp.BlockNumber, blockNumber)
+				require.Equal(t, fromAddress, resp.FromAddress)
+				require.Equal(t, key, resp.Keys[0])
+				return
```
rianhughes: Lets leave a note that this assumes the block contains such an event, and why it's likely to

rpc/websocket_test.go
```diff
@@ -0,0 +1,543 @@
...
+			},
+			{
+				input: EventSubscriptionInput{
+					BlockID: WithBlockNumber(blockNumber - 1025),
+				},
+				expectedError: ErrTooManyBlocksBack,
+			},
+			{
+				input: EventSubscriptionInput{
+					BlockID: WithBlockNumber(blockNumber + 2),
```
rianhughes: Let's bump this for safety, for when SN gets super low block times :)

rpc/websocket_test.go
rianhughes: Same for TestSubscribePendingTransactions

rpc/websocket_test.go
```diff
@@ -0,0 +1,543 @@
...
+	}[testEnv]
+
+	for index, test := range testSet {
+		t.Run(fmt.Sprintf("test %d", index+1), func(t *testing.T) {
+			t.Parallel()
+
+			wsProvider, err := NewWebsocketProvider(testConfig.wsBase)
+			require.NoError(t, err)
+			defer wsProvider.Close()
+
```
rianhughes: We can move this outside of the loop

rpc/websocket_test.go
```diff
@@ -0,0 +1,543 @@
...
+			sub, err := wsProvider.SubscribePendingTransactions(context.Background(), test.pendingTxns, test.options)
+
+			if test.expectedError != nil {
+				require.Error(t, err)
+				return
+			} else {
+				require.NoError(t, err)
+			}
```
rianhughes: I wonder if we could just do something like
```
require.EqualError(t,err,test.ExpectedErr)
if test.expectedError != nil { return}
```

rpc/websocket_test.go
rianhughes: similar for TestSubscribePendingTransactions

rpc/websocket_test.go
```diff
@@ -0,0 +1,543 @@
...
+			t.Fatal("timeout waiting for unsubscription")
+		}
+	}
+
+	// Unsubscribe again to make sure nothing happens
+	sub.Unsubscribe()
+}
+
+// TODO: Add mock for testing reorg events.
+// A simple test was made to make sure the reorg events are received; it'll be added in the PR 651 comments
```
rianhughes: Note for myself to revisit this

rpc/websocket.go
thiagodeev: The spec was updated. Take a look at my new approach

rpc/websocket_test.go
thiagodeev: Did you see my comment?

https://github.com/NethermindEth/starknet.go/pull/651#discussion_r1927356194

rpc/provider.go
```diff
@@ -76,4 +106,12 @@ type RpcProvider interface {
 	TraceTransaction(ctx context.Context, transactionHash *felt.Felt) (TxnTrace, error)
 }
 
+type WebsocketProvider interface {
+	SubscribeEvents(ctx context.Context, events chan<- *EmittedEvent, options *EventSubscriptionInput) (*client.ClientSubscription, error)
+	SubscribeNewHeads(ctx context.Context, headers chan<- *BlockHeader, subBlockID *SubscriptionBlockID) (*client.ClientSubscription, error)
+	SubscribePendingTransactions(ctx context.Context, pendingTxns chan<- *SubPendingTxns, options *SubPendingTxnsInput) (*client.ClientSubscription, error)
+	SubscribeTransactionStatus(ctx context.Context, newStatus chan<- *NewTxnStatusResp, transactionHash *felt.Felt) (*client.ClientSubscription, error)
+}
```
rianhughes, line 114: Maybe we should add some mocks for the WebsocketProvider, although I would suggest doing this in a separate PR

rianhughes: It would allow us to run mock tests

examples/websocket/main.go
thiagodeev: But each block takes ~15-20 seconds to build, so they'll need to wait almost 2 minutes for it.
We can indeed simplify it, but IMO the example should be both simple and complete. My intuition for including the last code part was to show the effects of the parameters and how to use them.
WDYT?

client/subscription.go
rianhughes: Nice, good to see that it works! However, given the importance of this feature (thinking about any external service that might depend on this), I need to think about how we could test this more regularly. Saying that, it doesn't seem like the most straight forward thing to test..

rpc/websocket_test.go
thiagodeev: Unfortunately not with 'EqualError', because it '...asserts that a function returned an error (i.e. not `nil`)...' first.
I think your previous suggestion of removing the 'else' is enough

rpc/websocket_test.go
thiagodeev: I changed it a little bit. It's about testing whether events from different addresses and keys are being returned, as this testcase doesn't have any filter

rpc/websocket.go
rianhughes: Yup, good point. I've created a PR for it here https://github.com/NethermindEth/juno/pull/2410

rpc/client.go
thiagodeev: The answer is that the client was passing the args as an array by default, and I was facing an issue with Juno related to optional parameters (it is not accepting to omit any parameters when passing as an array, even if they're optional), so I adapted the "Subscribe" function to send "any" instead of "[]any", and created another method, "SubscribeWithSliceArgs", to maintain compatibility with the remaining codebase and tests.
So when we need to have optional parameters, we need to use "Subscribe", but when all of them are needed, instead of creating a struct for it, we can just send them through "SubscribeWithSliceArgs"

client/subscription.go
thiagodeev: Indeed hahaha, I also can't figure out how we could test it.
Maybe with a mock test it's possible, sending a specific JSON in the request, but with testnet I believe it's not possible.

rpc/client.go
rianhughes: I think the issue might have been related to not specifying the parameter names in the request, eg updating the `SubscribeTransactionStatus` method as follows seems to work

```
func (provider *WsProvider) SubscribeTransactionStatus(ctx context.Context, newStatus chan<- *NewTxnStatusResp, transactionHash QWEQWE) (*client.ClientSubscription, error) {
    sub, err := provider.c.Subscribe(ctx, "starknet", "_subscribeTransactionStatus", newStatus, transactionHash)
    if err != nil {
        return nil, tryUnwrapToRPCErr(err, ErrTooManyBlocksBack, ErrBlockNotFound)
    }
    return sub, nil
}

type QWEQWE struct {
    TransactionHash *felt.Felt `json:"transaction_hash"`
}
```
This explicitly sets the parameter names in the request, eg like 

`{"jsonrpc":"2.0","id":1,"method":"starknet_subscribeTransactionStatus","params":{"transaction_hash":"0x7b"}}`

instead of something like this
`{"jsonrpc":"2.0","id":1,"method":"starknet_subscribeTransactionStatus","params":["0x7b","0x7b"]}`

when using `SubscribeWithSliceArgs`

rianhughes: The `optional` json tag should be able to handle the case where the parameter is optional as well

thiagodeev: Yes, you're right. But the current solution for the problem is:
- `Subscribe` was changed to receive `any` instead of `[]any`
- A new method called `SubscribeWithSliceArgs` was created to perfectly replace the older version of the `Subscribe` method where it was used. This method simply takes an `[]any` and passes it to `Subscribe`.
It's simple and works well.

But if we remove the `SubscribeWithSliceArgs`, we'll need to wrap single values params in a struct only to make it work, we'll also need to create new structs inside the tests that came from ethrpc only to make it work. I don't see any advantages in doing this, only disadvantages.

WDYT?

rpc/provider.go
thiagodeev: Do you want me to fix this right after this RP was merged?

rianhughes: Yup sure, this PR is big enough

examples/websocket/main.go
rianhughes: > My intuition for including the last code part was to show the effects of the parameters and how to use them.

Yeah good point

rpc/client.go
rianhughes: Maybe I'm missing something here, but it seems that the question comes down to, do we want to:

1. Keep the `SubscribeWithSliceArgs`, which saves us from creating a new type `SubscribeTransactionStatusInput`, but requires that full-nodes handle the array case (which they should), or
2. Remove `SubscribeWithSliceArgs`, using `Subscribe` instead, and create a new type `SubscribeTransactionStatusInput` that just holds a transaction hash?

Imo the second solution is a bit cleaner, but I'll leave it up to you :)

rpc/websocket_test.go
rianhughes: Let's add a `t.Skip("To implement reorg test")` and come back to it in a separate PR

rpc/websocket_test.go
rianhughes: Why don't we check error equality when we expect an error? 

```
if test.expectedError != nil {
				require.EqualError(t, err,test.expectedError.Error())
				return
			}
```

rpc/websocket_test.go
rianhughes: lgtm

rpc/client.go
thiagodeev: Thanks, Rian! I'll keep it as is then.
That way I don't need to adapt the 'go-ethereum' tests and create new types for future optional values.

rpc/websocket_test.go
thiagodeev: Good point! Done

rpc/provider_test.go
```diff
@@ -92,19 +95,38 @@ func beforeEach(t *testing.T) *testConfiguration {
...
+	}
+	wsClient, err := NewWebsocketProvider(testConfig.wsBase)
+	if err != nil {
+		t.Fatal("connect should succeed, instead:", err)
+	}
+	testConfig.wsProvider = wsClient
+	t.Cleanup(func() {
+		testConfig.wsProvider.c.Close()
+	})
+
```
rianhughes: Currently this breaks testnet tests that don't depend on websocket connections. I think we should do something like `beforeEachHTTP`, and `beforeEachWS`. 

Also, I think we should update the error string in `t.Fatal` to be more descriptive, eg `"failed to make ws connection with client"` or something. Currently we just get `connect should succeed, instead: no known transport for URL scheme ""`, which isn't super clear.

.github/workflows/main_ci_check.yml
```diff
@@ -43,22 +43,24 @@ jobs:
         run: cd rpc && go test -timeout 1200s -v -env testnet .
         env:
           TESTNET_ACCOUNT_PRIVATE_KEY: ${{ secrets.TESTNET_ACCOUNT_PRIVATE_KEY }}
-          INTEGRATION_BASE: "https://free-rpc.nethermind.io/sepolia-juno"
+          HTTP_PROVIDER_URL: "https://free-rpc.nethermind.io/sepolia-juno"
+          WS_PROVIDER_URL: "wss://ethereum-rpc.publicnode.com" # TODO: change to a Starknet ws node before merge
```
rianhughes: Why are we connecting to an ethereum node?? Given there are no free rpc nodes available we should either 1) just not run ws tests in workflows, 2) connect to (eg) an internal Juno node that supports ws, without exposing the endpoint publically (@wojciechos knows how to do this)

.github/workflows/test_rpc.yml
```diff
@@ -46,12 +46,14 @@ jobs:
         #run: cd rpc && go test -timeout 1200s -v -env testnet .
         env:
           TESTNET_ACCOUNT_PRIVATE_KEY: ${{ secrets.TESTNET_ACCOUNT_PRIVATE_KEY }}
-          INTEGRATION_BASE: "https://free-rpc.nethermind.io/sepolia-juno"
+          HTTP_PROVIDER_URL: "https://free-rpc.nethermind.io/sepolia-juno"
+          WS_PROVIDER_URL: "wss://ethereum-rpc.publicnode.com" # TODO: change to a Starknet ws node before merge
```
rianhughes: Also marking this for myself

.github/workflows/main_ci_check.yml
rianhughes: I created a secret in the Github, I think we can change this to
`WS_PROVIDER_URL: ${{ secrets.WS_PROVIDER_URL }}`

rpc/provider_test.go
thiagodeev: Done

.github/workflows/main_ci_check.yml
thiagodeev: Fixed

rpc/provider_test.go
```diff
@@ -76,35 +79,59 @@ func TestMain(m *testing.M) {
...
 // - *testConfiguration: a pointer to the testConfiguration struct
-func beforeEach(t *testing.T) *testConfiguration {
+func beforeEach(t *testing.T, isWs bool) *testConfiguration {
 	t.Helper()
 	_ = godotenv.Load(fmt.Sprintf(".env.%s", testEnv), ".env")
 	testConfig, ok := testConfigurations[testEnv]
 	if !ok {
-		t.Fatal("env supports mock, testnet, mainnet, devnet, integration")
+		t.Fatal("env supports mock, testnet, wsTestnet, mainnet, devnet, integration")
 	}
```
rianhughes, line 91: I guess we'll have a wsMainnet case when mainnet supports 08?

thiagodeev: Actually, I just forgot to remove this.
The responsible for checking if it's a websocket test is `isWs` now, it doesn't matter if it's testnet or mainnet

thiagodeev: As the entire code base is being updated to v08, it requires a ws_endpoint to not fail in ws tests.
The `isWs` variable is to make the test not fail if it doesn't require a ws_endpoint (like mocks, and HTTP methods) and it's not provided

rianhughes (approved):

PR #656, opened 2025-01-17
Fix contract class unmarshall for modern cairo compilers

### LINKED ISSUE #655: Support for modern CASM format
Problem statement:

I'm trying to deploy [contract](https://gist.github.com/baitcode/10485b681936078dca01f5cb839992b5) built with `scarb >= 2.7.0` to sepolia network using recipe from [here](https://github.com/NethermindEth/starknet.go/blob/45554f0f2f672a586bc2ca0e46a8c0cbff0ef99e/account/account_test.go#L1109) (kindly provided by @thiagodeev). But all I got was error from the node, stating:
```json
{
     "jsonrpc": "2.0",
     "id": 3,
     "error": {
        "code": 60,
        "message": "The compiled class hash did not match the one supplied in the transaction"
    }
}
```
As I need this functionality in golang I started digging and realised that library has fallen behind agains the recent casm format, especially it lacks accounting of `bytecode_segment_lengths` field. That results in incorrect class hash.

baitcode:
Fixes: #655 

Added workaround for abi data that is not string, but json object.
* added contract casm compiled by scarn 2.7.0 compiler
* added tests to unmarshall contract class from new casm json format

Fix ClassHash calculation for casm compiled classes for modern cairo compilers
* Added bytecode_segment_lengths field to CasmClass, allowing for backward compatibility
* Added NestedUInts type to handle deserialisation of CasmClass
* Added test to check that ClassHash now respects bytecode_segment_lengths and calculates correct value

Here is the snippet for hash values generation with starknet.py:

https://gist.github.com/baitcode/5adec0f74121d0ae3a09aaf573ff8665


.tool-versions
thiagodeev, line 1: Let's remove this file

account/account.go
```diff
@@ -413,11 +412,12 @@ func (account *Account) TransactionHashDeclare(tx rpc.DeclareTxnType) (*felt.Fel
 		if err != nil {
 			return nil, err
 		}
+
 		return hash.CalculateTransactionHashCommon(
 			PREFIX_DECLARE,
 			txnVersionFelt,
 			txn.SenderAddress,
-			&felt.Zero,
+			utils.Uint64ToFelt(0),
```
thiagodeev: We need to make tests to check the accuracy of this hash, taking examples and results from trusted sources

contracts/contracts_test.go
```diff
@@ -33,6 +33,29 @@ func TestUnmarshalContractClass(t *testing.T) {
...
+//
+// It reads the content of a file and unmarshals it into a ContractClass struct.
+// Then it asserts the equality of certain values within the struct.
+//
+// Parameters:
+// - t: The testing.T instance for running the test
+// Returns:
+//
+//	none
+func TestUnmarshalContractClassWithNonStringAbi(t *testing.T) {
```
thiagodeev: No need to a new test, let's just include this test case in the "TestUnmarshalContractClass" test and leave a comment saying it's a "...contract class made with compiler >= 2.7.0 version..."

hash/hash_test.go
```diff
@@ -77,3 +77,27 @@ func TestCompiledClassHash(t *testing.T) {
...
+// It reads the contents of a file, unmarshals it into a CasmClass object, computes the hash using the CompiledClassHash function,
+// and asserts that the computed hash matches the expected hash.
+//
+// Parameters:
+// - t: A testing.T object used for running the test and reporting any failures.
+// Returns:
+//
+//	none
+func TestCompiledClassHashWithBytecodeSegmentLengths(t *testing.T) {
+	expectedHash := "0x28e4f13aed1aea3b4c2544bc10b0a2a4f37c71f830d0e3031e95aeb9cef9889"
```
thiagodeev: Where did you get that hash from?

hash/hash_test.go
thiagodeev, line 1: We need to test outputs from the new implementation of the `CompiledClassHash`. Let's take examples from trusted sources
Ex: [starknet.py](https://github.com/software-mansion/starknet.py/blob/development/starknet_py/tests/unit/hash/casm_class_hash_test.py), [cairo-lang](https://github.com/starkware-libs/cairo-lang/blob/v0.13.1/src/starkware/starknet/core/os/contract_class/compiled_class_hash_test.py)

Or we could make new contracts and use these libraries to get the hash, and then compare them with the ones generated by starknet.go

thiagodeev (changes requested):
Thanks for your help, @baitcode! 
I left some comments, mostly about testing

hash/hash_test.go
baitcode: I created contract myself and calculated hash using starknet.py library

hash/hash_test.go
baitcode: I used starknet.py on a contract I've created myself. I can use argent contracts for example.

account/account.go
baitcode: I don't quite understand. I can assure that I created contract myself, compiled it, calculated hash using `starknet.py` and it didn't match.

baitcode:
@thiagodeev Fixed!

Added argent account contract v0.4.0 to hash tests.

account/account.go
thiagodeev: What I meant was that we needed to test this change, but I had forgotten about existing tests like 'TestTransactionHashDeclare'. Sorry.

But about this change, the `felt.Zero` indeed is equal to 0 as felt. I ran 'TestTransactionHashDeclare' here and it worked the same for both '&felt.Zero' and 'utils.Uint64ToFelt(0)'.
What could have happened is that you changed the value of `felt.Zero` at some point. `felt.Zero` is just a pointer to a felt var containing the zero value, and it's used across the entire code as a reference. If you need a zero felt just do `new(felt.Felt)` instead.

thiagodeev: In the func `getByteCodeSegmentHasher`, at line 198, you change the value of `felt.Zero`. Change it to `new(felt.Felt)` instead

hash/hash_test.go
thiagodeev: Great. No need for the argent contract then, please remove it.
Please send here evidence of your test with starknet.py (screenshot, the code itself,...etc)

baitcode: https://gist.github.com/baitcode/5adec0f74121d0ae3a09aaf573ff8665

rpc/types_contract.go
```diff
@@ -71,7 +89,7 @@ type ContractClass struct {
 
 	EntryPointsByType EntryPointsByType `json:"entry_points_by_type"`
 
-	ABI string `json:"abi,omitempty"`
+	ABI NestedString `json:"abi,omitempty"`
```
thiagodeev, line 107: Since this new type was added to fix the incompatibility with the new ABI being an array, it's good to improve the `TestClassHash` test to have a test case testing the `ClassHash` function with an updated sierra json file that has the ABI as array.

And about testing, please follow [this](https://github.com/NethermindEth/starknet.go/blob/main/rpc/contract_test.go#L352) standard when it contains at least two test cases

hash/hash_test.go
```diff
@@ -77,3 +77,25 @@ func TestCompiledClassHash(t *testing.T) {
...
+//
+// It reads the contents of a file, unmarshals it into a CasmClass object, computes the hash using the CompiledClassHash function,
+// and asserts that the computed hash matches the expected hash.
+//
+// Parameters:
+// - t: A testing.T object used for running the test and reporting any failures.
+// Returns:
+//
+//	none
+func TestCompiledClassHashWithBytecodeSegmentLengths(t *testing.T) {
```
thiagodeev: This test does exactly the same thing as `TestCompiledClassHash`. Let's put the two together and just make a comment explaining each scenario.

And please, follow [this](https://github.com/NethermindEth/starknet.go/blob/main/rpc/contract_test.go#L352) pattern

contracts/contracts_test.go
```diff
@@ -23,14 +23,29 @@ import (
 //
 //	none
 func TestUnmarshalContractClass(t *testing.T) {
-	content, err := os.ReadFile("./tests/hello_starknet_compiled.sierra.json")
-	require.NoError(t, err)
+	{
```
thiagodeev: Try to follow [this](https://github.com/NethermindEth/starknet.go/blob/main/rpc/contract_test.go#L352) pattern please

contracts/contracts.go
```diff
@@ -10,12 +11,74 @@ import (
...
-	Prime            string                     `json:"prime"`
-	Version          string                     `json:"compiler_version"`
-	ByteCode         []*felt.Felt               `json:"bytecode"`
-	EntryPointByType CasmClassEntryPointsByType `json:"entry_points_by_type"`
-	// Hints            any                        `json:"hints"`
+	Prime                  string                     `json:"prime"`
+	Version                string                     `json:"compiler_version"`
+	ByteCode               []*felt.Felt               `json:"bytecode"`
+	EntryPointByType       CasmClassEntryPointsByType `json:"entry_points_by_type"`
+	BytecodeSegmentLengths *NestedUInts               `json:"bytecode_segment_lengths,omitempty"`
```
thiagodeev: Since this new field was added, let's add a test case in the `TestUnmarshalCasmClass` to test the correct unmarshal of a casm.json file containing this new field

thiagodeev (changes requested):
Thanks @baitcode. Some new observations about testing

account/account.go
thiagodeev: Hey @baitcode, just to remind you of my answer here

baitcode: Oh my god! Thank you for catching that. Releasing that error would be a disaster!

baitcode:
@thiagodeev Thank you very much for the review. I've fixed all the comments.

contracts/contracts_test.go
```diff
@@ -60,13 +67,111 @@ func TestUnmarshalContractClass(t *testing.T) {
...
+	for _, testCase := range testCases {
+		casmClass, err := UnmarshalCasmClass(testCase.CasmPath)
+		require.NoError(t, err)
+
+		assert.Equal(t, casmClass.Prime, testCase.ExpectedCasmClass.Prime)
+		assert.Equal(t, casmClass.Version, testCase.ExpectedCasmClass.Version)
+
+		expectedEntryPoint := testCase.ExpectedCasmClass.EntryPointByType
+
+		// External
```
thiagodeev: We can just use:

```
assert.Equal(t, casmClass.EntryPointByType.External, expectedEntryPoint.External)
assert.Equal(t, casmClass.EntryPointByType.Constructor, expectedEntryPoint.Constructor)
assert.Equal(t, casmClass.EntryPointByType.L1Handler, expectedEntryPoint.L1Handler)
```

hash/hash.go
```diff
@@ -195,7 +195,9 @@ func getByteCodeSegmentHasher(
 			components[i*2+1] = val.Hash()
 		}
 
-		return felt.Zero.Add(
+		zero := &felt.Felt{}
+
+		return zero.Add(
```
thiagodeev: We don't need a new var, we can just pass `new(felt.Felt)` directly

contracts/contracts.go
thiagodeev, line 1: Since these functions are only being used in the `TestUnmarshalCasmClass` test, let's move them to the `contracts_test.go` file

thiagodeev (changes requested):
Thanks @baitcode! Nice work.
Just a few requests:

rpc/types_contract.go
thiagodeev: @baitcode , let's not forget this

contracts/contracts.go
thiagodeev: I'm talking about the `NewNestedFieldValue` and `NewNestedFieldValue` functions

contracts/contracts_test.go
baitcode: I can't remember why exactly I decided to compare everything explicitly so will have to fix. Will leave a comment next time.

hash/hash.go
baitcode: Thanks for catching!

contracts/contracts.go
baitcode: >> I'm talking about the NewNestedFieldValue and NewNestedFieldValue functions
Thank you for clarifying, firstly I though about `ClassHash` and `CasmClassHash` functions, was up to oppose that message. 

Moved.

hash/hash_test.go
```diff
@@ -21,59 +22,118 @@ import (
...
 //
 // Parameters:
-// - t: A testing.T object used for running the test and reporting any failures.
-// Returns:
-//
-//	none
-func TestCompiledClassHash(t *testing.T) {
-	//https://github.com/software-mansion/starknet.py/blob/development/starknet_py/hash/casm_class_hash_test.py
-	expectedHash := "0x785fa5f2bacf0bfe3bc413be5820a61e1ea63f2ec27ef00331ee9f46ad07603"
+func TestClassHashes(t *testing.T) {
```
thiagodeev, line 64: I made this test using the data from starknet.py (the refs are inside the test).
The CompiledClass hashes are passing, but the Sierra Class hashes are failing. I ran each of these jsons with a starknet.py script and they passed, so it's starknet.go fault.
Your "contract.sierra.json" was also having a different hash than when I tested with starknet.py.

I noticed that "hello_starknet_compiled" is working fine. Since it's the only one with the "abi" field as a string, I think that might be involved since all the others have the "abi" field as an array.

So, we need to fix our ClassHash func. Could you please do that?

thiagodeev (changes requested):
Thank you again @baitcode!

I made a new test for the `ClassHash` and `CompiledClassHash`. The code is failing in the `ClassHash`, I gave more details in the comment.
`CompiledClassHash` is working fine, congrats!

Could you please handle this?

thiagodeev (approved):

thiagodeev:
Thank you again @baitcode for your contribution!
It was a great help

Issue #657, opened 2025-01-22
Nil calldata leads to null calldata array


Hyodar:
When making a function call and not specifying any calldata array, such as with `rpc.FunctionCall{ ContractAddress: a, EntryPointSelector: b }`, which would make sense for functions that don't require any parameters, the resulting JSON-RPC call is of the following form:
```JSON
{
  "jsonrpc": "2.0",
  "id": 0,
  "method": "starknet_call",
  "params": [
    {
      "contract_address": "",
      "entry_point_selector": "",
      "calldata": null
    },
    ""
  ]
}
```
A call of this form works in Juno, but depending on the RPC it fails. Checking the Starknet RPC API specification on the `FUNCTION_CALL` type (https://github.com/starkware-libs/starknet-specs/blob/2030a650be4e40cfa34d5051a0334f375384a421/api/starknet_api_openrpc.json#L3006), `calldata` is specified as type `array`, so it would be correct to expect `null` would fail validation.

This can be avoided without any changes to the library though! The user simply needs to set the calldata array to an empty array, such as `rpc.FunctionCall{ ContractAddress: a, EntryPointSelector: b, Calldata: []*felt.Felt{} }`.

Still, I propose we either change the JSON encoding or, probably better, set the calldata internally to an empty array if it's nil.

Hyodar:
Closing this as #658 was merged.

PR #658, opened 2025-01-22
Ensure starknet_call calldata is not null

Hyodar:
Refers to #657. Feel free to merge this if it makes sense, and not otherwise!

### DIFF
--- a/rpc/call.go
+++ b/rpc/call.go
@@ -6,6 +6,8 @@ import (
 	"github.com/NethermindEth/juno/core/felt"
 )
 
+var emptyCalldata = []*felt.Felt{}
+
 // Call calls the Starknet Provider's function with the given (Starknet) request and block ID.
 //
 // Parameters:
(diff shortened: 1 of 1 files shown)


rpc/call.go
```diff
@@ -6,6 +6,8 @@ import (
 	"github.com/NethermindEth/juno/core/felt"
 )
 
+var emptyCalldata = []*felt.Felt{}
```
thiagodeev: Can you remove this var and just assing the `[]*felt.Felt{}` directly in the `if request.Calldata == nil` statement?

thiagodeev (changes requested):
Thanks!
Just a small request

rpc/call.go
Hyodar: Sure, no problem with that. I'm not sure of the compiler specifics here to compare both approaches, but it seems fine!

Hyodar: Done in [cf24819](https://github.com/NethermindEth/starknet.go/pull/658/commits/cf24819f97f63dbcf264f5c25e2b624fb8422389)

thiagodeev (approved):

PR #662, opened 2025-02-07
Thiagodeev/rpcv08 - implement/fix/update remaining tests for rpcv08

thiagodeev:

- TestEstimateMessageFee now passing
- TestEstimateFee now passing
- TestSimulateTransaction now passing
- TestBlockWithReceipts now passing
- TestBlockWithTxs now passing
- TestSpecVersion updated
- TestGetTransactionStatus: add test case for the new `FailureReason` field
- Add tests for CompiledCasm method (mock and testnet)
  - `types_executables.go` refactored in order to make it work
- Add tests for GetMessagesStatus method (mock and testnet)

New:
- UnmarshallJSONToType: reads a JSON file at the given path and unmarshals it into the specified type T
- TestUnmarshallJSONToType: same as above, but for testing
- RPCData refactor


utils/Test.go
```diff
@@ -37,3 +39,42 @@ func TestHexArrToFelt(t testing.TB, hexArr []string) []*felt.Felt {
...
+// TestUnmarshallJSONToType reads a JSON file at the given path and unmarshals it into the specified type T.
+// If any error occurs during file reading or unmarshalling, it fails the test.
+//
+// Parameters:
+// - t: testing.TB interface for test logging and error reporting
+// - filePath: string path to the JSON file
+// - subfield: string subfield to unmarshal from the JSON file
+// Returns:
+// - T: the unmarshalled data of type T
+func TestUnmarshallJSONToType[T any](t testing.TB, filePath string, subfield string) *T {
```
rianhughes: Why not re-use UnmarshallJSONToType?

rpc/executables_test.go
```diff
@@ -1,7 +1,95 @@
...
+		},
+		"mainnet": {},
+	}[testEnv]
+
+	for _, test := range testSet {
+		var expectedResult *CasmCompiledContractClass
+		var rawFile []byte
+		var err error
+
+		if test.ExpectedResultPath != "" {
```
rianhughes, line 64: Can we use `TestUnmarshallJSONToType` here?

rpc/types_executables.go
```diff
@@ -143,11 +371,72 @@ type Felt252DictEntryUpdate struct {
...
+		return fmt.Errorf("res operand must have exactly one field, got %d", len(raw))
+	}
+
+	// Get the single key and value
+	var hintType string
+	var hintData json.RawMessage
+	for k := range raw {
+		hintType = k
+		hintData = raw[k]
+		break
```
rianhughes: why break, we should only enter the loop once right?

rianhughes:
LGTM, nice one

rpc/executables_test.go
thiagodeev: Not here :/
We need the `rawFile` variable at the end to compare the JSONs

rpc/types_executables.go
thiagodeev: Well, that's true haha. I just left it because it's one less evaluation for the code to do (`for` checking if the `range` was already reached...). Nothing special.

Remove it then?

utils/Test.go
thiagodeev: Makes sense. 
I intended to return the errors with `t.fatal` in this case, but I can simply have `TestUnmarshallJSONToType` calling `UnmarshallJSONToType` and returning the error with `t.fatal` instead.
I'll do it.

thiagodeev: BTW, WDYT of renaming it to `UnmarshallJSONFIleToType`?

thiagodeev:
> LGTM, nice one

Thanks! We need to merge https://github.com/NethermindEth/starknet.go/pull/651 before this one to properly test the `GetStorageProof` method

utils/Test.go
rianhughes: Yeah that works

rpc/types_executables.go
rianhughes: Yeah

rpc/errors.go
```diff
@@ -271,52 +291,83 @@ var (
...
 }
 
-type TransactionExecutionErrorData struct {
-	// The index of the first transaction failing in a sequence of given transactions
-	TransactionIndex int `json:"transaction_index,omitempty"`
-	// the execution trace up to the point of failure
-	ExecutionError ContractExecutionError `json:"execution_error,omitempty"`
+func (c *ContractErrData) ErrorMessage() string {
+	return c.RevertError.Message
+}
```
rianhughes, line 342: Do we not also have to return information about ContractExecErrInner?
```
type ContractExecutionError struct {
	// the error raised during execution
	Message              string                       `json:",omitempty"`
	ContractExecErrInner *ContractExecutionErrorInner `json:",omitempty"`
}
```

rpc/errors.go
```diff
@@ -271,52 +291,83 @@ var (
...
+// Structured type for the ErrTransactionExecError data
+type TransactionExecErrData struct {
+	TransactionIndex int                    `json:"transaction_index,omitempty"`
+	ExecutionError   ContractExecutionError `json:"execution_error,omitempty"`
+}
+
+func (t *TransactionExecErrData) ErrorMessage() string {
+	return t.ExecutionError.Message
+}
+
```
rianhughes, line 353: Do we not also have to return information about ContractExecErrInner?

type ContractExecutionError struct {  
    // the error raised during execution  
    Message              string                       `json:",omitempty"`  
    ContractExecErrInner *ContractExecutionErrorInner `json:",omitempty"`  
}

rpc/errors.go
thiagodeev: We could.
My intent was to have the user who did not even cast `err` to `RPCError`, when calling `err.Error()`, get some piece of information about the RPCData error field. But if he really wants to know more about it, he needs to cast it to the correct RPCData type associated with the error

rpc/errors.go
thiagodeev: Same as above
https://github.com/NethermindEth/starknet.go/pull/662#discussion_r1973680488

rpc/errors.go
thiagodeev: But it's not bad at all, having the entire error when calling `err.Error()`. Do you want me to make it?

rpc/errors.go
```diff
@@ -65,92 +65,107 @@ func tryUnwrapToRPCErr(baseError error, rpcErrors ...*RPCError) *RPCError {
...
+		if err := json.Unmarshal(temp.Data, &data); err != nil {
+			return err
+		}
+		e.Data = &data
+	case 55, 56, 63: // ErrValidationFailure, ErrCompilationFailed, ErrUnexpectedError
+		var strData string
+		if err := json.Unmarshal(temp.Data, &strData); err != nil {
+			return err
+		}
+		e.Data = StringErrData(strData)
```
rianhughes, line 136: Do we need to marshal again? Can we not just set `e.Data = StringErrData(temp.Data)`

rianhughes: Similar to the default case?

rpc/errors.go
```diff
@@ -65,92 +65,107 @@ func tryUnwrapToRPCErr(baseError error, rpcErrors ...*RPCError) *RPCError {
...
-	Message string   `json:"message"`
-	Data    *RPCData `json:"data,omitempty"`
+	Code    int     `json:"code"`
+	Message string  `json:"message"`
+	Data    RPCData `json:"data,omitempty"`
 }
 
 func (e RPCError) Error() string {
-	if e.Data == nil || e.Data.Message == "" {
+	if e.Data == nil || e.Data.ErrorMessage() == "" {
```
rianhughes, line 81: This can panic if e.Data is nil

rpc/errors.go
thiagodeev: I don't get it. It's not being marshaled again 🤔, it's almost the same as the default.
 Here we check if "err != nil", and in default we check if "err == nil"

rpc/errors.go
rianhughes: Yeah, I think users who cast to RPCError, should be able to get the entire error data information in `ErrorMessage()`

rpc/errors.go
rianhughes: Hm, I guess we should leave it as is actually. I was thinking about the case where clients actually return a string in the data field, in which case I think we can skip calling Unmarshal, and just use `e.Data = StringErrData(temp.Data)`. But if there's a bug in the client, this version might be able to catch it

rpc/errors.go
thiagodeev: But the user who cast can access these field directly in the struct type (e.g: ContractAddress, ClassHash..., and the Error).
ErrorMessage() was meant to return a piece of the data field without having to cast it to RPCError, just by calling err.Error()

rianhughes (approved):

PR #666, opened 2025-02-17
chore: fix misleading comment for StrToHex function

fragchain:
I noticed a discrepancy in the comment for the `StrToHex` function. The comment states that the function generates hexadecimal from a string/number, but the function name (`StrToHex`) suggests it only converts a string to hex.

This could be confusing for anyone reading the code. I’ve updated the comment to clarify that the function specifically converts a string to hexadecimal, aligning it with the function name and its actual behavior.

### DIFF
--- a/utils/keccak.go
+++ b/utils/keccak.go
@@ -46,7 +46,7 @@ func StrToBig(str string) *big.Int {
 	return b
 }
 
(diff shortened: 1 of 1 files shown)


thiagodeev:
Thanks @fragchain, but both descriptions are good IMO, so I'll leave it as it is

PR #668, opened 2025-02-19
Fix ByteArrFeltToString hex decoding error

Hyodar:
Hello there! I got some decoding failure when using `utils.ByteArrFeltToString` - seems like in `feltToString` we rely on `hex.DecodeString`, but the felt string can be odd-length which causes issues. Ideally this should now be fixed and a bit cleaner! I also added some more input checks to avoid unexpected panics.


thiagodeev (changes requested):
Thanks @Hyodar!

I'll just ask one thing: could you put the case that was causing issues in the `ByteArrFeltToString` test cases? That way we can be sure it'll continue to work in the future.

thiagodeev:
Hi @Hyodar!
Could you please resolve the conflicts in this PR? Thanks!

Hyodar:
Should be done @thiagodeev!

thiagodeev (approved):

PR #673, opened 2025-02-28
Thiagodeev/rpcv08 new txn utilities

thiagodeev:

This PR aims to:

- Create new methods to improve the UX when sending and building txns. To do so, it creates new three methods:
  - BuildAndSendInvokeTxn
  - BuildAndSendDeclareTxn
  - BuildAndEstimateDeployAccountTxn
These methods encapsulate all the 'hard work' of building, estimating, calculating, and sending a transaction. If a user wants to do it manually, it's still possible, and new helpers were created for that too.
- New utilities:
  - BuildInvokeTxn
  - BuildDeclareTxn
  - BuildDeployAccountTxn
  - InvokeFuncCallsToFunctionCalls (a new type was created to encapsulate the process of the encoding 'entry points')
  - FeeEstToResBoundsMap
  - ResBoundsMapToOverallFee
  - WeiToETH
  - ETHToWei
  - FRIToSTRK
  - STRKToFRI 

The examples have also been updated to use these new utilities, but have not been fully updated as this is not the focus of this PR.
In addition, to build and test these helpers, v3 transactions have been completely patched, and are 100% compliant with rpcv08 spec.


account/account.go
```diff
@@ -79,6 +84,208 @@ func NewAccount(provider rpc.RpcProvider, accountAddress *felt.Felt, publicKey s
...
+//   - multiplier: A safety factor for fee estimation that helps prevent transaction failures due to
+//     fee fluctuations. It multiplies both the max amount and max price per unit by this value.
+//     A value of 1.5 (50% buffer) is recommended to balance between transaction success rate and
+//     avoiding excessive fees. Higher values provide more safety margin but may result in overpayment.
+//
+// Returns:
+// - *rpc.BroadcastDeployAccountTxnV3: the transaction to be broadcasted, signed and with the estimated fee based on the multiplier
+// - *felt.Felt: the precomputed account address as a *felt.Felt, it needs to be funded with appropriate amount of tokens
+// - error: an error if any
+func (account *Account) BuildAndEstimateDeployAccountTxn(
```
rianhughes, line 219: What do you think about making a comment that we don't submit the txn because the precomputedAddress needs to be funded beforehand?

account/account.go
```diff
@@ -79,6 +84,208 @@ func NewAccount(provider rpc.RpcProvider, accountAddress *felt.Felt, publicKey s
...
+		},
+		L1DataGas: rpc.ResourceBounds{
+			MaxAmount:       "0x0",
+			MaxPricePerUnit: "0x0",
+		},
+		L2Gas: rpc.ResourceBounds{
+			MaxAmount:       "0x0",
+			MaxPricePerUnit: "0x0",
+		},
+	})
```
rianhughes: This pattern is repeated in a couple of places (4 I think), what do you think about creating a function that returns a "zero'd" ResourceBounds?

account/account_test.go
```diff
@@ -385,20 +385,8 @@ func TestSignMOCK(t *testing.T) {
...
-				ChainId:    "SN_SEPOLIA",
-				FeltToSign: internalUtils.TestHexToFelt(t, "0x4b2e6743b03a0412f8450dd1d337f37a0e946603c3e6fbf4ba2469703c1705b"),
-				ExpectedSig: []*felt.Felt{
-					internalUtils.TestHexToFelt(t, "0xfa671736285eb70057579532f0efb6fde09ecefe323755ffd126537234e9c5"),
-					internalUtils.TestHexToFelt(t, "0x27bf55daa78a3ccfb7a4ee6576a13adfc44af707c28588be8292b8476bb27ef"),
-				},
-			},
-		},
+		"devnet":  {},
+		"testnet": {},
```
rianhughes, line 424: Why remove the testnet case?

account/account_test.go
```diff
@@ -1190,3 +1148,27 @@ func newDevnet(t *testing.T, url string) (*devnet.DevNet, []devnet.TestAccount,
 	acnts, err := devnet.Accounts()
 	return devnet, acnts, err
 }
+
+// TestBuildAndSendInvokeTxn is a test function that tests the BuildAndSendInvokeTxn method.
+//
+// This function tests the BuildAndSendInvokeTxn method by setting up test data and invoking the method with different test sets.
+// It asserts that the expected hash and error values are returned for each test set.
+func TestBuildAndSendInvokeTxn(t *testing.T) {
+	t.Skip("TODO: implement this test once devnet supports full v3 transaction type")
```
rianhughes: Why not test on testnet, similar to how we do in TestSendInvokeTxn etc? Similarly for TestBuildAndSendDeclareTxn and TestBuildAndSendDeployAccTxn

utils/transactions.go
```diff
@@ -0,0 +1,255 @@
...
+// Parameters:
+//   - senderAddress: The address of the account sending the transaction
+//   - nonce: The account's nonce
+//   - calldata: The data expected by the account's `execute` function (in most usecases,
+//     this includes the called contract address and a function selector)
+//   - resourceBounds: Resource bounds for the transaction execution
+//
+// Returns:
+//   - rpc.BroadcastInvokev3Txn: A broadcast invoke transaction with default values
+//     for signature, tip, paymaster data, etc. Need to be signed before being sent.
```
rianhughes: Type: Needs

rianhughes: Similarly for BuildInvokeTxn, BuildDeclareTxn and BuildDeployAccountTxn

utils/transactions.go
```diff
@@ -0,0 +1,255 @@
...
+// Returns:
+//   - rpc.ResourceBounds: Resource bounds with applied multiplier
+func toResourceBounds(
+	gasPrice uint64,
+	gasConsumed uint64,
+	multiplier float64,
+) rpc.ResourceBounds {
+	maxAmount := float64(gasConsumed) * multiplier
+	maxPricePerUnit := float64(gasPrice) * multiplier
+
```
rianhughes: Currently this can overflow, so I think we should multiply in felt form (which can't overflow), then cast to U64 etc

utils/transactions.go
```diff
@@ -0,0 +1,255 @@
...
+	l2GasAmount := HexToBN(string(resBounds.L2Gas.MaxAmount))
+	l2GasPrice := HexToBN(string(resBounds.L2Gas.MaxPricePerUnit))
+
+	// calculate fee
+	l1GasFee := new(big.Int).Mul(l1GasAmount, l1GasPrice)
+	l1DataGasFee := new(big.Int).Mul(l1DataGasAmount, l1DataGasPrice)
+	l2GasFee := new(big.Int).Mul(l2GasAmount, l2GasPrice)
+
+	// return fee
+	return l1GasFee.Add(l1GasFee, l1DataGasFee).Add(l1GasFee, l2GasFee)
```
rianhughes: I'm slightly concerned that users might use this function to caluclate the overall fee. In which case it's safert to perform the arithmetic using felts than big.Ints

utils/transactions_test.go
```diff
@@ -0,0 +1,107 @@
...
+
+import (
+	"math/big"
+	"testing"
+
+	"github.com/NethermindEth/starknet.go/rpc"
+	"github.com/stretchr/testify/assert"
+)
+
+func TestResBoundsMapToOverallFee(t *testing.T) {
```
rianhughes, line 15: Can we add a test case where we try and force an overflow?

rianhughes: And maybe negative values / underflow?

utils/transactions.go
```diff
@@ -0,0 +1,255 @@
...
+// Parameters:
+//   - resBounds: The resource bounds to calculate the fee for
+//   - multiplier: Multiplier for max amount and max price per unit. Recommended to be 1.5, but at least 1
+//
+// Returns:
+//   - *big.Int: The overall fee in FRI
+func ResBoundsMapToOverallFee(
+	resBounds rpc.ResourceBoundsMapping,
+	multiplier float64,
+) *big.Int {
```
rianhughes: Imo we shouldn't allow negative values, and should maybe cap it to the maximum allowed felt

examples/simpleInvoke/main.go
```diff
@@ -57,74 +56,26 @@ func main() {
...
-		}}
-
 	// Converting the contractAddress from hex to felt
 	contractAddress, err := utils.HexToFelt(someContract)
 	if err != nil {
 		panic(err)
 	}
 
 	amount, _ := utils.HexToFelt("0xffffffff")
+
```
rianhughes, line 69: Can we handle the error on the line above?
amount, _ := utils.HexToFelt("0xffffffff")

examples/simpleDeclare/README.md
```diff
@@ -0,0 +1,14 @@
...
+1. Uncomment, and assign your Sepolia testnet endpoint to the `RPC_PROVIDER_URL` variable in the ".env" file
+1. Uncomment, and assign your account address to the `ACCOUNT_ADDRESS` variable in the ".env" file (make sure to have a few ETH in it)
+1. Uncomment, and assign your starknet public key to the `PUBLIC_KEY` variable in the ".env" file
+1. Uncomment, and assign your private key to the `PRIVATE_KEY` variable in the ".env" file
+1. Make sure you are in the "simpleInvoke" directory
+1. Execute `go run main.go`
+
+The transaction hash and status will be returned at the end of the execution.
+
+<!-- TODO: update this -->
```
rianhughes, line 14: We can do this now right?

rianhughes: We also need to add ACCOUNT_CAIRO_VERSION in the template

account/account_test.go
thiagodeev: The test itself is called 'TestSignMOCK', and you can notice that the 'testnet' case is exactly the same as the 'mock' one. As this test is intended to test the Sign function, and it just compares the signature without even interacting with the blockchain (except for the ChainId, but it is mocked), the 'testnet' case is just duplicated code without purpose

account/account_test.go
thiagodeev: Because the TestSendInvokeTxn sends a txn with fixed values, so the txn hash always matches the one in the test, but these new functions estimate the fee and get the nonce for sending, so it'll always be different

utils/transactions.go
thiagodeev: You're right. I don't think this would be a problem with maxAmount, since the field by the spec is a U64, but `maxPricePerUnit` is a U128 so the operation indeed must be made with big.Ints

utils/transactions.go
thiagodeev: I'm okay with returning a felt instead of big.Int. I wasn't thinking about < v3 txns when I made these helpers.
Doing this, the felt cap will be done automatically by the felt type.

Regarding negative values, I think there's no problem with keeping this. We would need to create a validation and return an error for this, this will remove the simplicity of this func. The user putting a negative value is his problem IMO, because the func descriptions explain it all.

WDYT?

examples/simpleDeclare/README.md
thiagodeev: I had decided to update the examples in a specific PR, it's 677.
I have done it here: https://github.com/NethermindEth/starknet.go/pull/677/files#diff-1a8b04ca53900c0b596c93848bb8db2f386d30e3f196b9b7cada503dd4141695

It's okay to you to proceed this way?

account/account.go
thiagodeev: Done

account/account.go
thiagodeev: Done

utils/transactions.go
thiagodeev: Done

utils/transactions.go
thiagodeev: Done and tested

utils/transactions.go
thiagodeev: I changed my mind, I'm doing it

utils/transactions.go
thiagodeev: Done and tested

utils/transactions_test.go
thiagodeev: Done

utils/transactions.go
thiagodeev: Done

examples/simpleInvoke/main.go
thiagodeev: Done

account/account_test.go
rianhughes: `TestSendInvokeTxn` doesn't depend on the txn hash, we just submit the (fixed) txn and check we get a nil error and non-nil resp. I think we could do something similar here. Or even implement a mock test. What do you think?

examples/simpleDeclare/README.md
rianhughes: Yup!

utils/transactions.go
```diff
@@ -0,0 +1,362 @@
...
+		return nil, fmt.Errorf(negativeResourceBoundsErr, l1DataGasPrice)
+	}
+
+	l2GasAmount, ok := new(big.Int).SetString(string(resBounds.L2Gas.MaxAmount), 0)
+	if !ok {
+		return nil, fmt.Errorf(invalidResourceBoundsErr, resBounds.L2Gas.MaxAmount)
+	}
+	if l2GasAmount.Sign() < 0 {
+		return nil, fmt.Errorf(negativeResourceBoundsErr, l2GasAmount)
+	}
```
rianhughes: There's a lot of repeated code here, what do you think about creating an anonymous function? This helps reduce repetition, and indicate that the same operation is being repeated for each type.
```
func ResBoundsMapToOverallFee(
	resBounds rpc.ResourceBoundsMapping,
	multiplier float64,
) (*felt.Felt, error) {
	// negative multiplier is not allowed
	if multiplier < 0 {
		return nil, errors.New("multiplier cannot be negative")
	}

parseBound := func(value string) (*big.Int, error) {
		val, ok := new(big.Int).SetString(value, 0)
		if !ok {
			return nil, fmt.Errorf(invalidResourceBoundsErr, value)
		}
		if val.Sign() < 0 {
			return nil, fmt.Errorf(negativeResourceBoundsErr, val)
		}
		return val, nil
	}

	// Parse all resource bound values
	l1GasAmount, err := parseBound(string(resBounds.L1Gas.MaxAmount))
	if err != nil {
		return nil, err
	}

	l1GasPrice, err := parseBound(string(resBounds.L1Gas.MaxPricePerUnit))
	if err != nil {
		return nil, err
	}
    ```

File: utils/transactions.go [hunk 13]
```diff
@@ -0,0 +1,362 @@
...
+
+	// multiply fee by multiplier
+	multipliedOverallFee := new(big.Float).Mul(new(big.Float).SetInt(overallFee), big.NewFloat(multiplier))
+	overallFeeInt, _ := multipliedOverallFee.Int(nil) // truncated int
+
+	// Convert big.Int to felt. SetString() validates if it's a valid felt
+	overallFeeFelt, err := new(felt.Felt).SetString(fmt.Sprintf("%#x", overallFeeInt))
+	if err != nil {
+		return nil, err
+	}
```
rianhughes: overallFeeFelt := new(felt.Felt).SetBigInt(overallFeeInt)

rianhughes: Actually, we can just return early, ```return new(felt.Felt).SetBigInt(overallFeeInt), nil```

thiagodeev: I avoided using SetBigInt because it didn't return an error, and that intrigued me since the big.Int doesn't have a limit but the felt has.
I preferred to use SetString because it returns a clear error when the number doesn't fit a felt.

account/account_test.go
thiagodeev: I said that because, since the transaction is already sent, it will never be sent again, so the test now only tests if the returned error is error 59: "A transaction with the same hash already exists in the mempool". That will not be the case with the new methods, as the txn hash will be always different. We will need an address with a lot of STRK token to send the txns to tesnet until devnet implements v3 txns.

utils/transactions.go
thiagodeev: Done

utils/transactions.go
rianhughes: Good point, I _think_ if `overallFeeInt` is larger than the `maxFelt`, then  `new(felt.Felt).SetBigInt(overallFeeInt)` will just return `overallFeeInt mod(maxFelt)`. This prevents any potential overflow

rianhughes (approved):
Very nice work! The UX is much nicer now :)

PR #629, opened 2024-09-30
RPC v0.8.0

thiagodeev:
Issue: https://github.com/NethermindEth/starknet.go/issues/623

This PR aggregates all v0.8.0 changes

PR #675, opened 2025-03-03
fix: fix the wrong error return value

drawdrop:
fix the wrong error return value


thiagodeev:
Thanks @drawdrop, but these errors where fixed in https://github.com/NethermindEth/starknet.go/pull/629/

PR #680, opened 2025-03-21
Thiagodeev/rpcv08 final adjustments v2

thiagodeev:

This PR aims to:
// general
- Allow RPC and Account tests to run on CI
- Fix and improve some tests and descriptions
- Update some functions to use pointers instead of values
- RPCError now returns the full error in `err.Error()`
- Fix StorageProof types
- Remove the `setup.PanicRPC` func as it is no longer necessary due to the RPCError.Error() updates

// account
- New `account.Nonce()` method
- New `TransactionHash...` functions in the `account` pkg
- Make `account.SignTxn...` methods to accept all txns versions
- Make the private field `provider` of the account be public
- Remove duplicated provider methods from the account type

// call
- New `CallContextWithSliceArgs` client method, and change `CallContext` to accept only a single value
- New `doAsObject` to call `CallContext`, used for methods with optional parameters (like `starknet_getStorageProof`)


account/account.go
```diff
@@ -426,86 +523,109 @@ func (account *Account) TransactionHashDeployAccount(tx rpc.DeployAccountType, c
 
 // If the transaction type is unsupported, the function returns an error.
 func (account *Account) TransactionHashInvoke(tx rpc.InvokeTxnType) (*felt.Felt, error) {
-
-	// https://docs.starknet.io/architecture-and-concepts/network-architecture/transactions/#v0_deprecated_hash_calculation
 	switch txn := tx.(type) {
+	// invoke v0, pointer and struct
+	case *rpc.InvokeTxnV0:
```
MaksymMalicki, line 528: will it ever be a pointer?

account/account.go
```diff
@@ -356,64 +437,80 @@ func (account *Account) TransactionHashDeployAccount(tx rpc.DeployAccountType, c
 
 	// https://docs.starknet.io/architecture-and-concepts/network-architecture/transactions/#deploy_account_transaction
 	switch txn := tx.(type) {
+	// deployAccTxn v0, pointer and struct
+	case *rpc.DeployAccountTxn:
```
MaksymMalicki, line 441: will it ever be a pointer?

account/account.go
```diff
@@ -761,315 +914,6 @@ func (account *Account) SendTransaction(ctx context.Context, txn rpc.BroadcastTx
 	}
 }
 
-// BlockHashAndNumber returns the block hash and number for the account.
```
MaksymMalicki, line 764: why are those methods removed? deprecated?

client/client.go
```diff
@@ -322,15 +322,23 @@ func (c *Client) SetHeader(key, value string) {
...
 	ctx := context.Background()
-	return c.CallContext(ctx, result, method, args...)
+	return c.CallContextWithSliceArgs(ctx, result, method, args...)
+}
+
+// Call 'CallContext' with a slice of arguments.
+//
+// For RPC-Calls with optional arguments, use 'CallContext' instead and pass a struct containing
+// the arguments, because Juno doesn't support optional arguments being passed in an array, only within an object.
+func (c *Client) CallContextWithSliceArgs(ctx context.Context, result interface{}, method string, args ...interface{}) error {
```
MaksymMalicki, line 332: nice 😀

contracts/types_casm_class.go
```diff
@@ -219,7 +219,8 @@ func (hints *Hints) Tuple() [2]any {
 	return [2]any{hints.Int, hints.HintArr}
 }
 
-// Can be one of the following hints
+// Can be one of various hint types described in the spec and in the UnmarshalJSON method
+// Spec ref: https://github.com/starkware-libs/starknet-specs/blob/d70a2e57c9a66db1bbf86c388f38d6295a6a2a75/api/starknet_executables.json#L276
```
MaksymMalicki, line 223: nice, that you added the source

hash/hash.go
```diff
@@ -244,7 +244,7 @@ func hashCasmClassByteCode(
 //
 // Returns:
 //   - *felt.Felt: a pointer to a felt.Felt object that represents the calculated hash.
-func CompiledClassHash(casmClass contracts.CasmClass) (*felt.Felt, error) {
+func CompiledClassHash(casmClass *contracts.CasmClass) (*felt.Felt, error) {
```
MaksymMalicki, line 247: why this modification?

account/account.go
```diff
@@ -33,13 +33,13 @@ var (
 type AccountInterface interface {
```
MaksymMalicki, line 33: I was wondering, why methods here return a pointer (mosty pointer to the felt)? Won't it cause additional heap allocations?

examples/internal/setup.go
```diff
@@ -20,21 +19,6 @@ func init() {
 	}
 }
 
-// Default "panic" but printing all RPCError fields (code, message, and data)
```
MaksymMalicki, line 23: nice, cool cleanup

MaksymMalicki (approved):

account/account.go
thiagodeev: It could. It's a public method from the account type, so If the user wants to pass it as a pointer he can

account/account.go
thiagodeev: It could. It's a public method from the account type, so If the user wants to pass it as a pointer he can

account/account.go
thiagodeev: They are just duplicates of the provider methods, makes no sense to me.
If the user wants to use the provider's methods, he can do it by calling the provider itself (or by the `acc.Provider()` method). I want to make them truly different abstractions rather than a semi-copy

hash/hash.go
thiagodeev: These classes can be quite large, so passing them as pointers instead of a values is a better approach IMO

account/account.go
thiagodeev: That one I can't answer :/, I'm just following the code patterns (although I saw that Juno does the same a lot)

PR #681, opened 2025-03-25
Thiagodeev/rpcv08 final adjustments v3

thiagodeev:

This PR aims to:
- Remove deprecated Broadcast Txn versions, keeping only V3 txns.
- Move all `TransactionHash...` functions from the `account` pkg to the `hash` pkg
- Make `AddTransactions...`  and `acc.SendTransaction()` methods only accept V3 txns.
- Remove BlockIDWithoutPending and SubscriptionBlockID types, use `BlockID` instead with a helper func.
- New `fillEmptyFeeEstimation` internal util to bypass an estimateFee bug.
- New `HexToU256Felt` and `U256FeltToHex` utils to deal with U256 cairo var type.
- Update examples to use the `HexToU256Felt` utils
- Make `BuildTxns...` utils return a pointer instead of a struct
- Fix and improve some tests and descriptions.

### DIFF
--- a/account/account.go
+++ b/account/account.go
@@ -569,10 +569,11 @@ func (account *Account) WaitForTransactionReceipt(ctx context.Context, transacti
 }
 
 // SendTransaction can send Invoke, Declare, and Deploy transactions. It provides a unified way to send different transactions.
+// It can only send v3 transactions.
 //
 // Parameters:
 //   - ctx: the context.Context object for the transaction.
-//   - txn: the Broadcast Transaction to be sent.
+//   - txn: the Broadcast V3 Transaction to be sent.
 //
 // Returns:
 //   - *rpc.TransactionResponse: the transaction response.
@@ -619,7 +620,7 @@ func (account *Account) SendTransaction(ctx context.Context, txn rpc.BroadcastTx
 		}
 		return &rpc.TransactionResponse{TransactionHash: resp.TransactionHash, ContractAddress: resp.ContractAddress}, nil
 	default:
-		return nil, errors.New("unsupported transaction type")
+		return nil, fmt.Errorf("unsupported transaction type: should be a v3 transaction, instead got %T", tx)
 	}
 }
 
--- a/account/account.go
+++ b/account/account.go
@@ -176,7 +176,7 @@ func (account *Account) BuildAndSendDeclareTxn(
 	if err != nil {
 		return nil, err
 	}
-	err = account.SignDeclareTransaction(ctx, &broadcastDeclareTxnV3)
+	err = account.SignDeclareTransaction(ctx, broadcastDeclareTxnV3)
 	if err != nil {
 		return nil, err
 	}
@@ -190,7 +190,7 @@ func (account *Account) BuildAndSendDeclareTxn(
 	broadcastDeclareTxnV3.ResourceBounds = utils.FeeEstToResBoundsMap(txnFee, multiplier)
 
 	// signing the txn again with the estimated fee, as the fee value is used in the txn hash calculation
-	err = account.SignDeclareTransaction(ctx, &broadcastDeclareTxnV3)
+	err = account.SignDeclareTransaction(ctx, broadcastDeclareTxnV3)
 	if err != nil {
 		return nil, err
 	}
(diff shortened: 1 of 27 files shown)


account/account.go
```diff
@@ -881,46 +569,69 @@ func (account *Account) WaitForTransactionReceipt(ctx context.Context, transacti
 }
 
 // SendTransaction can send Invoke, Declare, and Deploy transactions. It provides a unified way to send different transactions.
+// It can only send v3 transactions.
```
MaksymMalicki, line 572: sorry if I lack a bit of context here, why do we send only v3 transactions?

account/account.go
```diff
@@ -1005,3 +719,22 @@ func makeResourceBoundsMapWithZeroValues() rpc.ResourceBoundsMapping {
...
 	}
 }
+
+// When there's no transaction in the pending block, the L1DataGasConsumed and L1DataGasPrice fields are comming empty.
+// This is causing the transaction to fail with the error:
+// "55 Account validation failed: Max L1DataGas amount (0) is lower than the minimal gas amount: 128"
+// This function fills the empty fields.
+//
+// TODO: remove this function once the issue is fixed in the RPC
+func fillEmptyFeeEstimation(ctx context.Context, feeEstimation *rpc.FeeEstimation, provider rpc.RpcProvider) {
```
MaksymMalicki, line 729: 👍

rpc/contract.go
```diff
@@ -182,6 +182,11 @@ func (provider *Provider) EstimateMessageFee(ctx context.Context, msg MsgFromL1,
 // the path to it may end in an edge node whose path is not a prefix of the requested leaf, thus effectively proving non-membership
 // - error: an error if any occurred during the execution
 func (provider *Provider) GetStorageProof(ctx context.Context, storageProofInput StorageProofInput) (*StorageProofResult, error) {
+	err := checkForPending(storageProofInput.BlockID)
```
MaksymMalicki, line 185: why do this check?

rpc/types_block.go
```diff
@@ -27,14 +27,13 @@ type BlockID struct {
...
-// Used in the 'starknet_getStorageProof' RPC method, as it doesn't support pending blocks
-type BlockIDWithoutPending BlockID
-
-// Block hash, number or tag, same as BLOCK_ID, but without 'pending'
-type SubscriptionBlockID struct {
-	Number uint64     `json:"block_number,omitempty"`
-	Hash   *felt.Felt `json:"block_hash,omitempty"`
-	Tag    string     `json:"block_tag,omitempty"`
+// checkForPending checks if the block ID has the 'pending' tag. If it does, it returns an error.
+// This is used to prevent the user from using the 'pending' tag on methods that do not support it.
```
MaksymMalicki, line 31: *nit: does not support it.
Why are all those types removed? Again, Im lacking context a bit

rpc/types_block.go
```diff
@@ -27,14 +27,13 @@ type BlockID struct {
 	Tag    string     `json:"block_tag,omitempty"`
 }
 
-// Used in the 'starknet_getStorageProof' RPC method, as it doesn't support pending blocks
-type BlockIDWithoutPending BlockID
```
MaksymMalicki, line 31: Why are all those types removed? Again, Im lacking context a bit

MaksymMalicki (approved):

account/account.go
thiagodeev: No problem.
Because rpc v0.8.0 ends support for transactions v0,1 and 2.
https://docs.starknet.io/resources/version-notes/#version0.13.4

rpc/contract.go
thiagodeev: This, and the other methods I've used this helper with, don't allow "pending" blocks.
https://github.com/starkware-libs/starknet-specs/blob/master/api/starknet_api_openrpc.json#L897

rpc/types_block.go
thiagodeev: I had created specific types to handle `block_id` requests without the "pending" tag, but now I've removed them and am using the original `BlockID` type, using a helper ( the `checkForPending` func) to check if it has a "pending" tag on it.

PR #687, opened 2025-04-04
v0.8.1 release changes

thiagodeev:
- change `subscription_id` to string
- update spec test to v0.8.1


client/handler.go
```diff
@@ -463,9 +456,15 @@ func (h *handler) handleSubscriptionResult(msg *jsonrpcMessage) {
 		return
 	}
 
-	id := strconv.FormatUint(result.StarknetID, 10)
-	if id == "0" {
-		id = result.ID
+	id := result.StarknetID
+	// if the StarknetID is empty, means it could be subscription from the ethereum tests
```
wojciechos, line 460: Why do we need ethereum tests compatibility here? Isn't this client only supposed for starknet?

thiagodeev: To take advantage of all the go-ethereum client tests with minimal effort.
I preferred to make these simple adjustments instead of refactoring all their tests.

wojciechos (approved):

PR #692, opened 2025-04-08
Upgrade juno version to v0.14.0, and Go to 1.23.4

thiagodeev:


wojciechos:
Why not update to latest? I mean Juno to v0.14.3 and Go to 1.24.2?

thiagodeev:
@wojciechos I thought that it would be good to pick a release that is not too recent to avoid having to upgrade again due to a bug on it found later; just a small precaution. But I can update it to v0.14.3 if you want

MaksymMalicki (approved):

wojciechos:
> @wojciechos I thought that it would be good to pick a release that is not too recent to avoid having to upgrade again due to a bug on it found later; just a small precaution. But I can update it to v0.14.3 if you want

Gotcha, either works for me

wojciechos (approved):

PR #695, opened 2025-04-09
Allow empty block_id, and update code to use it

thiagodeev:


rpc/websocket.go
```diff
@@ -26,18 +26,9 @@ import (
 //   - clientSubscription: The client subscription object, used to unsubscribe from the stream and to get errors
 //   - error: An error, if any
 func (provider *WsProvider) SubscribeEvents(ctx context.Context, events chan<- *EmittedEvent, options *EventSubscriptionInput) (*client.ClientSubscription, error) {
-	if options == nil {
```
MaksymMalicki, line 29: some context on why did you remove this check?

rpc/websocket.go
```diff
@@ -59,13 +50,9 @@ func (provider *WsProvider) SubscribeEvents(ctx context.Context, events chan<- *
 //   - clientSubscription: The client subscription object, used to unsubscribe from the stream and to get errors
 //   - error: An error, if any
 func (provider *WsProvider) SubscribeNewHeads(ctx context.Context, headers chan<- *BlockHeader, blockID BlockID) (*client.ClientSubscription, error) {
-	if blockID == (BlockID{}) {
```
MaksymMalicki, line 62: can you also give some context here?

MaksymMalicki (approved):

rpc/websocket.go
thiagodeev: Sure. This check was necessary because the old version of the BlockID.MarshalJSON() function didn't allow empty BlockID structs. The problem is that for these methods the blockId was optional but the code did not allow us to pass it empty; when it was marshaled to be sent to the server, the BlockID.MarshalJSON() function returned an error saying that it was invalid.

So the workaround to avoid this error was to check if the user passed an empty BlockID (meaning the user didn't specify it), and if so, we set the blockId to "latest" (since the default behavior of blockId for these methods, if not specified, is to be interpreted as "latest"). But now, this issue has been solved.

rpc/websocket.go
thiagodeev: Same as above

PR #697, opened 2025-04-10
WIP: Change FeeEstimation type to use u64 and u128 instead of felt.Felt 

### LINKED ISSUE #690: Change FeeEstimation type to use u64 and u128 instead of felt.Felt
Spec ref: https://github.com/starkware-libs/starknet-specs/blob/master/api/starknet_api_openrpc.json#L3478

Depends on: https://github.com/NethermindEth/starknet.go/issues/709

danielcdz:
## Summary
- Closes #690 

## WIP


thiagodeev:
As stated in the issue, this will not be a priority for now.

PR #699, opened 2025-04-15
Refactor: Rename redundant struct field names

thiagodeev:


rpc/types_transaction_receipt.go
```diff
@@ -30,7 +30,7 @@ type MsgFromL1 struct {
 
 type MessageStatusResp struct {
 	// The hash of a L1 handler transaction
-	TransactionHash *felt.Felt `json:"transaction_hash"`
+	Hash *felt.Felt `json:"transaction_hash"`
```
MaksymMalicki, line 34: I like all the changes, but I'm just wondering, if changing this field here gives enough information about the field

MaksymMalicki (approved):

rpc/types_transaction_receipt.go
thiagodeev: After reading a lot about L1 Messages, I improved the description ✅

PR #700, opened 2025-04-16
Thiagodeev/refactor txn types

thiagodeev:
Summary
The PR significantly refactors transaction types to improve clarity, consistency, and version specificity. Key changes include:
1. Renaming fields for consistency (`TransactionHash` → `Hash`)
2. Restructuring `BlockTransaction` type
3. Enhancing transaction interfaces with specific methods (GetCalldata, GetSenderAddress, GetConstructorCalldata)
4. Removing unused types (`UnknownTransaction`, `TXN`)
5. Simplifying broadcast transaction types through type aliases
6. Adding methods to `TransactionVersion` for better version handling

### Newly Created Types
1. Enhanced Transaction Interfaces:
   * `InvokeTxnType` interface with GetCalldata() []*felt.Felt method - Provides a common interface for all invoke transaction types
   * `DeclareTxnType` interface with GetSenderAddress() *felt.Felt method - Provides a common interface for all declare transaction types
   * `DeployAccountType` interface with GetConstructorCalldata() []*felt.Felt method - Provides a common interface for all deploy account transaction types
2. Type Aliases:
   * `BroadcastInvokeTxnV3 = InvokeTxnV3` - Simplifies the relationship between broadcast and regular transaction types
   * `BroadcastDeployAccountTxnV3 = DeployAccountTxnV3` - Simplifies the relationship between broadcast and regular transaction types

### Deleted Types
1. `UnknownTransaction`:
   * Removed as it was unused in the codebase
2. `TXN`:
   * Removed as it was unused in the codebase
3. `L1HandlerTxnVersion`:
   * Removed and replaced with the more general TransactionVersion type
   * Previously was a string type with only one constant: L1HandlerTxnVersionV0

### Modified Types
1. `BlockTransaction`:
   * Changed from embedding transaction types to using composition
   * Old structure: type BlockTransaction struct { IBlockTransaction }
   * New structure: type BlockTransaction struct { Hash *felt.Felt; Transaction }
   * Added UnmarshalJSON and MarshalJSON methods to handle the new structure
2. `Transaction` Interface:
   * Added GetVersion() TransactionVersion method
   * All transaction types now implement this method
3. `TransactionVersion`:
   * Added `Int()` method that returns the integer version number
   * Enhanced `BigInt()` method to handle versions with query bit
   * Improved error handling for invalid versions
4. `SubPendingTxns` → `PendingTxn`:
   * Renamed to better reflect its purpose
   * Field `TransactionHash` renamed to `Hash`
   * Updated UnmarshalJSON method to work with the new structure
5. `TransactionResponse`:
   * Field `TransactionHash` renamed to `Hash`
   * Same change applied to all response types (AddInvokeTransactionResponse, AddDeclareTransactionResponse, AddDeployAccountTransactionResponse)
6. `DeployAccountTxn` → `DeployAccountTxnV1`:
   * Renamed to explicitly indicate the version
   * No structural changes to the type itself
8. `TransactionReceipt`:
   * Reordered fields for better organization
   * Improved documentation for optional fields
9. Block Types:
   * Changed `Transactions BlockTransactions` to `Transactions []BlockTransaction`
   * Removed `BlockTransactions` type entirely
   * Restructured block-related types for better organization
10. `L1HandlerTxn`:
   * Changed Version field from `L1HandlerTxnVersion` to `TransactionVersion`
   * Made Type and Nonce fields required instead of optional

### Field Changes
1. Hash Field Standardization:
   * All transaction types now use Hash instead of TransactionHash for consistency
   * This affects BlockTransaction, PendingTxn, TransactionResponse, and all response types
2. `BlockTransaction` Structure:
   * Previously embedded different transaction types (BlockInvokeTxnV0, BlockDeclareTxnV2, etc.)
   * Now uses a Transaction field that contains the actual transaction
   * All transaction-specific types (BlockInvokeTxnV0, BlockDeclareTxnV2, etc.) have been removed
3. `Transaction` Interface Implementation:
   * All transaction types now implement both GetType() and GetVersion() methods
   * This provides a consistent way to access type and version information


rpc/types_transaction.go
```diff
@@ -411,9 +285,42 @@ func (v *TransactionVersion) BigInt() (*big.Int, error) {
...
+
+	// Handle versions with query bit.
+	// Remove the 0x prefix and convert to big.Int
+	version, ok := new(big.Int).SetString(string(*v)[2:], 16)
+	if !ok {
 		return big.NewInt(-1), errors.New(fmt.Sprint("TransactionVersion %i not supported", *v))
 	}
+
+	return version, nil
+}
```
rianhughes, line 302: Do we not just want to return `TransactionV0WithQueryBit` as `bit.NewInt(0)`? etc

rianhughes: I see we do this in the `Int()` method actually. When would we ever want `0x100000000000000000000000000000000` as a big int?

account/account.go
```diff
@@ -456,10 +456,10 @@ func (account *Account) TransactionHashDeployAccount(tx rpc.DeployAccountType, c
 
 	// https://docs.starknet.io/architecture-and-concepts/network-architecture/transactions/#deploy_account_transaction
 	switch txn := tx.(type) {
-	// deployAccTxn v0, pointer and struct
-	case *rpc.DeployAccountTxn:
+	// deployAccTxn v1, pointer and struct
+	case *rpc.DeployAccountTxnV1:
 		return hash.TransactionHashDeployAccountV1(txn, contractAddress, account.ChainId)
-	case rpc.DeployAccountTxn:
+	case rpc.DeployAccountTxnV1:
```
rianhughes, line 495: Why do we need to handle both the pointer and value types?

rianhughes (approved):

account/account.go
thiagodeev: Because if it is a pointer, it'll not enter the value case; the opposite is also true.
This method received the `rpc.DeployAccountType` interface as input, and it could be either a txn or a pointer to it.

rpc/types_transaction.go
thiagodeev: I really don't know haha, but I also don't know the reason for the existence of the query bit version, so I just thought that if this version is necessary, it being returned in a bigint as is could be useful to someone.
The `0x0` txn version is indeed `0x100000000000000000000000000000000` as query bit ([ref](https://github.com/starkware-libs/starknet-specs/blob/f0ae2e919c2564ea9028d7564bccdba6f3badcad/api/starknet_api_openrpc.json#L1949))

PR #714, opened 2025-04-29
Hotfix/not estimating querybitversion txns

thiagodeev:
This PR aims to add a new optional parameter called `withQueryBitVersion` to the `BuildAndEstimateDeployAccountTxn`, `BuildAndSendInvokeTxn`, and `BuildAndSendDeclareTxn` methods, to allow the user to choose whether they want the transaction version to be in standard format (e.g. `0x3`) or in queryBit format (e.g. `0x100000000000000000000000000000003`)

### DIFF
--- a/account/account.go
+++ b/account/account.go
@@ -22,9 +22,9 @@ var (
 
 //go:generate mockgen -destination=../mocks/mock_account.go -package=mocks -source=account.go AccountInterface
 type AccountInterface interface {
-	BuildAndEstimateDeployAccountTxn(ctx context.Context, salt *felt.Felt, classHash *felt.Felt, constructorCalldata []*felt.Felt, multiplier float64, hasQueryBitVersion ...bool) (*rpc.BroadcastDeployAccountTxnV3, *felt.Felt, error)
-	BuildAndSendInvokeTxn(ctx context.Context, functionCalls []rpc.InvokeFunctionCall, multiplier float64, hasQueryBitVersion ...bool) (*rpc.AddInvokeTransactionResponse, error)
-	BuildAndSendDeclareTxn(ctx context.Context, casmClass *contracts.CasmClass, contractClass *contracts.ContractClass, multiplier float64, hasQueryBitVersion ...bool) (*rpc.AddDeclareTransactionResponse, error)
+	BuildAndEstimateDeployAccountTxn(ctx context.Context, salt *felt.Felt, classHash *felt.Felt, constructorCalldata []*felt.Felt, multiplier float64, withQueryBitVersion bool) (*rpc.BroadcastDeployAccountTxnV3, *felt.Felt, error)
+	BuildAndSendInvokeTxn(ctx context.Context, functionCalls []rpc.InvokeFunctionCall, multiplier float64, withQueryBitVersion bool) (*rpc.AddInvokeTransactionResponse, error)
+	BuildAndSendDeclareTxn(ctx context.Context, casmClass *contracts.CasmClass, contractClass *contracts.ContractClass, multiplier float64, withQueryBitVersion bool) (*rpc.AddDeclareTransactionResponse, error)
 	Nonce(ctx context.Context) (*felt.Felt, error)
 	SendTransaction(ctx context.Context, txn rpc.BroadcastTxn) (*rpc.TransactionResponse, error)
 	Sign(ctx context.Context, msg *felt.Felt) ([]*felt.Felt, error)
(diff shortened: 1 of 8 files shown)


account/account.go
```diff
@@ -133,6 +137,11 @@ func (account *Account) BuildAndSendInvokeTxn(ctx context.Context, functionCalls
 
 	// building and signing the txn, as it needs a signature to estimate the fee
 	broadcastInvokeTxnV3 := utils.BuildInvokeTxn(account.Address, nonce, callData, makeResourceBoundsMapWithZeroValues())
+
+	if len(hasQueryBitVersion) > 0 && hasQueryBitVersion[0] {
+		broadcastInvokeTxnV3.InvokeTxnV3.Version = rpc.TransactionV3WithQueryBit
+	}
```
hudem1: should we check it is exactly of length 1 ? or not necessary ?

thiagodeev: Not necessary IMO. The description says it can be `true` or `false`, defaulting to `false` if omitted.
If the user put `true, true, true...`, it will just check the first and ignore the rest

hudem1: yeah okok

hudem1 (approved):

account/account.go
```diff
@@ -180,6 +193,7 @@ func (account *Account) BuildAndSendDeclareTxn(
 	casmClass *contracts.CasmClass,
 	contractClass *contracts.ContractClass,
 	multiplier float64,
+	hasQueryBitVersion ...bool,
```
rodrodros: Why it is set as a slice instead of just on value `bool`?

The user shouldn't have an option of not providing it in the funciton. It makes the code hard to read for two reasons:
1. You are passing a slice when you only use a single value, then just a single value should be passed.
2. The user by ignoring it is taking a decision he didn't explicitly commit to. 

I assume that the reason behind this is not to break retro-compatibility but in this case it's hurting the project long term. 

Also, I would change the variable name to `withQueryBitVersion`

account/account.go
```diff
@@ -171,6 +180,10 @@ func (account *Account) BuildAndSendInvokeTxn(ctx context.Context, functionCalls
 //     fee fluctuations. It multiplies both the max amount and max price per unit by this value.
 //     A value of 1.5 (50% buffer) is recommended to balance between transaction success rate and
 //     avoiding excessive fees. Higher values provide more safety margin but may result in overpayment.
+//   - hasQueryBitVersion: A optional boolean flag indicating whether the transaction version should have the query bit.
+//     If true, the transaction version will be rpc.TransactionV3WithQueryBit (0x100000000000000000000000000000003).
+//     If false, the transaction version will be rpc.TransactionV3 (0x3).
+//     If not provided, the default is false.
```
rodrodros: Great description!!!

account/account.go
thiagodeev: Yes, it's also for retro-compatibility, but the other reason is that most users will never need to use this query bit version, and it being a variadic parameter, it could be ignored seamlessly. WDYT?

rodrodros: I don't think it should be variadic, that it could be ignored seamlessly is bad to my point of view. 

Is best for code safety and readability to have explicit code. Ignoring is easy now but a pain later, also you are using variadic parameters to solve some problem which is not the intended use. It causes confusion.

In another programming language you could use optional parameters but this is not available in Go. You shouldn't proceed with this. If adding an explict parameter is not he best, then probably the code require refactoring to accommodate this new feature. I am not familiar with the code enough to offer alternative suggestion.

I don't think you should go with this solution, solves the problem now by degrading code quality.

thiagodeev: Ok, I'll change it then. Thanks for the suggestion!

account/account.go
```diff
@@ -116,11 +116,15 @@ func (account *Account) Nonce(ctx context.Context) (*felt.Felt, error) {
...
+//   - withQueryBitVersion: A boolean flag indicating whether the transaction version should have the query bit.
+//     If true, the transaction version will be rpc.TransactionV3WithQueryBit (0x100000000000000000000000000000003).
+//     If false, the transaction version will be rpc.TransactionV3 (0x3).
+//     In case of doubt, set to 'false'.
 //
 // Returns:
 //   - *rpc.AddInvokeTransactionResponse: the response of the submitted transaction.
 //   - error: An error if the transaction building fails.
-func (account *Account) BuildAndSendInvokeTxn(ctx context.Context, functionCalls []rpc.InvokeFunctionCall, multiplier float64) (*rpc.AddInvokeTransactionResponse, error) {
+func (account *Account) BuildAndSendInvokeTxn(ctx context.Context, functionCalls []rpc.InvokeFunctionCall, multiplier float64, withQueryBitVersion bool) (*rpc.AddInvokeTransactionResponse, error) {
```
rodrodros: Just a nitpick, this line is too long and makes it hard to read. Suggestion:
```go
func (account *Account) BuildAndSendInvokeTxn(
    ctx context.Context, 
    functionCalls []rpc.InvokeFunctionCall,
    multiplier float64, 
    withQueryBitVersion bool,
) (*rpc.AddInvokeTransactionResponse, error) {

rodrodros (approved):
LGTM

Issue #711, opened 2025-04-25
Warning when using incompatible RPC provider


thiagodeev:
Add a warning when using incompatible RPC provider (between the node and SDK) for UX improvement.

Ayoseun:
Hi, is this issue still open.

thiagodeev:
Hi @Ayoseun, thank you for your interest.
Sorry, but this issue is being handled internally

Ayoseun:
Thanks

thiagodeev:
Merged in #725

PR #725, opened 2025-05-05
Feature/version compatibility check

RafieAmandio:
Solves https://github.com/NethermindEth/starknet.go/issues/711

New version compatibility check to prevent API mismatches. It returns a warning if a provider implementing a different starknet RPC version than Starknet.go is instantiated with `rpc.NewProvider`


rpc/version.go
```diff
@@ -13,3 +13,17 @@ func (provider *Provider) SpecVersion(ctx context.Context) (string, error) {
 	}
 	return result, nil
 }
+
+// Version represents the current version of the Starknet RPC client.
+type Version struct {
```
thiagodeev: Why this? It's not even being used. Remove it please

rpc/version.go
```diff
@@ -13,3 +13,17 @@ func (provider *Provider) SpecVersion(ctx context.Context) (string, error) {
...
+	Major int
+	// The minor version number.
+	Minor int
+	// The patch version number.
+	Patch int
+}
+
+// SDKVersion is the current version of the Starknet Go SDK.
+// This should be updated when releasing new versions of the SDK.
+const SDKVersion = "0.8.1"
```
thiagodeev: Let's rename that. This won't be the SDK version, but the RPC version that the SDK is using.
Rename it to something like 'RPCVersion', and move it to the top of the `provider.go` file

rpc/version_compatibility.go
```diff
@@ -0,0 +1,52 @@
+package rpc
+
+import (
+	"fmt"
+	"strings"
+)
+
+// VersionCompatibilityError represents an error when the RPC provider version is incompatible with the SDK
+type VersionCompatibilityError struct {
```
thiagodeev: I believe that creating a new Struct type just to log an error message is overkill. Let's remove it

rpc/version_compatibility.go
```diff
@@ -0,0 +1,52 @@
...
+
+// CheckVersionCompatibility checks if the RPC provider version is compatible with the SDK version
+//
+// Parameters:
+//   - providerVersion: The version string returned by the RPC provider
+//   - sdkVersion: The version string of the SDK
+//
+// Returns:
+//   - error: A VersionCompatibilityError if versions are incompatible, nil otherwise
+func CheckVersionCompatibility(providerVersion, sdkVersion string) error {
```
thiagodeev: 1. We don't need to create a new file only for this; it can be a helper function in the `provider.go` file. Also, make it internal (starting with lowercase), since we don't need to export it to the user.
2. No need to compare only the major or minor, let's compare the entire string and return a warning if they don't match. We can use the `strings.Contains()` for that, to avoid the case of mismatch between '0.8.1' and 'v0.8.1', for example

rpc/version_compatibility_test.go
```diff
@@ -0,0 +1,86 @@
...
+				_, ok := err.(*VersionCompatibilityError)
+				require.True(t, ok, "error should be of type VersionCompatibilityError")
+			} else {
+				require.NoError(t, err)
+			}
+		})
+	}
+}
+
+func TestVersionCompatibilityError(t *testing.T) {
```
thiagodeev: This test does basically the same as the test above. Let's remove it

rpc/version_compatibility_test.go
```diff
@@ -0,0 +1,86 @@
+package rpc
+
+import (
+	"testing"
+
+	"github.com/stretchr/testify/require"
+)
+
+// TestCheckVersionCompatibility tests the version compatibility check functionality
+func TestCheckVersionCompatibility(t *testing.T) {
```
thiagodeev: Instead of just having this unit test, create a new integration test that creates a new provider and checks if no errors occur

rpc/provider.go
```diff
@@ -68,7 +85,27 @@ func NewWebsocketProvider(url string, options ...client.ClientOption) (*WsProvid
 		return nil, err
 	}
 
-	return &WsProvider{c: c}, nil
+	provider := &WsProvider{c: c}
```
thiagodeev: Let's remove the check for the `NewWebsocketProvider` method. It probably will fail because the `SpecVersion` method is an HTTP call instead of WS

rpc/provider.go
```diff
@@ -49,7 +50,23 @@ func NewProvider(url string, options ...client.ClientOption) (*Provider, error)
 		return nil, err
 	}
 
-	return &Provider{c: c}, nil
+	provider := &Provider{c: c}
+
+	// Check version compatibility
+	version, err := provider.SpecVersion(context.Background())
```
thiagodeev: Wrap this inside the `checkVersionCompatibility` helper, so that all steps necessary for checking are inside this function

thiagodeev (changes requested):
Hey @RafieAmandio, congrats on your first PR here!
I've requested some changes

rpc/version.go
```diff
@@ -14,3 +14,7 @@ func (provider *Provider) SpecVersion(ctx context.Context) (string, error) {
 
 	return result, nil
 }
+
+// RPCVersion is the version of the Starknet JSON-RPC specification that this SDK is compatible with.
+// This should be updated when supporting new versions of the RPC specification.
+const RPCVersion = "0.8.1"
```
thiagodeev: Move this to the top of the 'provider.go' file

rpc/version_test.go
```diff
@@ -29,3 +30,23 @@ func TestSpecVersion(t *testing.T) {
 		require.Equal(t, test.ExpectedResp, resp)
 	}
 }
+
+// TestVersionCompatibility tests that the provider correctly handles version compatibility
+func TestVersionCompatibility(t *testing.T) {
```
thiagodeev: This test does almost the same thing as the `TestSpecVersion` above (call `SpecVersion` and compare the result), so it's useless.
What we really need to test is if there's a warning coming from the 'NewProvider' function when we pass a URL of a provider that returns a different Spec Version than the one we defined at the `RPCVersion` const, right? That's the purpose of this PR.

Look at the `NewAccount` function [here](https://github.com/NethermindEth/starknet.go/blob/fd8a5ef75080e78212c852c8938bf40967a3c393/account/account.go#L87). I created a warning to be displayed when users instantiate a Braavos account (very similar to this new feature).
Look at the test for this functionality [here](https://github.com/NethermindEth/starknet.go/blob/fd8a5ef75080e78212c852c8938bf40967a3c393/account/account_test.go#L1501). In this test, we read all the data being written to stdout and check if the warning is present. Do something like this. Check if the warning is being returned and not returned when it should be.

Also, move this test to the 'provider_test.go'

rpc/provider.go
thiagodeev: This is still unresolved from the last review.
What I mean is: why are we calling the `SpecVersion` here? Isn't it to check the version compatibility? Since that's the case, it should then be part of the `checkVersionCompatibility` helper.
We should wrap all related code into its specific functions.

thiagodeev (changes requested):

rpc/version_test.go
RafieAmandio: done

rpc/version.go
RafieAmandio: done

thiagodeev: > done

Actually, not. You put it at the top of the same file, not in the 'provider.go' file

RafieAmandio: My apologies, I misread the instruction, I have move it to the correct file now.

rpc/version_test.go
thiagodeev, line 1: Why did you remove this file?

rpc/provider.go
```diff
@@ -18,6 +20,30 @@ var (
 	errNotFound = errors.New("not found")
 )
 
+// RPCVersion is the version of the Starknet JSON-RPC specification that this SDK is compatible with.
+// This should be updated when supporting new versions of the RPC specification.
+const RPCVersion = "0.8.1"
+
+// checkVersionCompatibility checks if the RPC provider version is compatible with the SDK version
+// and returns a warning if they don't match.
+func checkVersionCompatibility(provider *Provider) error {
```
thiagodeev: There's no reason to return an error since our goal is just to display a warning.
Make this function return nothing, and print the warning inside it.

rpc/provider.go
```diff
@@ -18,6 +20,30 @@ var (
...
+	version, err := provider.SpecVersion(context.Background())
+	if err != nil {
+		// Return a warning but don't fail
+		return fmt.Errorf("warning: Could not check RPC version compatibility: %v", err)
+	}
+
+	if !strings.Contains(version, RPCVersion) {
+		return fmt.Errorf(
+			"warning: RPC provider version %s is different from expected version %s. "+
```
thiagodeev: Make the warning message a const variable alongside the `RPCVersion` for being used multiple times (in the tests, for example)

rpc/provider.go
```diff
@@ -48,7 +74,15 @@ func NewProvider(url string, options ...client.ClientOption) (*Provider, error)
 		return nil, err
 	}
 
-	return &Provider{c: c}, nil //nolint:exhaustruct
+	provider := &Provider{c: c, chainID: ""}
+
+	// Check version compatibility
+	if err := checkVersionCompatibility(provider); err != nil {
```
thiagodeev: Making the required changes, this will be just 
`checkVersionCompatibility(provider)`
instead of 
`if err := checkVersionCompatibility(provider); err != nil {...`

rpc/provider_test.go
```diff
@@ -15,6 +18,8 @@ import (
 
 const (
 	DevNetETHAddress = "0x49d36570d4e46f48e99674bd3fcc84644ddd6b96f7c741b1562b82f9e004dc7"
+	// specVersionMethod is the RPC method name for fetching spec version
+	specVersionMethod = "starknet_specVersion"
```
thiagodeev: This variable is only used inside the `TestVersionCompatibility` test, so we can move it inside the test

rpc/provider_test.go
```diff
@@ -156,3 +188,128 @@ func TestCookieManagement(t *testing.T) {
...
+				"result":  "0.5.0", // Different version
+			}
+			if err := json.NewEncoder(w).Encode(data); err != nil {
+				log.Fatal(err)
+			}
+		}
+	}))
+	defer incompatibleServer.Close()
+
+	errorServer := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
```
thiagodeev: One server for each test case is overkill. Find a way to unify them in a single http server, and include conditions inside it to define the server's response (e.g. using query parameters, etc)

thiagodeev (changes requested):
Nice work! The new `TestVersionCompatibility` test is working as we expected, it's checking if the warning is being displayed or not.
There are some things to be changed, otherwise looks good to merge

rpc/version_test.go
RafieAmandio: I apologize for the confusion. I mistakenly thought the previous comment applied to all the tests. I've addressed this in the latest version

rpc/provider.go
RafieAmandio: Done, the  code is already not returning any error and implement this

rpc/provider.go
RafieAmandio: done

rpc/provider.go
RafieAmandio: done

rpc/provider_test.go
RafieAmandio: done, already moved inside the test

thiagodeev (approved):

Issue #724, opened 2025-05-04
Change BuildTx methods to allow `tip` specification


thiagodeev:
Currently, the v3 txn `tip` field is not being used, the doc says the current value for it is 0 ([ref](https://docs.starknet.io/architecture-and-concepts/network-architecture/transactions/#v3_transaction_fields))

Starknet v0.14.0 will change that, users will be able to specify a tip for a txn, so we need to change the current `BuildTxn...` utilities and methods to allow them specify it, as we currently set it to 0 automatically

pheobeayo:
May I take care of this?

thiagodeev:
Hi @pheobeayo! Thanks for your interest, but we are already working on this issue internally. Sorry for forgetting to update it

PR #729, opened 2025-05-12
New `TxnOptions` when building/sending txns

RafieAmandio:
Solves https://github.com/NethermindEth/starknet.go/issues/724

Thiagodeev:
## Key changes:
- New `CairoVersion` type in the `account` pkg.
- A new `TxnOptions` in the `account` pkg to be used in the `BuildAndEstimateDeployAccountTxn`, `BuildAndSendInvokeTxn`, and `BuildAndSendDeclareTxn` methods.
- A new `TxnOptions` in the `utils` pkg to be used in the `BuildInvokeTxn`, `BuildDeclareTxn`, and `BuildEstimateDeployAccountTxn` functions.


utils/transactions.go
```diff
@@ -13,13 +13,50 @@ import (
 	"github.com/NethermindEth/starknet.go/rpc"
 )
 
+const maxU64 uint64 = 0xFFFFFFFFFFFFFFFF
```
thiagodeev: Why change this? I prefer to use `math.MaxUint64` as it is completely reliable.

However, I liked the idea of being a `const`. Please change the other variables below to `const` as well

account/account.go
```diff
@@ -141,24 +138,20 @@ func (account *Account) Nonce(ctx context.Context) (*felt.Felt, error) {
...
 //     multiple function calls in the same transaction.
-//   - multiplier: A safety factor for fee estimation that helps prevent transaction failures due to
-//     fee fluctuations. It multiplies both the max amount and max price per unit by this value.
-//     A value of 1.5 (50% buffer) is recommended to balance between transaction success rate and
-//     avoiding excessive fees. Higher values provide more safety margin but may result in overpayment.
-//   - withQueryBitVersion: A boolean flag indicating whether the transaction version should have the query bit.
-//     If true, the transaction version will be rpc.TransactionV3WithQueryBit (0x100000000000000000000000000000003).
-//     If false, the transaction version will be rpc.TransactionV3 (0x3).
-//     In case of doubt, set to 'false'.
+//   - opts: TransactionOptions containing options for building the transaction
```
thiagodeev: Sorry. Now I see that the `multiplier` should not be part of the `TransactionOptions`, it's a field related to these "BuildAndSend..." methods. So let's put the multiplier back here

account/account.go
```diff
@@ -141,24 +138,20 @@ func (account *Account) Nonce(ctx context.Context) (*felt.Felt, error) {
...
 //   - *rpc.AddInvokeTransactionResponse: the response of the submitted transaction.
 //   - error: An error if the transaction building fails.
 func (account *Account) BuildAndSendInvokeTxn(
 	ctx context.Context,
 	functionCalls []rpc.InvokeFunctionCall,
-	multiplier float64,
-	withQueryBitVersion bool,
+	opts *utils.TransactionOptions,
 ) (*rpc.AddInvokeTransactionResponse, error) {
+	if opts == nil {
```
thiagodeev: We don't need this validation here. Let's just pass the `opts` as is to the "Build...Txn" function (`BuildInvokeTxn` in this case) and validate there.

account/account.go
```diff
@@ -169,10 +162,9 @@ func (account *Account) BuildAndSendInvokeTxn(
 		return nil, err
 	}
 
-	// building and signing the txn, as it needs a signature to estimate the fee
-	broadcastInvokeTxnV3 := utils.BuildInvokeTxn(account.Address, nonce, callData, makeResourceBoundsMapWithZeroValues())
+	broadcastInvokeTxnV3 := utils.BuildInvokeTxn(account.Address, nonce, callData, makeResourceBoundsMapWithZeroValues(), opts)
 
-	if withQueryBitVersion {
+	if opts.WithQueryBitVersion {
```
thiagodeev: Let's remove this from all the "BuildAndSend..." methods, we will do this in the "Build...Txn" utilities

account/account.go
```diff
@@ -215,14 +206,7 @@ func (account *Account) BuildAndSendInvokeTxn(
...
 //   - contractClass: The sierra contract class of the contract to be declared
-//   - multiplier: A safety factor for fee estimation that helps prevent transaction failures due to
-//     fee fluctuations. It multiplies both the max amount and max price per unit by this value.
-//     A value of 1.5 (50% buffer) is recommended to balance between transaction success rate and
-//     avoiding excessive fees. Higher values provide more safety margin but may result in overpayment.
-//   - withQueryBitVersion: A boolean flag indicating whether the transaction version should have the query bit.
-//     If true, the transaction version will be rpc.TransactionV3WithQueryBit (0x100000000000000000000000000000003).
-//     If false, the transaction version will be rpc.TransactionV3 (0x3).
-//     In case of doubt, set to 'false'.
+//   - opts: TransactionOptions containing options for building the transaction
```
thiagodeev: Add something like "See more info in the TransactionOptions type description" to these descriptions

utils/transactions.go
```diff
@@ -13,13 +13,50 @@ import (
...
-	negativeResourceBoundsErr        = "resource bounds cannot be negative, got '%#x'"
-	invalidResourceBoundsErr         = "invalid resource bounds: '%v' is not a valid big.Int"
+	maxUint128                = "0xffffffffffffffffffffffffffffffff"
+	negativeResourceBoundsErr = "resource bounds cannot be negative, got '%#x'"
+	invalidResourceBoundsErr  = "invalid resource bounds: '%v' is not a valid big.Int"
 )
 
+// TransactionOptions holds options for building transactions
+// Multiplier: safety factor for fee estimation
+// WithQueryBitVersion: whether to use the query bit version
```
thiagodeev: Improve the queryBit description. Take it from the `account.BuildAndSendInvokeTxn` description you removed

utils/transactions.go
```diff
@@ -13,13 +13,50 @@ import (
...
+// Tip: tip amount for the transaction
+type TransactionOptions struct {
```
thiagodeev: Obs: rename to `TxnOptions`

utils/transactions.go
```diff
@@ -13,13 +13,50 @@ import (
...
+	Multiplier          float64
+	WithQueryBitVersion bool
+	Tip                 rpc.U64
+}
+
+// validate sets defaults and checks for edge cases
+func (opts *TransactionOptions) validate() {
```
thiagodeev: Let's rename it to `ApplyOptions()`, and the child methods to `applyTip()`...

utils/transactions.go
```diff
@@ -13,13 +13,50 @@ import (
...
+	invalidResourceBoundsErr  = "invalid resource bounds: '%v' is not a valid big.Int"
 )
 
+// TransactionOptions holds options for building transactions
+// Multiplier: safety factor for fee estimation
+// WithQueryBitVersion: whether to use the query bit version
+// Tip: tip amount for the transaction
+type TransactionOptions struct {
+	Multiplier          float64
+	WithQueryBitVersion bool
```
thiagodeev: Create a method called `TxnVersion` that will return the corresponding txn version based on the `WithQueryBitVersion` value. We will use this method for populate the `Version` field of the transactions, just like you did here: `Tip: opts.Tip,`. We'll do something like `Version: opts.TxnVersion()`

utils/transactions.go
```diff
@@ -13,13 +13,50 @@ import (
...
+
+func (opts *TransactionOptions) validateMultiplier() {
+	if opts.Multiplier <= 0 {
+		opts.Multiplier = 1.5
+	}
+}
+
+func (opts *TransactionOptions) validateTip() {
+	if opts.Tip == "" {
+		opts.Tip = "0x0"
```
thiagodeev: Since it was set to "0x0", we can already call `return` and leave the function instead of falling on each one of the remaining `if`s, right?

utils/transactions.go
```diff
@@ -13,13 +13,50 @@ import (
...
+	}
+	if !strings.HasPrefix(string(opts.Tip), "0x") {
+		opts.Tip = "0x0"
```
thiagodeev: Let's remove this condition. The `Tip.ToUint64()` will check if it has the "0x" prefix or not

utils/transactions.go
```diff
@@ -13,13 +13,50 @@ import (
...
+	}
+	if val, err := opts.Tip.ToUint64(); err != nil {
+		opts.Tip = "0x0"
+	} else if val > maxU64 {
+		opts.Tip = "0xFFFFFFFFFFFFFFFF" // max U64
```
thiagodeev: I believe this will never be reached, the `Tip.ToUint64()` will check for overflows. We should check this in a test

thiagodeev (changes requested):
Good work until now! Good use of the `U64.ToUint64()` method.
I requested some changes

utils/transactions.go
RafieAmandio: done

account/account.go
RafieAmandio: done

account/account.go
RafieAmandio: done, but the multiplier still inside the transactionOptions as discussed on the meeting

account/account.go
RafieAmandio: done

utils/transactions.go
RafieAmandio: done

utils/transactions.go
RafieAmandio: done

utils/transactions.go
RafieAmandio: done

utils/transactions.go
RafieAmandio: done

utils/transactions.go
RafieAmandio: done

thiagodeev (approved):

account/account.go
```diff
@@ -24,21 +26,18 @@ type AccountInterface interface {
 		salt *felt.Felt,
 		classHash *felt.Felt,
 		constructorCalldata []*felt.Felt,
-		multiplier float64,
-		withQueryBitVersion bool,
+		opts *TxnOptions,
```
rodrodros, line 29: 👍

account/account.go
```diff
@@ -63,12 +62,20 @@ type Account struct {
 	ChainId      *felt.Felt
 	Address      *felt.Felt
 	publicKey    string
-	CairoVersion int
+	CairoVersion CairoVersion
```
rodrodros: 👍

account/transaction.go
```diff
@@ -18,23 +18,15 @@ import (
...
 //     multiple function calls in the same transaction.
-//   - multiplier: A safety factor for fee estimation that helps prevent transaction failures due to
-//     fee fluctuations. It multiplies both the max amount and max price per unit by this value.
-//     A value of 1.5 (50% buffer) is recommended to balance between transaction success rate and
-//     avoiding excessive fees. Higher values provide more safety margin but may result in overpayment.
-//   - withQueryBitVersion: A boolean flag indicating whether the transaction version should have the query bit when estimating fees.
-//     If true, the transaction version will be rpc.TransactionV3WithQueryBit (0x100000000000000000000000000000003).
-//     If false, the transaction version will be rpc.TransactionV3 (0x3).
-//     In case of doubt, set to 'false'.
+//   - opts: options for building/estimating the transaction. See more info in the TxnOptions type description.
```
rodrodros, line 21: This is very subjective but I would remove the second sentence because anyone looking at the code would do this implicitly and it's harder to maintain overall, but this is also me being very extra nitpicky, so feel free to dismiss.

account/transaction.go
```diff
@@ -46,14 +38,22 @@ func (account *Account) BuildAndSendInvokeTxn(
...
 
-	// building and signing the txn, as it needs a signature to estimate the fee
-	broadcastInvokeTxnV3 := utils.BuildInvokeTxn(account.Address, nonce, callData, makeResourceBoundsMapWithZeroValues())
-
-	if withQueryBitVersion {
-		// the query bit txn version is used for custom validation logic from wallets/accounts when estimating fee/simulating txns
-		broadcastInvokeTxnV3.Version = rpc.TransactionV3WithQueryBit
+	if opts == nil {
+		opts = &TxnOptions{} //nolint:exhaustruct
 	}
```
rodrodros: I think `TxnOptions` should have a constructor for defaults and it should be used here. The constructor should also return by value because you don't want to create a heap allocation just for that, I would go for something like:

```
if opts == nil {
   default := makeDefaultOpts()
   opts = &default
}
```

Ignoring the linter is usually considered a red flag and it means there is room for a better design (90% of the time in my opinion)

account/transaction.go
```diff
@@ -46,14 +38,22 @@ func (account *Account) BuildAndSendInvokeTxn(
...
+	// building and signing the txn, as it needs a signature to estimate the fee
+	broadcastInvokeTxnV3 := utils.BuildInvokeTxn(
+		account.Address,
+		nonce,
+		callData,
+		makeResourceBoundsMapWithZeroValues(),
+		&utils.TxnOptions{
+			Tip:                 opts.Tip,
+			WithQueryBitVersion: opts.WithQueryBitVersion,
+		},
```
rodrodros: Why are just some field of `opts`  used here?

account/transaction.go
```diff
@@ -112,31 +105,33 @@ func (account *Account) BuildAndSendDeclareTxn(
...
+	opts *TxnOptions,
 ) (*rpc.AddDeclareTransactionResponse, error) {
 	nonce, err := account.Nonce(ctx)
 	if err != nil {
 		return nil, err
 	}
 
+	if opts == nil {
+		opts = &TxnOptions{} //nolint:exhaustruct
+	}
```
rodrodros: Similar comment to before.

Question, does it makes sense for people not to pass opts, why don't you work on the assumption that the argument it's always there.

account/transaction.go
```diff
@@ -201,26 +189,27 @@ func (account *Account) BuildAndEstimateDeployAccountTxn(
 	salt *felt.Felt,
 	classHash *felt.Felt,
 	constructorCalldata []*felt.Felt,
-	multiplier float64,
-	withQueryBitVersion bool,
+	opts *TxnOptions,
 ) (*rpc.BroadcastDeployAccountTxnV3, *felt.Felt, error) {
+	if opts == nil {
+		opts = &TxnOptions{} //nolint:exhaustruct
```
rodrodros: Marking this part of the code to not forget (similar to previous comments)

account/txn_options.go
```diff
@@ -0,0 +1,60 @@
...
+}
+
+// SafeMultiplier returns the multiplier for the transaction. If the multiplier is not set or negative, returns 1.5.
+func (opts *TxnOptions) SafeMultiplier() float64 {
+	if opts == nil || opts.Multiplier <= 0 {
+		return 1.5 //nolint:mnd
+	}
+
+	return opts.Multiplier
+}
```
rodrodros: Why would be the use case of this?

account/txn_options.go
```diff
@@ -0,0 +1,60 @@
...
+	WithQueryBitVersion bool
+
+	// A safety factor for fee estimation that helps prevent transaction
+	// failures due to fee fluctuations. It multiplies both the max amount
+	// and max price per unit by this value.
+	// A value of 1.5 (estimated fee + 50%) is recommended to balance between
+	// transaction success rate and avoiding excessive fees. Higher values
+	// provide more safety margin but may result in overpayment.
+	// If multiplier <= 0, it'll be set to 1.5. Default: `1.5`.
+	Multiplier float64
```
rodrodros, line 25: The default is not 1.5, the default is zero 🤔 , you are assuming people will access this property through `SafeMultiplier` but there is no obligation for that.

account/txn_options.go
rodrodros, line 1: Please remove the checks from this file that check if `opts` is nil, that doesn't make sense. If `opts` is nil then someone is using the code wrong.

account/txn_options.go
```diff
@@ -0,0 +1,60 @@
...
+	if opts == nil || opts.EstimationBlockTag == "" {
+		return rpc.WithBlockTag(rpc.BlockTagPending)
+	}
+
+	return rpc.WithBlockTag(opts.EstimationBlockTag)
+}
+
+// Returns a `[]rpc.SimulationFlag` containing the SimulationFlag.
+// If the flag is not set, returns an empty slice.
+func (opts *TxnOptions) SimulationFlags() []rpc.SimulationFlag {
```
rodrodros, line 47: I am not sure I understand why you return a slice if the value is always going to be either a single or null value. You should return a pointer I think?

```
func (opts *TxnOptions) SimulationFlags() *rpc.SimulationFlag {
     if opts.SimulationFlag == "" {
         return nil 
      }
      return &opts.SimulationFlag
}

File: utils/transactions.go [hunk 23]
```diff
@@ -13,44 +13,77 @@ import (
...
-var (
+const (
 	maxUint64                 uint64 = math.MaxUint64
 	maxUint128                       = "0xffffffffffffffffffffffffffffffff"
 	negativeResourceBoundsErr        = "resource bounds cannot be negative, got '%#x'"
 	invalidResourceBoundsErr         = "invalid resource bounds: '%v' is not a valid big.Int"
 )
 
+// Optional settings when building a transaction.
+type TxnOptions struct {
```
rodrodros, line 24: This is a bit confusing, why there are two types with the same name?

account/transaction.go
thiagodeev: I'll study more about heap allocation to get this.
This does not make any difference with my current knowledge :sweat_smile:

``` go
   default := makeDefaultOpts()
   opts = &default
```

utils/transactions.go
thiagodeev: Because the txnOptions use-cases in the `account` and `utils`pkg are different.
The `utils.Build...` functions are only made to build a txn, but the `account.BuildAnd...` methods do more than that, they will send and estimate the txn.

When simply building a txn, we don't need to know the fee multiplier, nor how it will be estimated. The required values are `queryBit` and `tip`. Since the requirements are different for each pkg, I've created different options for both. So the `account.TxnOptions` has all the fields from the `utils.TxnOptions` + the ones used when estimating/sending, and then, inside the functions, we just pass the `tip` and `queryBit` values to the `utils.Build..` functions, letting them validate it. 

That way, we decouple the `utils.Build....` from the `account`, making it easier for the users who just want to build a raw txn with it.

account/transaction.go
thiagodeev: Related to https://github.com/NethermindEth/starknet.go/pull/729#discussion_r2129213376
The values required for the `utils.Build...` functions are only `tip` and `queryBit`.

account/transaction.go
thiagodeev: Yes, it makes 🤔
They are all optional fields with default values, so if the user just wants to use the default options, passing `nil` is the easiest form.

account/txn_options.go
thiagodeev: Passing a `nil` options, or by omitting the `opts.Multiplier` field when declaring the `TxnOptions` field, either by forgetting or by wanting to use the default value

account/txn_options.go
thiagodeev: Yes, but in the `account.BuildAnd....` methods, the `Multiplier` is only accessed by the `SafeMultiplier()` method, and the method validates whether it's 0 and sets it to 1.5. Get it?
The 'default' values are defined by the methods, because the raw values of the `TxnOptions` are never used inside the `Build...` functions.

account/txn_options.go
thiagodeev: Related https://github.com/NethermindEth/starknet.go/pull/729#discussion_r2129222065
`nil` is a valid value to a user who wants to use all the default values for the txn

account/txn_options.go
thiagodeev: Because the output of the `SimulationFlags()` method goes directly to the `EstimateFee()` method, which accepts a   `[]rpc.SimulationFlag` type. So returning a slice is simpler than something like:

``` go
[]rpc.SimulationFlag{opts.SafeSimulationFlag()},
```

There's also a little difference between returning an empty slice and a slice with a "" value

account/transaction.go
rodrodros: This article might serve as a an introduction:
* https://medium.com/@pranoy1998k/understanding-escape-analysis-in-go-b2db76be58f0

account/transaction.go
rodrodros: It can be the easiest, but I am not sure it's the correct one. We can discuss this more during the daily

account/txn_options.go
rodrodros: Is only accessed by the `SafeMultiplier` in your code, at the same time is part of the public API of Starknet.go and people can use Multiplier directly

account/txn_options.go
```diff
@@ -0,0 +1,118 @@
...
+
+func defaultTxnOpts() *txnOptions {
+	return &txnOptions{
+		tip:                 rpc.U64("0x0"),
+		withQueryBitVersion: false,
+		multiplier:          1.5,
+		usePending:          true,
+		simulationFlag:      "",
+	}
+}
```
rodrodros: Return by value

account/txn_options.go
```diff
@@ -0,0 +1,118 @@
...
+	Tip                 rpc.U64
+	WithQueryBitVersion bool
+	Multiplier          float64
+	BlockID             rpc.BlockID
+	SimulationFlags     []rpc.SimulationFlag
+}
+
+func NewTxnOpts() *txnOptions {
+	return defaultTxnOpts()
+}
```
rodrodros: Why two constructors?

account/txn_options.go
```diff
@@ -0,0 +1,118 @@
...
+	}
+}
+
+type fmtOptions struct {
+	Tip                 rpc.U64
+	WithQueryBitVersion bool
+	Multiplier          float64
+	BlockID             rpc.BlockID
+	SimulationFlags     []rpc.SimulationFlag
+}
```
rodrodros: Do we need this type? It seems to be just the same as `txnOptions`

account/txn_options.go
rodrodros, line 1: I believe this new approach is better but still a bit over complicated. I think you should only need one type, `TxnOptions` can have all their fields public + constructor and no setter methods. The `fmtOptions` doesn't make sense to me right now

account/txn_options.go
thiagodeev: It's an internal type for easy use of the options inside the `acc.BuildAnd...` methods.
I thought this would be easier than making something like this inside each method:

``` go
	if opts.simulationFlag != "" {
		// validate...
	}

	var blockID rpc.BlockID
	if opts.usePending {
		blockID = rpc.WithBlockTag(rpc.BlockTagPending)
	} else {
		blockID = rpc.WithBlockTag(rpc.BlockTagLatest)
	}
        // ...
```

account/txn_options.go
thiagodeev: Well, the second func could indeed be replaced just by a `txnOptions` return. Makes sense

``` go
func NewTxnOpts() *txnOptions {
	return &txnOptions{
		tip:                 rpc.U64("0x0"),
		withQueryBitVersion: false,
		multiplier:          1.5,
		usePending:          true,
		simulationFlag:      "",
	}
}
```

account/txn_options.go
thiagodeev: I agree, I honestly prefer my previous approach.
My intention with this approach was to only make possible for the user to use `txnOptions` through this builder, removing then the possibility of "incorrect use" you described. This was the "simplest" way that I found to do so.

>I think you should only need one type, TxnOptions can have all their fields public + constructor

But with the type & fields being public, the user can populate and pass it directly, without having to rely on the constructor for it. That being said, we'll have to process the data before using it inside the methods. Since we'll need to do so, it's easy to just validate it and not use the constructor. Makes sense?

thiagodeev: My main 'pain point' is how to handle the multiplier since you didn't like the idea of having a default of 1.5 overriding it if 0 or less 🫠

account/txn_options.go
```diff
@@ -0,0 +1,75 @@
...
+	if opts.Multiplier <= 0 {
+		opts.Multiplier = 1.5
+	}
+
+	if opts.Tip == "" {
+		opts.Tip = "0x0"
+	}
+
+	return *opts
+}
```
rodrodros: This function is not very clear, it works as constructor but also modifies an argument if it is passed and it doesn't give code too much clarity. 

I think you should remove it and have just a 

```golang
func defaultTxnOptions() TxnOptions {
		return TxnOptions{
			Tip:            "0x0",
			UseQueryBit:    false,
			Multiplier:     1.5,
			UseLatest:      false,
			SimulationFlag: "",
		}
}

```

thiagodeev: But we still need to check for Tip and Multiplier. I'll need to create a separate function just for that then

thiagodeev: Done. What about now?

rodrodros (approved):
Looks good

PR #736, opened 2025-05-16
thiagodeev/abigen

thiagodeev:
Todo:
- at the end, resolve and remove all @todo comments


thiagodeev:
Unfortunately, the `abigen` go-ethereum code is under a GPL + LGPL license that conflicts with our MIT license, so we decided to create a new one from scratch.

thiagodeev:
All go-ethereum `abigen` code has been removed. I'll now proceed to create a new implementation from scratch

PR #739, opened 2025-05-20
refactor: use maps.copy for cleaner map handling

carrychair:
use maps.copy for cleaner map handling


thiagodeev:
Hey @carrychair, thanks for the suggestion, but this code came from Geth, with just minor changes to adapt to starknet spec, so I'll prefer to leave it as is

PR #744, opened 2025-05-23
`curve` pkg migration to `gnark-crypto` implementation

thiagodeev:
## Overview
This PR migrates the curve package from the custom StarkCurve implementation to use the gnark-crypto library, resulting in a more maintainable, efficient, and well-tested implementation of StarkNet curve operations.

### Key Changes
- Replace the custom StarkCurve implementation with gnark-crypto library;
- Remove unused code and dependencies;
- Add hex normalization utility for key handling;

### Tests and Benchmarks
The new implementation was tested alongside the old one, ensuring compatibility between both.  Also, both implementations were benchmarked, and the results were great for the new implementation.
You can check the tests and benchmarks [here](https://github.com/NethermindEth/starknet.go/blob/703a90450c18e41070f13d6724bff57870ec60b6/curve/curve_test.go) from a previous commit from this branch.

When signing:
- **3.75x** faster execution (1.77ms vs 0.47ms)
- **240x** less memory (738KB vs 3KB)
- **289x** fewer memory allocations (12,427 vs 43)

When verifying:
- **14x** faster execution (4950ms vs 357ms)
- **660x** less memory usage (2MB vs 3KB)
- **590x** fewer allocations (33,616 vs 57)

### Removed:
- `StarkCurve` type and all its methods
  - `StarkCurve.Add()`
  - `StarkCurve.Double()`
  - `StarkCurve.ScalarMult()`
  - `StarkCurve.ScalarBaseMult()`
  - `StarkCurve.IsOnCurve()`
  - `StarkCurve.InvModCurveSize()`
  - `StarkCurve.MimicEcMultAir()`
  - `StarkCurve.EcMult()`
  - `StarkCurve.SignFelt()`
  - `StarkCurve.GenerateSecret()`
  - `StarkCurve.GetYCoordinate()` -> func `GetYCoordinate()`
  - `StarkCurve.Verify()` -> func `Verify()`
  - `StarkCurve.Sign()` -> func `Sign()`
  - `StarkCurve.PrivateToPoint()` -> func `PrivateKeyToPoint()`
  - `StarkCurve.GetRandomPrivateKey()` -> func `GetRandomKeys()`
- `StarkCurvePayload` type
- `PedersenParamsRaw` variable
- `PedersenParams` variable
- `Curve` variable implementing the `StarkCurve`
- `VerifySignature()` function -> func `Verify()`
- `CurveOption` type
- `WithConstants` function
- `curve.DivMod` function
- `curve.FmtKecBytes` function
- `curve.MaskBits` function
- `pedersen_params.json` file

### New:
- `Verify(msgHash, r, s, pubX *big.Int) (bool, error)` - New signature verification using Juno + gnark-crypto
- `VerifyFelts(msgHash, r, s, pubX *felt.Felt) (bool, error)` - Same as Verify, but with felt.Felt parameters
- `Sign(msgHash, privKey *big.Int) (r, s *big.Int, err error)` - New signing method using gnark-crypto
- `GetRandomKeys() (privKey, x, y *big.Int, err error)` - New random key generation, using gnark-crypto, that returns a private key, X (public key), and Y
- `PrivateKeyToPoint(privKey *big.Int) (x, y *big.Int)` - New private key to point conversion, using gnark-crypto
- `GetYCoordinate(starkX *felt.Felt) *felt.Felt` - New y-coordinate calculation using gnark-crypto, taken from Juno

### Migration summary:
- `StarkCurve.GetRandomPrivateKey()` -> `GetRandomKeys()`
- `StarkCurve.GetYCoordinate()` -> `GetYCoordinate()`
- `StarkCurve.Sign()` -> `Sign()`
- `StarkCurve.PrivateToPoint()` -> `PrivateKeyToPoint()`
- `StarkCurve.Verify()` -> `Verify()`
- `VerifySignature()` -> `Verify()`

Plus:
- New `FillHexWithZeroes` utility function to fill a hex string with zeroes until it reaches 66 characters


account/account_test.go
```diff
@@ -408,74 +408,6 @@ func TestChainId(t *testing.T) {
...
-//   - Calls the Sign method of the account with the felt to sign and retrieves the signature.
-//   - Verifies that the obtained signature matches the expected signature.
-//
-// Parameters:
-//   - t: The testing.T instance for running the test
-//
-// Returns:
-//
-//	none
-func TestSignMOCK(t *testing.T) {
```
thiagodeev, line 429: Removed because this new curve implementation generates a different signature every time, so this test (that compares an old signature with a new one) doesn't apply to it. It seems like it uses random numbers as a salt when signing

internal/utils/keccak_test.go
thiagodeev, line 1: Tests migrated from `curve_test.go`

curve/curve_test.go
```diff
@@ -338,242 +250,62 @@ func TestGeneral_ComputeHashOnElements(t *testing.T) {
...
-// The test cases consist of program hashes, program outputs,
-// and expected hash values.
-//
-// Parameters:
-//   - t: The testing.T object for running the test
-//
-// Returns:
-//
-//	none
-func TestGeneral_ComputeFact(t *testing.T) {
```
thiagodeev, line 380: Moved to `internal/utils/keccak_test.go`

curve/curve_test.go
```diff
@@ -338,242 +250,62 @@ func TestGeneral_ComputeHashOnElements(t *testing.T) {
...
-// It verifies the behavior of the SplitFactStr function by providing different inputs and checking the output.
-// The function takes no parameters and returns no values.
-//
-// Parameters:
-//   - t: The testing.T object for running the test
-//
-// Returns:
-//
-//	none
-func TestGeneral_SplitFactStr(t *testing.T) {
```
thiagodeev, line 516: Moved to `internal/utils/keccak_test.go`

curve/curve_test.go
```diff
@@ -1,322 +1,234 @@
...
-// The function takes a 2D slice of big.Int values as input and measures the time
-// it takes to execute the PedersenHash function for each test case.
-//
-// Parameters:
-//   - b: a *testing.B value representing the testing context
-//
-// Returns:
-//
-//	none
-func BenchmarkPedersenHash(b *testing.B) {
```
thiagodeev, line 30: Removed since the `Pedersen` function is not part of our code, we don`t need to benchmark it

curve/curve_test.go
```diff
@@ -1,322 +1,234 @@
...
-// The function tests the PedersenHash method by providing different test cases and comparing the computed hash with the expected hash.
-// It uses the testing.T type from the testing package to report any errors encountered during the tests.
-//
-// Parameters:
-//   - t: a *testing.T value representing the testing context
-//
-// Returns:
-//
-//	none
-func TestGeneral_PedersenHash(t *testing.T) {
```
thiagodeev, line 159: Removed since the `Pedersen` function is not part of our code, it's already tested on Juno

curve/curve_test.go
```diff
@@ -1,322 +1,234 @@
...
-// It then compares the output of DivMod with the expected output and throws an error if they are not equal.
-// The function is used to test the correctness of the DivMod function.
-//
-// Parameters:
-//   - t: a *testing.T value representing the testing context
-//
-// Returns:
-//
-//	none
-func TestGeneral_DivMod(t *testing.T) {
```
thiagodeev, line 202: Removed since the DivMod utility was removed

curve/curve_test.go
```diff
@@ -1,322 +1,234 @@
...
-// It iterates over the test cases, computes the sum using the Add function, and checks if it matches the expected sum.
-// If the computed sum does not match the expected sum, an error is reported using the require.Equal function.
-//
-// Parameters:
-//   - t: a *testing.T value representing the testing context
-//
-// Returns:
-//
-//	none
-func TestGeneral_Add(t *testing.T) {
```
thiagodeev, line 240: Removed since the `Add` method, and the StarkCurve type were removed

curve/curve_test.go
```diff
@@ -1,322 +1,234 @@
...
-// The function then compares the calculated values with the expected values and reports any
-// mismatches as test failures.
-//
-// Parameters:
-//   - t: a *testing.T value representing the testing context
-//
-// Returns:
-//
-//	none
-func TestGeneral_MultAir(t *testing.T) {
```
thiagodeev, line 283: Removed since the `MultAir` method, and the `StarkCurve` type were removed

curve/curve_test.go
```diff
@@ -338,242 +250,62 @@ func TestGeneral_ComputeHashOnElements(t *testing.T) {
...
 
-// TestGeneral_HashAndSign is a test function that verifies the hashing and signing process.
-//
-// Parameters:
-//   - t: The testing.T object for running the test.
-//
-// Returns:
-//
-//	none
-func TestGeneral_HashAndSign(t *testing.T) {
```
thiagodeev, line 349: Removed since the `TestPrivateKeyEndToEnd` and the `TestSignature` replace it

curve/curve_test.go
```diff
@@ -338,242 +250,62 @@ func TestGeneral_ComputeHashOnElements(t *testing.T) {
...
-// operations including signing, verifying, converting, and initializing
-// variables.
-//
-// Parameters:
-//   - t: The testing.T object for running the test
-//
-// Returns:
-//
-//	none
-func TestGeneral_Signature(t *testing.T) {
```
thiagodeev, line 450: Removed since the new curve implementation generates signatures using random salts, so a signature, even over the same message hash, will not be the same as the previous signature

account/account.go
```diff
@@ -76,10 +76,11 @@ RPC 0.8.0. Ref: https://community.starknet.io/t/starknet-devtools-for-0-13-5/115
...
 // Parameters:
-//   - provider: is the provider of type rpc.RpcProvider
-//   - accountAddress: is the account address of type *felt.Felt
-//   - publicKey: is the public key of type string
-//   - keystore: is the keystore of type Keystore
+//   - provider: the provider to use
+//   - accountAddress: the account address
+//   - publicKey: the public key of the account
+//   - keystore: the keystore to use
+//   - cairoVersion: the cairo version of the account (0 or 2)
```
rodrodros, line 83: Why not have an enum for Cairo Version? I guess this is old code but maybe for the future

account/account_test.go
rodrodros: Maybe you can specify a specific random seed so the salt is deterministic?

account/account.go
thiagodeev: I agree

account/account_test.go
thiagodeev: We can't. The `ecdsa.Sign` function doesn't accept any salt param ([ref](https://github.com/Consensys/gnark-crypto/blob/master/ecc/stark-curve/ecdsa/ecdsa.go#L235)), only the `mshHash` and a `hashFunc`.
I guess [this](https://github.com/Consensys/gnark-crypto/blob/0517915476d283b92224365bd8659bca10153ffd/ecc/stark-curve/ecdsa/ecdsa.go#L246) is the source of the randomness.
But I see this as a feature, seems to be more secure in some way

rodrodros (approved):
LGTM!

PR #771, opened 2025-07-12
Optimize regular expression usage in typedData package

reallesee:
Extract duplicated regexp pattern into a package-level variable to avoid repeated compilation and improve performance.


thiagodeev (approved):
Thanks @reallesee!

PR #775, opened 2025-08-01
Don't omitempty for TransactionExecErrData TransactionIndex

tarrencev:
Currently it results in malformed json when marshaling with 0 index


thiagodeev:
Thanks @tarrencev!
As we talked in Telegram, merged in https://github.com/NethermindEth/starknet.go/pull/776

PR #789, opened 2025-09-17
Thiagodeev/Fix SNIP-12 JSON Marshaling for TypedData

thiagodeev:
Problem
The TypedData struct wasn't preserving the original JSON format when marshaling/unmarshaling, causing issues with SNIP-12 compliance and backward compatibility. It was also invalid in some points.

Solution
- Added JSON tags to the `TypedData` struct
- Added custom JSON marshaling for `Domain` and `TypeDefinition` structs
- Support both `chainId` and `chain_id` field names for backward compatibility
- Preserve original data types (string vs numeric) in marshaled output

Key Changes
- Custom UnmarshalJSON/MarshalJSON methods for exact JSON preservation
- Internal flags to track original field types and naming
- New test TestMarshalJSON to validate output matches input
- New benchmark BenchmarkUnmarshalJSON for performance testing

Testing
✅ All linting issues resolved
✅ New tests pass and validate exact JSON preservation
✅ Backward compatibility maintained
✅ Performance benchmarks added

### DIFF
--- a/typedData/typedData.go
+++ b/typedData/typedData.go
@@ -991,7 +991,7 @@ func (domain Domain) MarshalJSON() ([]byte, error) {
 	} else {
 		chainId, err = strconv.Atoi(domain.ChainId)
 		if err != nil {
-			return nil, err
+			return nil, fmt.Errorf("cannot parse 'chain_id' value: %w", err)
 		}
 	}
 
--- a/typedData/typedData.go
+++ b/typedData/typedData.go
@@ -426,8 +426,16 @@ func EncodeData(typeDef *TypeDefinition, td *TypedData, context ...string) (enc 
 			return enc, err
 		}
 
+		// The ChainId can be either `chainId` or `chain_id`.
 		// ref: https://community.starknet.io/t/signing-transactions-and-off-chain-messages/66
-		domainMap["chain_id"] = domainMap["chainId"]
+		// We find the one that contains the value and we copy the value to the other field.
+		// It's an workaround to handle both cases.
+
+		if domainMap["chainId"] != nil {
+			domainMap["chain_id"] = domainMap["chainId"]
+		} else {
+			domainMap["chainId"] = domainMap["chain_id"]
+		}
 
 		return encodeData(typeDef, td, domainMap, false, context...)
 	}
@@ -994,7 +1002,7 @@ func (domain Domain) MarshalJSON() ([]byte, error) {
 		revision = nil
 	} else {
 		if domain.HasStringRevision {
-			revision = fmt.Sprintf("%v", domain.Revision)
+			revision = strconv.FormatUint(uint64(domain.Revision), 10)
 		} else {
 			revision = domain.Revision
 		}
--- a/typedData/typedData.go
+++ b/typedData/typedData.go
@@ -32,9 +32,10 @@ type Domain struct {
 	ChainId  string `json:"chainId"`
 	Revision uint8  `json:"revision,omitempty"`
 
-	// Flags to deal with edge chainId cases, used to marshal the ChainId exactly as it is in the original JSON.
+	// Flags to deal with edge cases, used to marshal the `Domain` exactly as it is in the original JSON.
 	hasStringChainId  bool `json:"-"`
 	hasOldChainIdName bool `json:"-"`
+	HasStringRevision bool `json:"-"`
 }
 
 type TypeDefinition struct {
@@ -906,13 +907,24 @@ func (domain *Domain) UnmarshalJSON(data []byte) error {
 		return err
 	}
 
-	revision, err := getField("revision")
-	if err != nil {
-		revision = "0"
+	// Custom logic to handle the `revision` field,
+	// used to marshal the Revision exactly as it is in the original JSON.
+	rawRevision, ok := dec["revision"]
+	if !ok {
+		rawRevision = "0"
 	}
-	numRevision, err := strconv.ParseUint(revision, 10, 8)
-	if err != nil {
-		return err
+
+	var numRevision uint64
+	switch revision := rawRevision.(type) {
+	case string:
+		domain.HasStringRevision = true
+		numRevision, err = strconv.ParseUint(revision, 10, 8)
+		if err != nil {
+			return err
+		}
+	case float64:
+		domain.HasStringRevision = false
+		numRevision = uint64(revision)
 	}
 
 	// Custom logic to handle the `chainId` field,
@@ -946,6 +958,7 @@ func (domain *Domain) UnmarshalJSON(data []byte) error {
 	}
 	chainId := fmt.Sprintf("%v", rawChainId)
 
+	// Final step
 	*domain = Domain{
 		Name:     name,
 		Version:  version,
@@ -954,18 +967,20 @@ func (domain *Domain) UnmarshalJSON(data []byte) error {
 
 		hasStringChainId:  domain.hasStringChainId,
 		hasOldChainIdName: domain.hasOldChainIdName,
+		HasStringRevision: domain.HasStringRevision,
 	}
 
 	return nil
 }
 
-// MarshalJSON implements the json.Marshaler interface for Domain
+// MarshalJSON implements the json.Marshaler interface for Domain.
+// Some logic was added to marshal the `Domain` exactly as it is in the original JSON.
 func (domain Domain) MarshalJSON() ([]byte, error) {
 	var chainId any
+	var revision any
 	var err error
 
-	// The purpose of this is to marshal the ChainId exactly as it is in the original JSON.
-	// So, for example, if it's `1`, we marshal it as `1`, not `"1"`.
+	// E.g: if it's `1`, we will marshal it as `1`, not `"1"`.
 	if domain.hasStringChainId {
 		chainId = domain.ChainId
 	} else {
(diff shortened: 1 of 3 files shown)


typedData/typedData.go
```diff
@@ -19,18 +19,23 @@ import (
...
-	PrimaryType string
-	Domain      Domain
-	Message     map[string]any
-	Revision    *revision
+	Types       map[string]TypeDefinition `json:"types"`
+	PrimaryType string                    `json:"primaryType"`
+	Domain      Domain                    `json:"domain"`
+	Message     map[string]any            `json:"message"`
+	Revision    *revision                 `json:"-"`
 }
```
rodrodros, line 27: Why do you store the Revision twice (here and in `Domain`)?

typedData/typedData.go
```diff
@@ -902,34 +915,128 @@ func (domain *Domain) UnmarshalJSON(data []byte) error {
...
 		}
-		var err2 error
+
+		// `chain_id` was also used in the past, so we check for it if the `chainId` field is not found
 		// ref: https://community.starknet.io/t/signing-transactions-and-off-chain-messages/66
-		chainId, err2 = getField("chain_id")
-		if err2 != nil {
+		rawChainId, ok = dec["chain_id"]
+		if !ok {
+			err2 := errors.New("error getting the value of 'chain_id' from 'domain' struct")
```
rodrodros: Suggestion: "error getting the value of 'chain_id' from the 'domain' field"

Or something along those lines, I am not too happy with my suggestion. I assume this error is for the users and they have to be agnostic to the implementation. `struct` is a programming concept that they may or may not understand

typedData/typedData.go
```diff
@@ -902,34 +915,128 @@ func (domain *Domain) UnmarshalJSON(data []byte) error {
...
+
+	// Custom logic to handle the `chainId` field,
+	// used to marshal the ChainId exactly as it is in the original JSON.
+	rawChainId, ok := dec["chainId"]
+	if !ok {
+		err = errors.New("error getting the value of 'chainId' from 'domain' struct")
 	}
 
-	chainId, err := getField("chainId")
 	if err != nil {
```
rodrodros: This is weird, there is an error check, but I am not sure which error are you trying to check?

typedData/typedData.go
rodrodros: `err2` is not a very good name; please use a more descriptive one, or initialize the expression directly in the return statement

typedData/typedData.go
```diff
@@ -902,34 +915,128 @@ func (domain *Domain) UnmarshalJSON(data []byte) error {
...
+		hasOldChainIdName: domain.hasOldChainIdName,
+		HasStringRevision: domain.HasStringRevision,
 	}
 
 	return nil
 }
+
+// MarshalJSON implements the json.Marshaler interface for Domain.
+// Some logic was added to marshal the `Domain` exactly as it is in the original JSON.
+func (domain Domain) MarshalJSON() ([]byte, error) {
```
rodrodros, line 983: Receive `Domain` by pointer instead of by value. This will cause a very heavy copy. I am surprised there is no linter error here? Do you have golangci-linter as part of the actions?

typedData/typedData.go
```diff
@@ -902,34 +915,128 @@ func (domain *Domain) UnmarshalJSON(data []byte) error {
...
+	var revision any
+	var err error
+
+	// E.g: if it's `1`, we will marshal it as `1`, not `"1"`.
+	if domain.hasStringChainId {
+		chainId = domain.ChainId
+	} else {
+		chainId, err = strconv.Atoi(domain.ChainId)
+		if err != nil {
+			return nil, err
```
rodrodros: Please give some context to this error: fmt.Errorf("cannot parse \"chain_id\" value: %w", err)

typedData/typedData.go
```diff
@@ -902,34 +915,128 @@ func (domain *Domain) UnmarshalJSON(data []byte) error {
...
+	// The purpose here is to marshal the `Domain` exactly as it is in the original JSON.
+	// This is achieved by having two chainId fields, one for the old name and one for the new name,
+	// and using the `omitempty` tag to only include one of them, the one that is the same as the original JSON.
+	// Similar for the `Revision` field.
+	var temp struct {
+		Name       string `json:"name"`
+		Version    string `json:"version"`
+		ChainIdOld any    `json:"chain_id,omitempty"` // old chainId json name
+		ChainIdNew any    `json:"chainId,omitempty"`  // new chainId json name
+		Revision   any    `json:"revision,omitempty"`
```
rodrodros, line 1017: Why are these of type `any`? Don't you know the type of each one?

typedData/typedData.go
thiagodeev: They are two different things.
The revision here is of type `*revision`, storing some methods and values used in some parts of the code. It's part of the logic, not part of the real JSON.
The revision in the `Domain` type is the revision value itself that comes from the JSON.

typedData/typedData.go
thiagodeev: Here's the inner function: https://github.com/NethermindEth/starknet.go/blob/ae6c8b88d90d42844f77f32a653e4f85504ed82b/typedData/typedData.go#L899-L906

It was just a way to not repeat these lines 5 times in this function

typedData/typedData.go
thiagodeev: Makes sense since this code is extracting this from a JSON and not from the struct.
I'll change it

thiagodeev: But it's literally a second error with the same scope as the previous error.
1 - if there's an error getting the `chainId` field, and it's revision `1` (which only has the `chainId` name), we return the error.
2- if it's revision `0`, since `chain_id` is also a valid name for it, we look for it, and if we don't find it, we return two errors together: `err` regarding `chainId` not found, and `err2` regarding `chain_id` not found.

Do you prefer `chainIdErr` and `chain_idErr` names?

typedData/typedData.go
thiagodeev: Yeah, we do have. No alert from linter.
If we receive `Domain` as a pointer here, it'll not work. 
This can help you to understand: https://chatgpt.com/share/68ccac36-7e20-8002-b1ca-023181099318
There's also this playground from someone in the web: https://go.dev/play/p/w7fGZQBjmp6
It seems this is solved for encode/v2 ([ref](https://pkg.go.dev/encoding/json#:~:text=In%20v1%2C%20MarshalJSON%20methods%20declared%20on%20a%20pointer%20receiver%20are%20only%20called%20if%20the%20Go%20value%20is%20addressable.%20In%20contrast%2C%20in%20v2%20a%20MarshalJSON%20method%20is%20always%20callable%20regardless%20of%20addressability.%20The%20CallMethodsWithLegacySemantics%20option%20controls%20this%20behavior%20difference.))

Also, it's not a `very heavy copy`. `Domain` only has a few small strings, bools, and a uint8.

It would probably work if I changed the `TypedData` type to have a `*Domain` instead of `Domain`. Do you want me to do that?

typedData/typedData.go
thiagodeev: Got it

typedData/typedData.go
thiagodeev: We can see in the `if`s statements above this that we are dealing with two possible values: they can be a `number` or a `string`. It's `any` because this way we can assign one of these two types for it, and they'll be marshalled correctly as a `number` or as a `string`

This is because of the changes in the SNIP-12 and their different versions. Previously, they were numbers; now they're strings.

typedData/typedData.go
rodrodros: I see now that you assign previously an error and do:
```golang
if !ok {
     err = <new error>
} 
if err != nil {
    // err body
}
```

I believe the second if is redundant and it is not very legible. A better refactor is
```golang
if !ok {
    // err boyd
}
```

Let me know what you think

typedData/typedData.go
rodrodros: I think you are right, let's leave `err2`. It's a nitpick thing, I am happy with whatever you decide

typedData/typedData.go
rodrodros: I see your point. Thanks for the thoughtful explanation, and the full inquiry. Makes sense as it is.

typedData/typedData.go
thiagodeev: Makes total sense haha.
Thanks, I'll update it

typedData/typedData.go
rodrodros: Yes, I thought that maybe it was possible to define an interface that grouped together int and strings, something akin to:
```golang
ChainIdOld interface { ~int | ~string}
```

But it is not possible :(

typedData/typedData.go
rodrodros: It's certainly confusing, but let's leave it as is

rodrodros (dismissed):
LGTM!

Other possible issues small issues detected, which are not strictly related with this PR:

1. typedData it's a very big file, I would consider refactoring a bit. Perhaps is not as simple but each sub field of `typedData` being it's own file may work.
2. The `revision` type has vars that are sent on `init()` which is not needed. The code can be refactored for this not to happen. From what I read, I don't think you need global revision vars since they are only used in this module.

typedData/typedData.go
thiagodeev: Done

typedData/typedData.go
thiagodeev: Yeah :/

PR #795, opened 2025-09-30
Update breaking function

dhruv035:
Minor PR to bump the juno version to v0.15.7 and update a function in curve.go which was throwing compile error


thiagodeev (approved):

thiagodeev:
Thanks @dhruv035!

PR #768, opened 2025-06-30
feat: implement SNIP-29 (issue #745)

Salonii02:
## Summary

This pull request introduces supporting methods for the **SNIP-29 Applicative Paymaster API**, enabling off-chain paymaster integration for sponsored and flexible transaction fee payments.
This PR **closes [#745](https://github.com/NethermindEth/starknet.go/issues/745)**.

## Key Changes

- A new `paymaster` pkg was created in the root of the repository to provide a dedicated interface for the paymaster API.
- It implements all methods from the OpenRPC specification:  https://github.com/starknet-io/SNIPs/blob/main/assets/snip-29/paymaster_api.json
  - `paymaster_isAvailable`
  - `paymaster_getSupportedTokens`
  - `paymaster_buildTransaction`
  - `paymaster_executeTransaction`
  - `paymaster_trackingIdToLatestHash`
- Mock and integration tests with the AVNU paymaster to assert correctness.
- New CI workflow to continuously verify the paymaster functionality.
- New `paymaster` example in the `example` folders showing how to use it.

PR #796, opened 2025-10-03
Paymaster PR 1: new paymaster pkg

thiagodeev:
Part of the effort to break down the Paymaster [PR](https://github.com/NethermindEth/starknet.go/pull/768) for better review.

This PR:
- create the Paymaster pkg
- implement the `paymaster_isAvailable` method
- implement all the errors from the SNIP-29
- new test helpers to instantiate paymasters

### DIFF
--- a/paymaster/is_available_test.go
+++ b/paymaster/is_available_test.go
@@ -12,11 +12,12 @@ import (
 )
 
 // Test the 'paymaster_isAvailable' method
+//
+//nolint:tparallel // Each subtest runs in different environments
 func TestIsAvailable(t *testing.T) {
 	t.Parallel()
 	t.Run("integration", func(t *testing.T) {
 		tests.RunTestOn(t, tests.IntegrationEnv)
-		t.Parallel()
 
 		pm, spy := SetupPaymaster(t)
 		available, err := pm.IsAvailable(context.Background())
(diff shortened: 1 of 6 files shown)


paymaster/is_available_test.go
```diff
@@ -0,0 +1,42 @@
...
+	"github.com/stretchr/testify/require"
+	"go.uber.org/mock/gomock"
+)
+
+// Test the 'paymaster_isAvailable' method
+func TestIsAvailable(t *testing.T) {
+	t.Parallel()
+	t.Run("integration", func(t *testing.T) {
+		tests.RunTestOn(t, tests.IntegrationEnv)
+		t.Parallel()
```
rodrodros: Why is `t.Parallel()` here?

paymaster/is_available_test.go
```diff
@@ -0,0 +1,42 @@
...
+		available, err := pm.IsAvailable(context.Background())
+		require.NoError(t, err)
+
+		assert.Equal(t, string(spy.LastResponse()), strconv.FormatBool(available))
+		assert.True(t, available)
+	})
+
+	t.Run("mock", func(t *testing.T) {
+		tests.RunTestOn(t, tests.MockEnv)
+		t.Parallel()
```
rodrodros: Similar question as to previous comment?

rodrodros (approved):

paymaster/is_available_test.go
thiagodeev: Indeed! Since there's only one subtest for this environment, there's no need for the `t.Parallel()` call.
I'll change it

PR #798, opened 2025-10-07
chore: update linters

rodrodros:


.github/workflows/linter.yml
```diff
@@ -14,12 +14,20 @@ jobs:
...
       - name: golangci-lint
         uses: golangci/golangci-lint-action@v8
+      # Linter for diffs only. Apply new lint rules gradually that would be a huge refactor
+      # to do in one go
+      - name: golangci-lint-diffs
+        uses: golangci/golangci-lint-action@v8
         with:
-          version: v2.4.0
\ No newline at end of file
+          version: latest
```
thiagodeev: I disagree with this change.

Arguments:
- We will be susceptible to a supply chain attack if a future version of the linter is compromised, since our CI will always get the latest release, even though it's compromised.
- Incompatible new versions could break our workflow. E.g.: when golangci v3.0.0 is released, we'll fetch it and break our CI

It's a good practice to use fixed versions and update after

WDYT?

rodrodros: 1. The attack comes from using `golangci/golangci-lines-action@v8`, not from setting `latest`. I think it creates more hassle and hurts readability. Take Copilot comments with a pinch of salt.
2. What's the big deal with out CI breaking when golangci v3.0.0 releases? We then make the decision to update to v3.0.0 or set a different one.

thiagodeev (approved):

PR #800, opened 2025-10-11
refactor: use strings.Builder to improve performance

stemlaud:
strings.Builder has fewer memory allocations and better performance.

More info: [golang/go#75190](https://github.com/golang/go/issues/75190)


stemlaud:
@thiagodeev  Hi, I have resolve the conflict and rebased to the latest code. 

Please review again.

PR #801, opened 2025-10-13
Paymaster PR 2: add `paymaster_buildTransaction` method + tests

thiagodeev:
Part 2 of the effort to break down the Paymaster [PR](https://github.com/NethermindEth/starknet.go/pull/768) for better review.

This PR:
- implement the `paymaster_buildTransaction` method
- new test helpers

### DIFF
--- a/paymaster/build_txn.go
+++ b/paymaster/build_txn.go
@@ -23,7 +23,7 @@ import (
 //   - error: An error if the request fails
 func (p *Paymaster) BuildTransaction(
 	ctx context.Context,
-	request BuildTransactionRequest,
+	request *BuildTransactionRequest,
 ) (BuildTransactionResponse, error) {
 	var response BuildTransactionResponse
 	if err := p.c.CallContextWithSliceArgs(
--- a/paymaster/build_txn.go
+++ b/paymaster/build_txn.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"encoding/json"
 	"fmt"
+	"strconv"
 
 	"github.com/NethermindEth/juno/core/felt"
 	"github.com/NethermindEth/starknet.go/client/rpcerr"
@@ -22,13 +23,13 @@ import (
 //   - error: An error if the request fails
 func (p *Paymaster) BuildTransaction(
 	ctx context.Context,
-	request *BuildTransactionRequest,
-) (*BuildTransactionResponse, error) {
+	request BuildTransactionRequest,
+) (BuildTransactionResponse, error) {
 	var response BuildTransactionResponse
 	if err := p.c.CallContextWithSliceArgs(
 		ctx, &response, "paymaster_buildTransaction", request,
 	); err != nil {
-		return nil, rpcerr.UnwrapToRPCErr(
+		return response, rpcerr.UnwrapToRPCErr(
 			err,
 			ErrInvalidAddress,
 			ErrClassHashNotSupported,
@@ -40,16 +41,16 @@ func (p *Paymaster) BuildTransaction(
 		)
 	}
 
-	return &response, nil
+	return response, nil
 }
 
 // BuildTransactionRequest is the request to build a transaction for
 // the paymaster (transaction + parameters).
 type BuildTransactionRequest struct {
 	// The transaction to be executed by the paymaster
-	Transaction *UserTransaction `json:"transaction"`
+	Transaction UserTransaction `json:"transaction"`
 	// Execution parameters to be used when executing the transaction
-	Parameters *UserParameters `json:"parameters"`
+	Parameters UserParameters `json:"parameters"`
 }
 
 // UserTransaction represents a user transaction (deploy, invoke,
@@ -60,7 +61,7 @@ type UserTransaction struct {
 	// The deployment data for the transaction, used for `deploy` and
 	// `deploy_and_invoke` transaction types.
 	// Should be `nil` for `invoke` transaction types.
-	Deployment *AccDeploymentData `json:"deployment,omitempty"`
+	Deployment *AccountDeploymentData `json:"deployment,omitempty"`
 	// The invoke data for the transaction, used for `invoke` and
 	// `deploy_and_invoke` transaction types.
 	// Should be `nil` for `deploy` transaction types.
@@ -69,33 +70,34 @@ type UserTransaction struct {
 
 // An enum representing the type of the transaction to be executed
 // by the paymaster
-type UserTxnType string
+type UserTxnType int
 
 const (
-	// Represents a deploy transaction
-	UserTxnDeploy UserTxnType = "deploy"
-	// Represents an invoke transaction
-	UserTxnInvoke UserTxnType = "invoke"
-	// Represents a deploy and invoke transaction
-	UserTxnDeployAndInvoke UserTxnType = "deploy_and_invoke"
+	// Represents a deploy transaction ("deploy" tag)
+	UserTxnDeploy UserTxnType = iota + 1
+	// Represents an invoke transaction ("invoke" tag)
+	UserTxnInvoke
+	// Represents a deploy and invoke transaction ("deploy_and_invoke" tag)
+	UserTxnDeployAndInvoke
 )
+
+// String returns the string representation of the UserTxnType.
+func (u UserTxnType) String() string {
+	return []string{"deploy", "invoke", "deploy_and_invoke"}[u-1]
+}
 
 // MarshalJSON marshals the UserTxnType to JSON.
 func (u UserTxnType) MarshalJSON() ([]byte, error) {
-	switch u {
-	case UserTxnDeploy, UserTxnInvoke, UserTxnDeployAndInvoke:
-		return json.Marshal(string(u))
-	}
-
-	return nil, fmt.Errorf("invalid user transaction type: %s", u)
+	return strconv.AppendQuote(nil, u.String()), nil
 }
 
 // UnmarshalJSON unmarshals the JSON data into a UserTxnType.
 func (u *UserTxnType) UnmarshalJSON(b []byte) error {
-	var s string
-	if err := json.Unmarshal(b, &s); err != nil {
+	s, err := strconv.Unquote(string(b))
+	if err != nil {
 		return err
 	}
+
 	switch s {
 	case "deploy":
 		*u = UserTxnDeploy
@@ -111,7 +113,7 @@ func (u *UserTxnType) UnmarshalJSON(b []byte) error {
 }
 
 // Data required to deploy an account at an address.
-type AccDeploymentData struct {
+type AccountDeploymentData struct {
 	// The expected address to be deployed, used to double check
 	Address *felt.Felt `json:"address"`
 	// The hash of the deployed contract's class
@@ -119,12 +121,21 @@ type AccDeploymentData struct {
 	// The salt used for the contract address calculation
 	Salt *felt.Felt `json:"salt"`
 	// The parameters passed to the constructor
-	ConstructorCalldata []*felt.Felt `json:"calldata"`
+	Calldata []*felt.Felt `json:"calldata"`
 	// Optional array of felts to be added to the signature
 	SignatureData []*felt.Felt `json:"sigdata,omitempty"`
-	// The Cairo version of the account contract (e.g. 1 or 2). Cairo 0 is not supported.
-	Version uint8 `json:"version"`
+	// The Cairo version of the account contract. Cairo 0 is not supported.
+	Version CairoVersion `json:"version"`
 }
+
+// An enum representing the Cairo version of the account contract
+// to be deployed. Cairo 0 is not supported.
+type CairoVersion int
+
+const (
+	// Represents the Cairo 1 version
+	Cairo1 CairoVersion = 1
+)
 
 // Calls to be executed by the paymaster and the user account address that will be called
 type UserInvoke struct {
@@ -152,30 +163,42 @@ type UserParameters struct {
 	FeeMode FeeMode `json:"fee_mode"`
 	// Optional. Time constraint on the execution
 	TimeBounds *TimeBounds `json:"time_bounds"`
+
+	// Note: even being optional, the `time_bounds` field wasn't tagged with `omitempty`
+	// to facilitate the tests. The reason is that the `time_bounds` returned by the paymaster
+	// is always filled with `null` in the response if not provided, and if we tag it
+	// with `omitempty`, the JSON field would be omitted in the request, resulting in
+	// different JSONs when comparing the request and response.
 }
 
 // An enum representing the version of the execution parameters
-type UserParamVersion string
+type UserParamVersion int
 
 const (
-	// Represents the v1 of the execution parameters
-	UserParamV1 UserParamVersion = "0x1"
+	// Represents the v1 of the execution parameters ("0x1")
+	UserParamV1 UserParamVersion = iota + 1
 )
 
+// String returns the string representation of the UserTxnType.
+func (u UserParamVersion) String() string {
+	return []string{"0x1"}[u-1]
+}
+
 // MarshalJSON marshals the UserParamVersion to JSON.
-func (v UserParamVersion) MarshalJSON() ([]byte, error) {
-	return json.Marshal(string(v))
+func (u UserParamVersion) MarshalJSON() ([]byte, error) {
+	return strconv.AppendQuote(nil, u.String()), nil
 }
 
 // UnmarshalJSON unmarshals the JSON data into a UserParamVersion.
-func (v *UserParamVersion) UnmarshalJSON(b []byte) error {
-	var s string
-	if err := json.Unmarshal(b, &s); err != nil {
+func (u *UserParamVersion) UnmarshalJSON(b []byte) error {
+	s, err := strconv.Unquote(string(b))
+	if err != nil {
 		return err
 	}
+
 	switch s {
 	case "0x1":
-		*v = UserParamV1
+		*u = UserParamV1
 	default:
 		return fmt.Errorf("invalid user parameter version: %s", s)
 	}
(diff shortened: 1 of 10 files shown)


paymaster/build_txn.go
```diff
@@ -0,0 +1,336 @@
...
+// Parameters:
+//   - ctx: The context.Context object for controlling the function call
+//   - request: The BuildTransactionRequest containing the transaction and parameters
+//
+// Returns:
+//   - *BuildTransactionResponse: The response containing typed data and fee estimate
+//   - error: An error if the request fails
+func (p *Paymaster) BuildTransaction(ctx context.Context, request *BuildTransactionRequest) (*BuildTransactionResponse, error) {
+	var response BuildTransactionResponse
+	if err := p.c.CallContextWithSliceArgs(ctx, &response, "paymaster_buildTransaction", request); err != nil {
```
rodrodros: long line, please separate the call and the if.

paymaster/build_txn.go
```diff
@@ -0,0 +1,336 @@
...
+// data along with the estimated gas cost and the maximum gas cost suggested to ensure execution
+//
+// Parameters:
+//   - ctx: The context.Context object for controlling the function call
+//   - request: The BuildTransactionRequest containing the transaction and parameters
+//
+// Returns:
+//   - *BuildTransactionResponse: The response containing typed data and fee estimate
+//   - error: An error if the request fails
+func (p *Paymaster) BuildTransaction(ctx context.Context, request *BuildTransactionRequest) (*BuildTransactionResponse, error) {
```
rodrodros: Return by value `BuildTransactionResponse` instead of `*BuildTransactionResponse`

rodrodros (commented):
Initial quick scan. 

Please set escape analysis in your editor so you are aware of variables that escape to heap

paymaster/build_txn.go
rodrodros: I think it would be good to merge the linter PR I added at the beginning instead of delaying it as we merge stuff, allowing us to apply linter changes as we merge PRs

paymaster/build_txn.go
thiagodeev: But the type is big. Even the linter was complaining

paymaster/build_txn.go
thiagodeev: Done

paymaster/build_txn.go
```diff
@@ -0,0 +1,350 @@
...
+
+// The object that defines an invocation of a function in a contract
+type Call struct {
+	// The address of the contract to invoke
+	To *felt.Felt `json:"to"`
+	// The selector of the function to invoke
+	Selector *felt.Felt `json:"selector"`
+	// The parameters passed to the function
+	Calldata []*felt.Felt `json:"calldata"`
+}
```
rodrodros, line 156: Why are `To` and `Selector` pointers to `felt.Felt` if they cannot be optional?

Why is `Calldata` a slice of pointers to `felt.Felt` instead of `felt.Felt`

rodrodros, line 156: Additionally, why not start using the new types from Juno:
1. `To` make it a `felt.Address`
2. `Selector` a `felt.Hash`

Wdyt?

paymaster/build_txn.go
```diff
@@ -0,0 +1,350 @@
...
+	// Optional array of felts to be added to the signature
+	SignatureData []*felt.Felt `json:"sigdata,omitempty"`
+	// The Cairo version of the account contract (e.g. 1 or 2). Cairo 0 is not supported.
+	Version uint8 `json:"version"`
+}
+
+// Calls to be executed by the paymaster and the user account address that will be called
+type UserInvoke struct {
+	// The address of the user account
+	UserAddress *felt.Felt `json:"user_address"`
```
rodrodros, line 143: Similar comment here, it shouldn't be called `UserAddress *felt.Felt`, as far as I can see the type definition should be
```golang
type UserInvoke struct { 
  User felt.Address `json:...`
  ...
}
```

paymaster/build_txn.go
```diff
@@ -0,0 +1,350 @@
...
+
+	return nil
+}
+
+// Data required to deploy an account at an address.
+type AccDeploymentData struct {
+	// The expected address to be deployed, used to double check
+	Address *felt.Felt `json:"address"`
+	// The hash of the deployed contract's class
+	ClassHash *felt.Felt `json:"class_hash"`
```
rodrodros, line 120: Similar idea to the previous comments

paymaster/build_txn.go
```diff
@@ -0,0 +1,350 @@
...
+	// The salt used for the contract address calculation
+	Salt *felt.Felt `json:"salt"`
+	// The parameters passed to the constructor
+	ConstructorCalldata []*felt.Felt `json:"calldata"`
+	// Optional array of felts to be added to the signature
+	SignatureData []*felt.Felt `json:"sigdata,omitempty"`
+	// The Cairo version of the account contract (e.g. 1 or 2). Cairo 0 is not supported.
+	Version uint8 `json:"version"`
```
rodrodros: I think we can have an enum of `Cairo1` and `Cairo2`. Can you clarify to me what's the difference, as far as I was aware, there is only `Cairo0` and `Cairo`

paymaster/build_txn.go
```diff
@@ -0,0 +1,350 @@
...
+// Data required to deploy an account at an address.
+type AccDeploymentData struct {
+	// The expected address to be deployed, used to double check
+	Address *felt.Felt `json:"address"`
+	// The hash of the deployed contract's class
+	ClassHash *felt.Felt `json:"class_hash"`
+	// The salt used for the contract address calculation
+	Salt *felt.Felt `json:"salt"`
+	// The parameters passed to the constructor
+	ConstructorCalldata []*felt.Felt `json:"calldata"`
```
rodrodros: nitpick: i think this can be called directly `Calldata` since it is part of `AccDeploymentData`

paymaster/build_txn.go
```diff
@@ -0,0 +1,350 @@
...
+	// Optional array of felts to be added to the signature
+	SignatureData []*felt.Felt `json:"sigdata,omitempty"`
```
rodrodros, line 126: I am wary of this. Is SignatureData a slice of arbitrary numbers, or is it something we know the size of? I think a custom struct fits better here, but open to discussing it since I might be missing context.

paymaster/build_txn.go
```diff
@@ -0,0 +1,350 @@
...
+		*u = UserTxnDeployAndInvoke
+	default:
+		return fmt.Errorf("invalid user transaction type: %s", s)
+	}
+
+	return nil
+}
+
+// Data required to deploy an account at an address.
+type AccDeploymentData struct {
```
rodrodros: Does this struct represent an "account deployment data", then I think it is also good to use its full name "AccountDeploymentData".

paymaster/build_txn.go
```diff
@@ -0,0 +1,350 @@
...
+
+// UserTransaction represents a user transaction (deploy, invoke,
+// or deploy_and_invoke).
+type UserTransaction struct {
+	// The type of the transaction to be executed by the paymaster
+	Type UserTxnType `json:"type"`
+	// The deployment data for the transaction, used for `deploy` and
+	// `deploy_and_invoke` transaction types.
+	// Should be `nil` for `invoke` transaction types.
+	Deployment *AccDeploymentData `json:"deployment,omitempty"`
```
rodrodros: I believe a better name should be called `DeploymentData` or `AccountDeploymentData` but it also won't match the field in the JSON, so perhaps is best to leave it as it is

paymaster/build_txn.go
```diff
@@ -0,0 +1,350 @@
...
+	return &response, nil
+}
+
+// BuildTransactionRequest is the request to build a transaction for
+// the paymaster (transaction + parameters).
+type BuildTransactionRequest struct {
+	// The transaction to be executed by the paymaster
+	Transaction *UserTransaction `json:"transaction"`
+	// Execution parameters to be used when executing the transaction
+	Parameters *UserParameters `json:"parameters"`
```
rodrodros: Each field here can be a value type, as far as I am aware, since none of them are optional JSON fields. What do you think?

paymaster/build_txn.go
```diff
@@ -0,0 +1,350 @@
...
+			ErrClassHashNotSupported,
+			ErrInvalidDeploymentData,
+			ErrTokenNotSupported,
+			ErrInvalidTimeBounds,
+			ErrUnknownError,
+			ErrTransactionExecutionError,
+		)
+	}
+
+	return &response, nil
```
rodrodros: Why are you returning by reference a type that was created in this same function?

This is a topic we have touched multiple times during dailies. If you're creating an object inside a function 90% of the time it's better to return it by value (and not by reference)

paymaster/build_txn.go
```diff
@@ -0,0 +1,350 @@
...
+	TimeBounds *TimeBounds `json:"time_bounds"`
+}
+
+// An enum representing the version of the execution parameters
+type UserParamVersion string
+
+const (
+	// Represents the v1 of the execution parameters
+	UserParamV1 UserParamVersion = "0x1"
+)
```
rodrodros: Why a string, instead of a number?

paymaster/build_txn.go
```diff
@@ -0,0 +1,350 @@
...
+// An enum representing the fee mode to use for the transaction
+type FeeModeType string
+
+const (
+	// Specify that the transaction should be sponsored. This argument does not
+	// guaranteed sponsorship and will depend on the paymaster provider
+	FeeModeSponsored FeeModeType = "sponsored"
+	// Default fee mode where the transaction is paid by the user in the given gas token
+	FeeModeDefault FeeModeType = "default"
+)
```
rodrodros, line 220: I think you are abusing a bit the use of strings here as well. Why not use iotas?

paymaster/build_txn.go
```diff
@@ -0,0 +1,350 @@
...
+	switch s {
+	case "sponsored":
+		*feeMode = FeeModeSponsored
+	case "default":
+		*feeMode = FeeModeDefault
+	default:
+		return fmt.Errorf("invalid fee mode: %s", s)
+	}
+
+	return nil
```
rodrodros, line 248: All this string enums represent a lot of unnecessary memory usage (storing the string, when you only care about the tipe, hence you only need a unique id (using iotas)

paymaster/build_txn.go
```diff
@@ -0,0 +1,350 @@
...
+type TipPriorityEnum string
+
+const (
+	// Relative tip priority
+	TipPrioritySlow TipPriorityEnum = "slow"
+	// Relative tip priority
+	TipPriorityNormal TipPriorityEnum = "normal"
+	// Relative tip priority
+	TipPriorityFast TipPriorityEnum = "fast"
+)
```
rodrodros: Let's not use string enums

paymaster/build_txn.go
```diff
@@ -0,0 +1,350 @@
...
+}
+
+// FeeEstimate is a detailed fee estimation (in STRK and gas token, with suggested max).
+type FeeEstimate struct {
+	GasTokenPriceInStrk       *felt.Felt `json:"gas_token_price_in_strk"`
+	EstimatedFeeInStrk        *felt.Felt `json:"estimated_fee_in_strk"`
+	EstimatedFeeInGasToken    *felt.Felt `json:"estimated_fee_in_gas_token"`
+	SuggestedMaxFeeInStrk     *felt.Felt `json:"suggested_max_fee_in_strk"`
+	SuggestedMaxFeeInGasToken *felt.Felt `json:"suggested_max_fee_in_gas_token"`
+}
```
rodrodros, line 377: Are all of these fields optional, or are some of them optional?

paymaster/main_test.go
```diff
@@ -55,3 +59,37 @@ func SetupMockPaymaster(t *testing.T) *MockPaymaster {
...
+		assert.Error(t, err)
+	} else {
+		assert.NoError(t, err)
+		assert.Equal(t, expected, actual)
+
+		marshalled, err := json.Marshal(actual)
+		assert.NoError(t, err)
+		assert.Equal(t, input, string(marshalled))
+	}
+}
```
rodrodros, line 78: Maybe I am missing something, but I don't see this function used anywhere 👀

paymaster/main_test.go
```diff
@@ -55,3 +59,37 @@ func SetupMockPaymaster(t *testing.T) *MockPaymaster {
...
+	strkPrivKey := os.Getenv("STARKNET_PRIVATE_KEY")
+	strkPubKey := os.Getenv("STARKNET_PUBLIC_KEY")
+	strkAccountAddress := os.Getenv("STARKNET_ACCOUNT_ADDRESS")
+
+	privKey = internalUtils.TestHexToFelt(t, strkPrivKey)
+	pubKey = internalUtils.TestHexToFelt(t, strkPubKey)
+	accountAddress = internalUtils.TestHexToFelt(t, strkAccountAddress)
+
+	return privKey, pubKey, accountAddress
+}
```
rodrodros: Same for this one, it seems to me they are in the wrong PR?

rodrodros (commented):
Pre-liminary review. The code looks correct, but inefficient.

It was done with no regard for resource usage, such as memory, and it should improve in that regard. 

The main resource hog is using string enum values which is unnecessary. Logic changes have to be done to make sure they are cleaned from the rest of the code. 

Small resource usage in one function returning by reference, when it should return by value. 

Finally, there are types that hold references to other types. This creates double indirection and might make the GC work more. Nonetheless, it is not straightforward to me in this first review whether those are wrong or right. We can discuss the details of this in the next daily

rodrodros:
@thiagodeev why is this not targetting main?

paymaster/build_txn.go
thiagodeev: I confused things here. The linter complains about large inputs, but you're talking about the response.
I'll change it

paymaster/build_txn.go
thiagodeev: Talked internally. The new Juno types will be implemented in a future v2.0.0 release

paymaster/build_txn.go
thiagodeev: Talked internally. The new Juno types will be implemented in a future v2.0.0 release

paymaster/build_txn.go
thiagodeev: Talked internally. The new Juno types will be implemented in a future v2.0.0 release

paymaster/main_test.go
thiagodeev: It is being used in [testing](https://github.com/NethermindEth/starknet.go/blob/0cd81af76e987e04eec00f6c8d127b46c6084153/paymaster/build_txn_test.go#L125)

paymaster/main_test.go
thiagodeev: Yes, you're right. Wrong PR haha. Removed

paymaster/build_txn.go
thiagodeev: The SNIP [says](https://github.com/starknet-io/SNIPs/blob/906b19074038005dc26d3574b6853a7c68d14e02/assets/snip-29/paymaster_api.json#L1123) it's an "Optional array of felts to be added to the signature", so we can't know the size. I think it's better to leave it as an arbitrary array since it's how it is defined in the spec

paymaster/build_txn.go
thiagodeev: This was due to a confusion on my part, but this [link](https://www.starknet.io/blog/cairo-roadmap-join-the-ride/) helped me. Fixed

paymaster/build_txn.go
```diff
@@ -0,0 +1,395 @@
...
+// Parameters:
+//   - ctx: The context.Context object for controlling the function call
+//   - request: The BuildTransactionRequest containing the transaction and parameters
+//
+// Returns:
+//   - *BuildTransactionResponse: The response containing typed data and fee estimate
+//   - error: An error if the request fails
+func (p *Paymaster) BuildTransaction(
+	ctx context.Context,
+	request BuildTransactionRequest,
```
rodrodros: Pass `request` by reference, please. There is no need to copy the entire value every time this function is called.

paymaster/build_txn.go
```diff
@@ -0,0 +1,395 @@
...
+) (BuildTransactionResponse, error) {
+	var response BuildTransactionResponse
+	if err := p.c.CallContextWithSliceArgs(
+		ctx, &response, "paymaster_buildTransaction", request,
+	); err != nil {
+		return response, rpcerr.UnwrapToRPCErr(
```
rodrodros, line 32: I think is better to return a default value
```golang
return BuildTransactionResponse{}, rpcerr....
```

The same way you return nil on an error when returning a reference, you return the default zero value in this case

paymaster/build_txn.go
```diff
@@ -0,0 +1,395 @@
...
+}
+
+// An enum representing the Cairo version of the account contract
+// to be deployed. Cairo 0 is not supported.
+type CairoVersion int
+
+const (
+	// Represents the Cairo 1 version
+	Cairo1 CairoVersion = 1
+)
```
rodrodros, line 138: Why have an enum of just one variant? If it is just one variant then is redundant holding information of it as far as I can see

paymaster/build_txn.go
```diff
@@ -0,0 +1,395 @@
...
+	// different JSONs when comparing the request and response.
+}
+
+// An enum representing the version of the execution parameters
+type UserParamVersion int
+
+const (
+	// Represents the v1 of the execution parameters ("0x1")
+	UserParamV1 UserParamVersion = iota + 1
+)
```
rodrodros, line 180: Similar comment regarding this, are there more versions planned? Can you share the source?

rodrodros (approved):

paymaster/build_txn.go
thiagodeev: But if there's an error, the var `response` will be the zero value

paymaster/build_txn.go
thiagodeev: I'm just following the snip. Since it says it's an enum, I believe it's better to leave it as is. 
https://github.com/starknet-io/SNIPs/blob/906b19074038005dc26d3574b6853a7c68d14e02/assets/snip-29/paymaster_api.json#L1128

paymaster/build_txn.go
thiagodeev: [This](https://github.com/starknet-io/SNIPs/blob/906b19074038005dc26d3574b6853a7c68d14e02/assets/snip-29/paymaster_api.json#L494) is the snip-29 API spec.
It's an enum with a single option

PR #809, opened 2025-10-23
Paymaster PR 5: examples + CI workflow

thiagodeev:
Fifth and final part of the effort to break down the Paymaster [PR](https://github.com/NethermindEth/starknet.go/pull/768) for better review.

This PR:
- creates an example of how to interact with the Paymaster using Starknet.go in the `examples` folder
- add CI tests for the Paymaster

### DIFF
--- a/.github/workflows/test_paymaster.yml
+++ b/.github/workflows/test_paymaster.yml
@@ -1,0 +1,30 @@
+name: paymaster integration tests
+
+permissions:
+  contents: read
+
+on:
+  push:
+    branches:
+      - main
+
+jobs:
+  run:
+    runs-on: ubuntu-22.04
+
+    steps:
+      - name: Checkout branch
+        uses: actions/checkout@v4
+
+      - name: Set up Go
+        uses: actions/setup-go@v4
+        with:
+          go-version-file: go.mod
+
+      - name: Run paymaster tests
+        run: cd paymaster && go test -timeout 300s -v -env integration .
+        env:
+          STARKNET_PRIVATE_KEY: ${{ secrets.TESTNET_ACCOUNT_PRIVATE_KEY }}
+          STARKNET_PUBLIC_KEY: ${{ secrets.TESTNET_ACCOUNT_PUBLIC_KEY }}
+          STARKNET_ACCOUNT_ADDRESS: ${{ secrets.TESTNET_ACCOUNT_ADDRESS }}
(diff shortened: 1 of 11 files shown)


.github/workflows/test_paymaster.yml
```diff
@@ -0,0 +1,30 @@
...
+    branches:
+      - main
+
+jobs:
+  run:
+    runs-on: ubuntu-22.04
+
+    steps:
+      - name: Checkout branch
+        uses: actions/checkout@v4
```
rodrodros: Use the last version actions/checkout@v5

.github/workflows/test_paymaster.yml
```diff
@@ -0,0 +1,30 @@
...
+
+      - name: Set up Go
+        uses: actions/setup-go@v4
```
rodrodros: Use v6

examples/paymaster/README.md
```diff
@@ -0,0 +1,22 @@
+This example demonstrates how to send transactions on Starknet with a paymaster using the Starkent.go SNIP-29 implementation, allowing you to pay fees with tokens other than STRK.
```
rodrodros: Nitpick: add a link to SNIP-29 description

.golangci.yaml
```diff
@@ -112,7 +112,12 @@ linters:
           - gosec
           - lll
           - mnd
+          - funlen # it's ok for the examples to have big main() functions
         path: examples/
+      - linters:
+          - gocritic #commentedOutCode: "curve.SignFelts..." is there just as an explanation
+          - unused # it's part of the example to have two unused functions
+        path: examples/paymaster/
```
rodrodros, line 120: I don't think adding this exception is good. Why don't you add the lint ignore directly to the file?

rodrodros (approved):
Feel free to merge, but let's solve this comment

https://github.com/NethermindEth/starknet.go/pull/809#discussion_r2457060208

.golangci.yaml
thiagodeev: I did it this way to leave the example as clean as possible. Linter comments would pollute and maybe confuse some users.
Also, since the examples are not part of our code logic, linter comments are overkill IMO.
WDYT?

rodrodros: Makes sense 👍

PR #812, opened 2025-10-24
feat: tip estimation utility + automatically tip estimation for account methods

thiagodeev:
Adds automatic tip estimation for all transaction types. All `BuildAndSend*` methods now automatically estimate and set tips based on the average tip from the latest block, improving transaction prioritization without manual configuration.

### Changes
- the `rpc.Provider.IsCompatible` method is now a function instead of a method
- new `rpc.EstimateTip` function utility to estimate the tip to be used in a transaction.
- Now, all the `account.BuildAndSend*` methods will automatically estimate and set the tip to be used in the transaction
based on the average tip of all transactions in the latest block. The user can set a custom tip by setting the `CustomTip`
field in the `TxnOptions` type to override the automatic estimation.
- In the `account.TxnOptions` type:
  - The `Tip` field was renamed to `CustomTip`.
  - The `Multiplier` field was renamed to `FeeMultiplier`.
  - New `TipMultiplier` field was added to set the multiplier to be used when automatically estimating the tip.


rpc/provider.go
```diff
@@ -192,6 +201,12 @@ type RPCProvider interface {
 	TransactionStatus(ctx context.Context, transactionHash *felt.Felt) (*TxnStatusResult, error)
 }
 
+// RPCUtils is the interface that defines the RPC utility methods.
+type RPCUtils interface {
+	IsCompatible(ctx context.Context) (bool, string, error)
+	EstimateTip(ctx context.Context) (U64, error)
+}
+
```
rodrodros: I don't understand why this is a separate interface

rpc/provider.go
```diff
@@ -192,6 +201,12 @@ type RPCProvider interface {
 	TransactionStatus(ctx context.Context, transactionHash *felt.Felt) (*TxnStatusResult, error)
 }
 
+// RPCUtils is the interface that defines the RPC utility methods.
+type RPCUtils interface {
+	IsCompatible(ctx context.Context) (bool, string, error)
```
rodrodros: This is an implementation specific thing of your provider implementation. This shouldn't be an interface

rpc/provider.go
```diff
@@ -192,6 +201,12 @@ type RPCProvider interface {
...
+	EstimateTip(ctx context.Context) (U64, error)
```
rodrodros: This also a specific implementatino detail of your type, I don't think it should be an interface either

rodrodros (dismissed):

PR #819, opened 2025-10-31
Thiagodeev/fix-fee-limit

thiagodeev:
This PR fix the bug in the`utils.FeeEstToResBoundsMap` function, which was using the max uint64 value for the L2 gas amount limit instead of the limit defined by [Starknet](https://docs.starknet.io/learn/cheatsheets/chain-info#current-limits), causing the txn to be rejected by the node when the amount was greater than the limit. Now, the limit is used correctly, and it's returned in the case of overflow.
It also adds a new feature to specify custom limits for the resource bounds.

- New `utils.CustomFeeEstToResBoundsMap` function that does the same as `utils.FeeEstToResBoundsMap`, but accepts a `utils.FeeLimits` parameter to set custom limits for the resource bounds instead of using the Starknet default limits.
- New `U128.ToBigInt()` method to improve the type usage and validation
- New tests + old tests updated

### DIFF
--- a/rpc/types_contract.go
+++ b/rpc/types_contract.go
@@ -5,6 +5,7 @@ import (
 	"encoding/json"
 	"errors"
 	"fmt"
+	"math/big"
 	"strconv"
 	"strings"
 
@@ -32,6 +33,17 @@ func (u U64) ToUint64() (uint64, error) {
 
 // 128 bit unsigned integers, represented by hex string of length at most 32
 type U128 string
+
+func (u U128) ToBigInt() (*big.Int, error) {
+	hexStr := strings.TrimPrefix(string(u), "0x")
+
+	result, ok := new(big.Int).SetString(hexStr, 16)
+	if !ok {
+		return nil, fmt.Errorf("failed to parse hex string: %v", hexStr)
+	}
+
+	return result, nil
+}
 
 type ClassOutput interface{}
 
--- a/rpc/types_contract.go
+++ b/rpc/types_contract.go
@@ -34,12 +34,22 @@ func (u U64) ToUint64() (uint64, error) {
 // 128 bit unsigned integers, represented by hex string of length at most 32
 type U128 string
 
+// ToBigInt converts the U128 type to a *big.Int.
+// If the value is greater than max uint128, returns an error.
+//
+//nolint:mnd // 16 means hex base
 func (u U128) ToBigInt() (*big.Int, error) {
 	hexStr := strings.TrimPrefix(string(u), "0x")
 
 	result, ok := new(big.Int).SetString(hexStr, 16)
 	if !ok {
 		return nil, fmt.Errorf("failed to parse hex string: %v", hexStr)
+	}
+
+	maxUint128, _ := new(big.Int).SetString("ffffffffffffffffffffffffffffffff", 16)
+
+	if result.Cmp(maxUint128) > 0 {
+		return nil, fmt.Errorf("value is greater than max uint128: %v", u)
 	}
 
 	return result, nil
--- a/rpc/types_contract.go
+++ b/rpc/types_contract.go
@@ -20,6 +20,7 @@ type NumAsHex string
 type U64 string
 
 // ToUint64 converts the U64 type to a uint64.
+// If the value is greater than max uint64, returns an error.
 func (u U64) ToUint64() (uint64, error) {
 	hexStr := strings.TrimPrefix(string(u), "0x")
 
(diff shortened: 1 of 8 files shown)


rpc/types_contract.go
```diff
@@ -33,6 +35,27 @@ func (u U64) ToUint64() (uint64, error) {
 // 128 bit unsigned integers, represented by hex string of length at most 32
 type U128 string
 
+// ToBigInt converts the U128 type to a *big.Int.
+// If the value is greater than max uint128, returns an error.
+//
+//nolint:mnd // 16 means hex base
+func (u U128) ToBigInt() (*big.Int, error) {
```
thiagodeev, line 42: An easy way to have some sort of validation and conversion in this type without a big refactor in the code.
I plan to use a better u128 type in the near future

utils/transactions.go
```diff
@@ -272,34 +328,33 @@ func toResourceBounds(
 	gasPriceInt := gasPrice.BigInt(new(big.Int))
 	gasConsumedInt := gasConsumed.BigInt(new(big.Int))
 
-	// Check for overflow
```
thiagodeev, line 275: Removed. Simplified the logic to only check for overflow in the final result

utils/transactions.go
```diff
@@ -214,29 +216,77 @@ func InvokeFuncCallsToFunctionCalls(invokeFuncCalls []rpc.InvokeFunctionCall) []
...
+//     In case of doubt, set to nil to use the default Starknet limits.
 //
 // Returns:
 //   - rpc.ResourceBoundsMapping: Resource bounds with applied multipliers
 func FeeEstToResBoundsMap(
 	feeEstimation rpc.FeeEstimation,
 	multiplier float64,
+	limitOpts *FeeLimitOpts,
 ) *rpc.ResourceBoundsMapping {
+	if limitOpts == nil {
```
rodrodros: Having optional parameters is not good, it creates ambiguity int he code, leave `FeeEstToResBoundsMap` as is, without `limitOpts` and create another function `func CustomFeeEstToResBoundMap` which works for custom networks. Refactor both functions appropriately so that functionality is not duplicated

thiagodeev: Nice idea! Done

utils/transactions.go
```diff
@@ -228,17 +267,55 @@ func FeeEstToResBoundsMap(
...
+//     If resource bounds overflow, they will be set to the max allowed value (U64 or U128).
+//   - limits: Limits for the resource bounds. They are still capped to the max allowed
+//     values (U64 or U128).
+//
+// Returns:
+//   - rpc.ResourceBoundsMapping: Resource bounds with applied multipliers and limits
+func CustomFeeEstToResBoundsMap(
+	feeEstimation rpc.FeeEstimation,
+	multiplier float64,
+	limits FeeLimits,
```
rodrodros: FeeLimits is a big struct it should be passed by reference. It's not performance critical but just to set the coding style similar to Juno that most things bigger than one word we pass it by reference

utils/transactions.go
```diff
@@ -228,17 +267,55 @@ func FeeEstToResBoundsMap(
 	feeEstimation rpc.FeeEstimation,
 	multiplier float64,
 ) *rpc.ResourceBoundsMapping {
+	bounds := CustomFeeEstToResBoundsMap(feeEstimation, multiplier, starknetLimits)
+
+	return &bounds
+}
```
rodrodros, line 274: This is from an old code but it should return by value. If this is a breaking change, please leave a `Todo` to come check for it later

utils/transactions_test.go
```diff
@@ -221,13 +225,16 @@ func TestResBoundsMapToOverallFee(t *testing.T) {
...
+//nolint:dupl,tparallel // The tests are similar, but they are testing different things. // Run sequentially to avoid race conditions with the `tests` variable.
 func TestFeeEstToResBoundsMap(t *testing.T) {
+	t.Parallel()
+
 	tests := []struct {
 		name          string
 		feeEstimation rpc.FeeEstimation
 		multiplier    float64
 		expected      rpc.ResourceBoundsMapping
+		feeLimit      FeeLimits // Only used in the `CustomFeeEstToResBoundsMap` test.
```
rodrodros, line 237: I think better than adding a new field to the struct it might be good to have a separate test for custom resources 🤔

rodrodros (dismissed):
Very good refactor. Some small comments

utils/transactions_test.go
thiagodeev: Yeah... I agree it would be more organized, but the downside is that I would need to duplicate the entire test cases and the loop. Do you want me to change it?

rodrodros: I've no strong opinion, choose what you think is best 👍

thiagodeev:
Thanks to @tarrencev for the report!

PR #821, opened 2025-11-07
feat: Blake2s implementation

thiagodeev:
From Starknet v0.14.1 onward, the hash function to calculate the compiled class hash will be Blake2s instead of Poseidon.
This PR implements the Blake2s hash function in Starknet.go, and a feature to automatically decide whether to use it or not to calculate the compiled class hash.

Key changes:
- account pkg:
  - new `UseBlake2sHash` param in the `TxnOption` struct to be used by the `BuildAndSendDeclareTxn` method. It's a pointer to a bool value, so it can be: true, false, or nil. True or false is to determine whether or not to use the blake hash, and if nil (the default value if not manually set by the user), Starknet.go will automatically fetch the current Starknet version and decide whether to use the Blake2s hash function.
- curve pkg:
  - new `Blake2s` and `Blake2sArray` functions.
- hash pkg:
  - a refactor in the internal `getByteCodeSegmentHasher`, `hashCasmClassEntryPointByType` (renamed to `hashCasmEntryPoints`), and `hashCasmClassByteCode` functions to accept a custom hash function as a parameter, making them able to be used with any provided hash function instead of using the `PoseidonArray` function hardcoded inside them.
  - a refactor in the `CompiledClassHash` function: it was converted to an internal `compiledClassHash` function, accepting a custom hash function to be used. Now we have two functions to calculate the compiled class hash:
    - `CompiledClassHash`, which calls the `compiledClassHash` function with the `PoseidonArray` hash function
    - `CompiledClassHashV2`, which calls the `compiledClassHash` function with the `Blake2s` hash function
- utils pkg:
  - new `UseBlake2sHash` param in the `TxnOption` struct to be used by the `BuildDeclareTxn` function.


account/transaction.go
```diff
@@ -59,8 +60,9 @@ func (account *Account) BuildAndSendInvokeTxn(
 		callData,
 		makeResourceBoundsMapWithZeroValues(),
 		&utils.TxnOptions{
-			Tip:         tip,
-			UseQueryBit: opts.UseQueryBit,
+			Tip:            tip,
+			UseQueryBit:    opts.UseQueryBit,
+			UseBlake2sHash: false,
```
rodrodros, line 65: This option doesn't make sense in the context of an invoke transaction right? It shouldn't be here

account/transaction.go
```diff
@@ -59,8 +60,9 @@ func (account *Account) BuildAndSendInvokeTxn(
 		callData,
 		makeResourceBoundsMapWithZeroValues(),
 		&utils.TxnOptions{
```
rodrodros, line 62: Why there is a TxnOptions defined here and another defined in the `utils` package?

account/transaction.go
```diff
@@ -465,3 +479,35 @@ func (account *Account) WaitForTransactionReceipt(
...
+//
+// Parameters:
+//   - ctx: The context
+//   - provider: The provider
+//
+// Returns:
+//   - bool: whether to use the Blake2s hash function for the compiled class hash
+//   - error: an error if any
+func shouldUseBlake2sHash(ctx context.Context, provider rpc.RPCProvider) (bool, error) {
+	block, err := provider.BlockWithTxHashes(ctx, rpc.WithBlockTag(rpc.BlockTagLatest))
```
rodrodros, line 495: Is this the cheapest RPC call you can do to get the protocol version?

account/txn_options.go
```diff
@@ -45,6 +45,14 @@ type TxnOptions struct {
...
 	// The simulation flag to be used when estimating fees. Default: none.
 	SimulationFlag rpc.SimulationFlag
+
+	// ONLY FOR THE `BuildAndSendDeclareTxn` METHOD: A pointer to a boolean flag
+	// indicating whether to use the Blake2s hash function to calculate the compiled
+	// class hash.
+	// If omitted (nil), Starknet.go will automatically fetch the current Starknet
+	// version and decide whether to use the Blake2s hash function.
+	UseBlake2sHash *bool
+	// TODO: remove this field after the Starknet v0.14.1 upgrade
```
rodrodros, line 55: Is the intention is to represent three states, let's use an enum. Which will be hash: 
- "Auto"
- "Blake2s"
- "Poseidon"

I think it is more descriptive code

curve/curve_test.go
```diff
@@ -41,11 +41,9 @@ func BenchmarkSignatureVerify(b *testing.B) {
 	for range b.N {
 		// setup
 		b.StopTimer()
-		randFelt, err := new(felt.Felt).SetRandom()
-		require.NoError(b, err)
 		hash := Pedersen(
 			internalUtils.DeadBeef,
-			randFelt,
+			new(felt.Felt).SetRandom(),
```
rodrodros: Use the felt type constructors

hash/hash.go
```diff
@@ -225,26 +223,24 @@ func getByteCodeSegmentHasher(
...
+	}
 
-		return new(felt.Felt).Add(
-			internalUtils.Uint64ToFelt(1),
-			curve.PoseidonArray(components...),
-		)
-	}, totalLen, nil
+	return new(felt.Felt).Add(
+		internalUtils.Uint64ToFelt(1),
+		hashFunc(components...),
```
rodrodros, line 240: Use the felt constructors

hash/hash.go
```diff
@@ -225,26 +223,24 @@ func getByteCodeSegmentHasher(
...
-	return func() *felt.Felt {
-		components := make([]*felt.Felt, len(segments)*2)
+	components := make([]*felt.Felt, len(segments)*2)
 
-		for i, val := range segments {
-			components[i*2] = internalUtils.Uint64ToFelt(val.Size)
-			components[i*2+1] = val.Hash()
-		}
+	for i, val := range segments {
+		components[i*2] = internalUtils.Uint64ToFelt(val.Size)
```
rodrodros: Use the felt constructors

account/transaction.go
thiagodeev: The linter complains about omitted fields in structs

account/transaction.go
thiagodeev: AFAIK, yes. The protocol version is present in the block header, and there are three endpoints that return it:
- starknet_getBlockWithTxHashes
- starknet_getBlockWithTxs
- starknet_getBlockWithReceipts

`starknet_getBlockWithTxHashes` appears to be the lightest option.

account/transaction.go
thiagodeev: Because they're different, with different options, used by different functions.
`utils.TxnOptions` is used by the `utils.Build...` functions, while `account.TxnOptions` is used by the `account.BuildAndSend...` methods.
Even though some field names are the same, their use and documentation in the code are different.

account/txn_options.go
thiagodeev: I liked the idea, but since this is temporary (will be removed after the v0.14.1 mainnet upgrade), what about leaving it as is? Less code, and it will be simple to implement in older starknet.go versions as I will need to create another release for rpcv08 users

account/transaction.go
rodrodros: And it makes sense, the comment is not about initializing the field or not, is about why does this field exist in the context of an invoke transaction.

account/transaction.go
rodrodros: My initial question remains unanswered. Why there are two transaction options? Why there are two different methods to interact with transaction options?

account/txn_options.go
rodrodros: I don't agree with this logic. Let's revise this. 

Is this code only used for invoking transactions / hence it only mean they work on the tip of the chain?

Right?

account/transaction.go
thiagodeev: Because currently, the options used by all three `BuildAndSend...` methods are the same.

Since this is temporary, the easiest and simplest way I found was to add this new flag in the existing `TxnOptions` struct. Even though the "semantic" is not perfect, there's a comprehensive description. This solution brings the benefit of zero breaking changes, and at the same time, it allows high flexibility in its use.

rodrodros (dismissed):

PR #823, opened 2025-11-19
Update UDC helpers to use the new UDC contract

### LINKED ISSUE #772: Update UDC helpers to use the new UDC contract
Class hash: 0x01b2df6d8861670d4a8ca4670433b2418d78169c2947f46dc614e69f333745c8
Contract address: 0x02ceed65a4bd731034c01113685c831b01c15d7d432f71afb1cf1634b53a2125

contract: https://github.com/OpenZeppelin/cairo-contracts/blob/main/packages/presets/src/universal_deployer.cairo
Docs: https://docs.openzeppelin.com/contracts-cairo/2.0.0/udc

Update functions like `utils.BuildUDCCalldata`, `utils.PrecomputeAddressForUDC`, and `account.DeployContractWithUDC`.
See the new changes, and set it as the default option in our enum.

HACKER097:
Fixes: #772

Here are the changes I have made:

1. Updated `udcAddressCairoV2` to the new address
2. Added `NewUDCOptions()` which makes `UDCCairoV2` the default value
3. Update tests to expect the new default value

Note: The `TestPrecomputeAddressForUDC` tests still fail, because they rely on hardcoded values from real transactions, which I assume use the old address, so their values will have to be changed too.

Please let me know if this is the expected way of updating the address and setting the default value. The `NewUDCOptions` functions does feel a bit unnecessary for this tiny change, but also less hackey.

### DIFF
--- a/utils/udc.go
+++ b/utils/udc.go
@@ -48,7 +48,15 @@ type UDCOptions struct {
 	//nolint:lll // The links would be unclickable if we break the line.
 	OriginIndependent bool
 	// The UDC version to be used. If not provided, UDCCairoV0 will be used.
-	UDCVersion UDCVersion
+	UDCVersion UDCVersion;
+}
+
+// Creates a new UDCOptions instance
+// UDCCairoV2 will be used as the default UDC version
+func NewUDCOptions() *UDCOptions {
+	return &UDCOptions{
+		UDCVersion: UDCCairoV2,
+	}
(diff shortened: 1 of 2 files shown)


thiagodeev:
Hey @HACKER097! Thanks!
I'll take a look soon. Sorry, I'm focused on rpcv0.10.0

PR #840, opened 2026-01-07
Fix outdated code examples in docs

Giri-Aayush:
The getting started and deploy account examples were broken - they used the old `rpc.NewProvider` signature without context and referenced the internal package which can't be imported externally.

Changes:
- Updated `rpc.NewProvider` to include context as first arg
- Removed internal package dependency from deploy account example
- Added env variable setup with godotenv
- Added transaction confirmation polling
- Added credential persistence to .env file

Tested locally, all examples compile and run correctly.


thiagodeev (approved):
Blind approve. Docs are managed by the devrel team.

PR #842, opened 2026-01-07
(refacto): replace block interface{} returns with BlockWithTsx/TxHashes

waseemnaseeven:
- Added typed wrappers for block variants with transactions/hashes and updated the corresponding RPC methods to return these types.

- Adapted the business logic (tip estimation and version detection) to explicitly handle confirmed and pre-confirmed outputs.

- Updated the documentation and examples to reflect the new method signatures and the explicit selection of block variants.


bimakw (approved):
Nice refactor! The typed wrapper approach is much cleaner than returning `interface{}`.

Observations:
- Switch statement pattern for handling `Block`/`PreConfirmed` is consistent across files ✓
- Docs updated properly ✓  
- Tests updated to use new wrapper types ✓

Minor: There's a small formatting issue in `transaction_test.go:230` (extra indent after `Return()`).

Otherwise LGTM!

thiagodeev:
Hey @waseemnaseeven! Thanks for your contribution.
Sorry, I'll take some time until I review yours and other PRs. We are completely focused on the next release

PR #847, opened 2026-05-31
Fix StorageAt storage key handling

Kuhai9801:
## Summary
- update `StorageAt` to accept a `StorageKey` directly
- pass the provided key to `starknet_getStorageAt` without selector-name hashing
- return the RPC storage value as `*felt.Felt`
- update tests, mocks, and the storage example call sites

## Rationale
`starknet_getStorageAt` expects a storage key, but the previous implementation accepted a string and always derived a selector from it. That prevented callers from querying arbitrary storage keys directly.

## Verification
- `go test ./rpc -run TestStorageAt -count=1`


thiagodeev:
Hey. Thanks @Kuhai9801, but Starknet.go is in process to be archived as is stated in our README, so we are not accepting new contributions

